PR #1, opened 2022-02-02
deposit: use account not msg.sender

drortirosh:
Fixed 2 points:
- addDepositTo reported by nikesh. 
- While fixing, found a similar mistake in addStakeTo


drortirosh:
renamed branch, to be consistent with issues, but had to create new PR

PR #50, opened 2022-02-27
C01 deposit manipulation

### LINKED ISSUE #2: [C01] Deposit manipulation
The addStakeTo function of the StakeManager contract allows an attacker to update the deposit record associated with another account, and manipulate it in two significant ways.
Firstly, the new funds are added to the caller's current balance instead of the current account balance. This effectively allows anyone to delete the deposit from any account. Secondly, the account's new stake delay is directly chosen by the caller and could be unreasonably long.
Consider replacing the addStakeTo function with an addStake function that only allows the caller to update their own deposit record.

drortirosh:
fixes #2


drortirosh:
fixes #2

shahafn (approved):

shahafn (approved):

PR #52, opened 2022-02-27
[H02] Duplicate validation gas accounting

### LINKED ISSUE #4: [H02] Duplicate validation gas accounting
The handleOp function of the EntryPoint contract tracks and records both the pre-operation gas and the gas consumed before the final postOp call. However, in contrast to the equivalent calculations in handleOps, both values use the same preGas value. This means that the gas used during payment validation is accounted for twice, and the wallet or paymaster will be overcharged for the operation.

drortirosh:
different gas calculation of handleOp and handleOps

fixes #4

### DIFF
--- a/contracts/EntryPoint.sol
+++ b/contracts/EntryPoint.sol
@@ -65,7 +65,7 @@ contract EntryPoint is StakeManager {
         );
 
         uint actualGasCost;
-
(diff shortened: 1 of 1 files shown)


shahafn:
Not relevant since we removed handleOp() in #61

PR #61, opened 2022-03-01
[L06] handleOp function is redundant

### LINKED ISSUE #15: [L06] handleOp function is redundant
In the EntryPoint contract, in addition to the handleOps function documented in the EIP-4337 specification, there is also a handleOp function that just implements the special case of handleOps where there is only a single UserOperation to be processed. This is inconsistent with the specification and increases the attack surface. It also introduces the possibility that the logic will differ between the two implementations, as demonstrated by Duplicate validation gas accounting.
Consider removing the handleOp function from the EntryPoint API, or using the handleOps function as its implementation.

drortirosh:
remove handleOp. the gas gain is minimal (<1000gas) for an edge case

fixes #15


shahafn:
lgtm

shahafn (approved):

PR #64, opened 2022-03-01
[L14] DepositPaymaster warning

### LINKED ISSUE #23: [L14] DepositPaymaster warning
The DepositPaymaster contract is non-compliant with the current version of the EIP, because it accesses an external oracle during validation. Consider including a warning at the top of the contract to explain the associated risks and how bundlers should decide whether to support this paymaster.

drortirosh:
fixes #23


contracts/samples/DepositPaymaster.sol
```diff
@@ -15,6 +15,9 @@ import "./IOracle.sol";
  *  (but can't use the deposit for this or further operations)
  *
  * paymasterData should hold the token to use.
+ * @notice This paymaster is will be rejected by the standard rules of EIP4337, if it uses an external oracle.
```
shahafn: "is will"...

Issue #46, opened 2022-02-27
[N24] Naming suggestions


drortirosh:
We believe some functions are variables could benefit from renaming. These are our suggestions:
• The IOracle, DepositPaymaster and TokenPaymaster contracts all have a getTokenToEthOutputPrice function, but in all cases, it's not a price (because it accounts for the amount bought) and it appears to be described backwards. Something like getTokenValueOfEth would be clearer.
• The last parameter of validateUserOp in the IWallet interface should be missingWalletFunds or additionalFundsRequired to match how it's used in the EntryPoint contract.
• The gasUsedByValidateUserOp variable should be gasUsedByValidateWalletPrepayment
• The PaymentMode options should be paymasterDeposit and walletDeposit, because neither
uses the "stake" for gas payments.
• The sender parameter in the _call function of the SimpleWallet contract should be target.
• The maxPriorityFeePerGas component of the UserOperation struct should be priorityFeePerGas .
• In BasePaymaster, the setEntrypoint and _requireFromEntrypoint functions should capitalize the "p" for consistency with the rest of the code base.

drortirosh:
note that `maxPriorityFeePerGas` was left unchanged, since it defines the maximum, not actual fee paid by the user

PR #65, opened 2022-03-01
[N03] Duplicated naming

### LINKED ISSUE #25: [N03] Duplicated naming
In the EntryPoint contract, the paymasterStake variable has the same name as one of the PaymentMode options.
           
Consider using different names to improve code readability and avoid confusion.

drortirosh:
rename PaymentMode enum values

fixes #25

### DIFF
--- a/contracts/EntryPoint.sol
+++ b/contracts/EntryPoint.sol
@@ -15,8 +15,8 @@ contract EntryPoint is StakeManager {
     using UserOperationLib for UserOperation;
 
     enum PaymentMode {
-        paymasterStake, // if paymaster is set, use paymaster's stake to pay.
-        walletStake // pay with wallet deposit.
+        ePaymasterStake, // if paymaster is set, use paymaster's stake to pay.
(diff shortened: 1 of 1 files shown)


contracts/EntryPoint.sol
```diff
@@ -15,8 +15,8 @@ contract EntryPoint is StakeManager {
     using UserOperationLib for UserOperation;
 
     enum PaymentMode {
-        paymasterStake, // if paymaster is set, use paymaster's stake to pay.
-        walletStake // pay with wallet deposit.
+        ePaymasterStake, // if paymaster is set, use paymaster's stake to pay.
```
shahafn, line 18: Why "e" prefix? Can't you just have them as "paymaster/wallet" as PaymentModes?

drortirosh: the idea was to have unique names, which don't overload any variable name.
thus "paymaster" still overloaded. "e" is the Hungarian prefix for enum value.. any better suggestion

drortirosh:
will be done as part of #46

PR #66, opened 2022-03-01
[N04] ECDSA signature length check allows invalid values

### LINKED ISSUE #26: [N04] ECDSA signature length check allows invalid values
In the VerifyingPaymaster contract, the validatePaymasterUserOp function contains a check that the ECDSA signature being verified has a length >= 65 bytes. The tryRecover function in the OpenZeppelin ECDSA library does not support signatures longer than 65 bytes. It's possible for an invalid signature length to pass the check in validatePaymasterUserOp and later cause the transaction to revert.
Consider modifying the check in validatePaymasterUserOp to only allow ECDSA signature lengths equal to 65.

drortirosh:
validate sig length is either 64 or 65, as supported by ECDSA

fixes #26


contracts/samples/VerifyingPaymaster.sol
```diff
@@ -50,7 +50,8 @@ contract VerifyingPaymaster is BasePaymaster {
         (requiredPreFund);
 
         bytes32 hash = getHash(userOp);
-        require(userOp.paymasterData.length >= 65, "VerifyingPaymaster: invalid signature length in paymasterData");
+        uint sigLength = userOp.paymasterData.length;
+        require(sigLength >= 64 && sigLength <= 65, "VerifyingPaymaster: invalid signature length in paymasterData");
```
shahafn: Why not == 64 || ==65?

PR #74, opened 2022-03-02
N22 unclear use of mixed size values

### LINKED ISSUE #44: [N22] Unclear use of mixed size values
The StakeManager contract defines the DepositInfo struct with three values of size uint112, uint32 and uint64. While the intention to pack the contents into a single word can be inferred, the reason for the particular sizes are not obvious. Consider documenting the reason for this design pattern and the corresponding (reasonable) assumptions about the maximum sizes of each type.

drortirosh:
fixes #44


shahafn (approved):

drortirosh:
will be fixed with #76

PR #76, opened 2022-03-03
[M02] Separate stake and prepayment

### LINKED ISSUE #9: [M02] Separate stake and prepayment
The StakeManager contract holds ETH on behalf of users for two different reasons:
- Paymasters lock a deposit as an anti-sybil mechanism
- Wallets and paymasters preemptively transfer funds to pay the gas costs associated with user operations

However, the StakeManager contract treats these funds equivalently, making the boundary harder to identify, enforce and reason about. For example, paymasters are unable to decrease the amount of funds available to users without first unstaking and waiting for the withdrawal period. Moreover, miners may choose different thresholds for the minimum paymaster stake, which means they can unilaterally
         
partition a paymaster's deposit between the two amounts on a per-transaction basis. Consequently, unless they explicitly account for this in their validation function, paymasters cannot safely lock capital in the StakeManager without risking that it will be used to pay for user transactions.
Additionally, we believe the Paymasters can spend locked state issue is a consequence of this unclear boundary.
Consider separating the handling of transaction prepayments and paymaster stake to better encapsulate the two concepts.

### LINKED ISSUE #17: [L08] Paymasters cannot reduce unstaking delay after withdrawal window
The StakeManager contract allows paymasters to lock funds for a period of time and they are intentionally prevented from reducing the delay. However, after unstaking their funds and waiting for the withdrawal period, they should be able to stake again with any delay. This is possible if they withdraw their funds first (which clears the saved delay), but this is should not be a necessary requirement.
Consider updating the staking guard condition to allow the delay time to be reduced if the withdrawTime has been reached.

### LINKED ISSUE #44: [N22] Unclear use of mixed size values

### LINKED ISSUE #48: [N26] Abstract StakeManager contract
The StakeManager contract is intended to provide staking functionality to other contracts, but should not be deployed directly. To better signal this intention, consider declaring the contract as abstract.

drortirosh:
separate "stake" from deposit
- keep separate stake and deposit balances.
- stake is unmodified. paymaster only pays from its deposit.
- paymaster pre-pay for the request, just like the wallet does (and
refunded at the end)

fixes #9 
fixes #17 
fixes #48 
fixes #74 
fixes #44


contracts/BasePaymaster.sol
```diff
@@ -47,24 +48,33 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
         revert("must override");
     }
 
+    //add deposit, used for paying for transaction fees
+    function deposit() public payable {
+        entryPoint.depositTo{value : msg.value}(address(this));
+    }
+
+    function withdrawTo(address payable withdrawAddress, uint amount) public payable {
```
yoavw: onlyOwner missing?

Also, it's the withdrawal function, so why "payable"?

contracts/EntryPoint.sol
```diff
@@ -296,41 +288,40 @@ contract EntryPoint is StakeManager {
...
                     catch {
-                        revert FailedOp(opIndex, op.paymaster, "postOp revert");
+                        revert FailedOp(opIndex, paymaster, "postOp revert");
                     }
                 }
             }
-            //paymaster pays for full gas, including for postOp
-            actualGas += preGas - gasleft();
-            actualGasCost = actualGas * gasPrice;
             //paymaster balance known to be high enough, and to be locked for this block
```
yoavw: Is this comment still true, now that the deposit is not locked with the stake?  Couldn't the paymaster deposit be withdrawn in an op or a postOp, causing subsequent ops to fail?

contracts/EntryPoint.sol
```diff
@@ -296,41 +288,40 @@ contract EntryPoint is StakeManager {
...
+                        revert FailedOp(opIndex, paymaster, reason);
                     }
                     catch {
-                        revert FailedOp(opIndex, op.paymaster, "postOp revert");
+                        revert FailedOp(opIndex, paymaster, "postOp revert");
                     }
                 }
             }
-            //paymaster pays for full gas, including for postOp
-            actualGas += preGas - gasleft();
```
yoavw, line 322: So paymaster is no longer charged for postOp,   And it does seem to be refunded based on prefund rather than actualGas (line 320 below).  If we remove this section, shouldn't the `internalDecrementDeposit` in line 313 charge `opInfo.prefund` rather than `actualGasCost`?

contracts/EntryPoint.sol
```diff
@@ -296,41 +288,40 @@ contract EntryPoint is StakeManager {
...
-            actualGasCost = actualGas * gasPrice;
             //paymaster balance known to be high enough, and to be locked for this block
-            internalDecrementDeposit(op.paymaster, actualGasCost);
+            internalDecrementDeposit(paymaster, actualGasCost);
```
yoavw: Shouldn't it be `prefund` instead of `actualGasCost`? See comment above in the removed section.

And if we do that, then any reason to charge for it here rather than before the op, to be consistent with the way we charge wallets?  Aside from consistency, it solves the problem of a paymaster's deposit getting withdrawn by an earlier op (now that it's no longer locked).  If we switch to a precharge/refund model for paymasters, let's consider actually precharging rather than postcharging.

contracts/StakeManager.sol
```diff
@@ -4,55 +4,78 @@ pragma solidity ^0.8;
 import "hardhat/console.sol";
 
 /**
- * manage deposit of sender or paymaster, to pay for gas.
- * paymaster must stake some of the deposit.
+ * manage deposits and stakes.
+ * deposit is just a balance used to pay for transaction (either by a paymaster or a wallet)
+ * stake is value locked for at least "unstakeDelay" by a paymaster.
  */
 contract StakeManager {
```
yoavw: Maybe make it abstract as suggested in N26, unless there's a separate PR for this.

contracts/StakeManager.sol
```diff
@@ -78,60 +101,76 @@ contract StakeManager {
...
-     * will fail if the deposit is already staked or too low.
-     * after a paymaster unlocks and withdraws some of the value, it must call addStake() to stake the value again.
+     * withdraw from the (unlocked) stake.
+     * must first call unlockStake and wait for the unstakeDelay to pass
      * @param withdrawAddress the address to send withdrawn value.
      * @param withdrawAmount the amount to withdraw.
      */
-    function withdrawTo(address payable withdrawAddress, uint withdrawAmount) external {
-        DepositInfo memory info = deposits[msg.sender];
+    function withdrawStake(address payable withdrawAddress, uint withdrawAmount) external {
```
yoavw: Now that stake is separated from deposit, does it make sense for `withdrawStake` to take a `withdrawAmount` arg rather than just withdraw the whole stake?  Is there a possible use case for partially withdrawing the stake?  Unlock already set `info.staked = false` so the stake serves no purpose even if only partially withdrawn.

reports/gas-used-output.txt
```diff
@@ -5,41 +5,43 @@
...
 ·······················|······················|············|··············|·············|···············|··············
-|  EntryPoint          ·  addStake            ·     29068  ·       46168  ·      34768  ·            3  ·          -  │
+|  EntryPoint          ·  addStake            ·     31569  ·       68569  ·      44836  ·            3  ·          -  │
 ·······················|······················|············|··············|·············|···············|··············
-|  EntryPoint          ·  handleOps           ·     98037  ·     1182289  ·     424975  ·           27  ·          -  │
+|  EntryPoint          ·  depositTo           ·     45906  ·       45918  ·      45914  ·            3  ·          -  │
 ·······················|······················|············|··············|·············|···············|··············
-|  EntryPoint          ·  simulateValidation  ·         -  ·           -  ·     996557  ·            1  ·          -  │
+|  EntryPoint          ·  handleOps           ·     97803  ·     1181195  ·     425715  ·           27  ·          -  │
```
yoavw: Interesting that this change actually made `handleOps` cheaper.  I though tit would be the opposite.

test/paymaster.test.ts
```diff
@@ -233,17 +233,17 @@ describe("EntryPoint with paymaster", function () {
...
+        ).to.revertedWith('must call unlockStake')
       })
       it('should be able to withdraw after unstake delay', async () => {
-        await paymaster.unstakeDeposit()
-        const amount = await paymaster.getDeposit()
+        await paymaster.unlockStake()
+        const amount = await entryPoint.getDepositInfo(paymaster.address).then(info=>info.stake)
         expect(amount).to.be.gte(ONE_ETH.div(2))
         await ethers.provider.send('evm_mine', [Math.floor(Date.now() / 1000) + 100])
-        await paymaster.withdrawTo(withdrawAddress, amount)
```
yoavw, line 268: Now that we replaced all `withdrawTo` tests to `withdrawStake`, should we have a couple of new tests for `withdrawTo` flows, in paymaster.test or entrypoint.test?  Or is this already sufficiently covered by the wallet case which also uses it?

yoavw (commented):
Comments inside

contracts/BasePaymaster.sol
drortirosh: yes, wrong modifier... (that is, it should be onlyOwner, not payable..)

contracts/EntryPoint.sol
drortirosh: the comment is irrelevant. the paymaster already paid.

contracts/EntryPoint.sol
drortirosh: shouldn't be here at all. there is one refund (increment) place below, for both paymaster and wallet.

contracts/StakeManager.sol
drortirosh: right

contracts/StakeManager.sol
drortirosh: good idea.

reports/gas-used-output.txt
drortirosh: moved logic to addStake/unlockStake. now "isStaked" is a simple flag check

contracts/EntryPoint.sol
yoavw: Where?

contracts/EntryPoint.sol
yoavw: But that's a decrement, not an increment.  The increment below (the refund) is fine, but here seems to be the place where the paymaster is *charged*.  And it's charged `actualGasCost` rather than `prefund`.

contracts/EntryPoint.sol
```diff
@@ -309,8 +309,6 @@ contract EntryPoint is StakeManager {
                     }
                 }
             }
-            //paymaster balance known to be high enough, and to be locked for this block
-            internalDecrementDeposit(paymaster, actualGasCost);
```
yoavw: You removed it but I don't see where the paymaster is charged at all now.  I only see the `internalDecrementDeposit` of the wallet, not the paymaster.

contracts/StakeManager.sol
```diff
@@ -175,17 +173,4 @@ contract StakeManager {
...
         require(success, "failed to withdraw");
     }
-
-    /**
-     * check if the given account is staked and didn't unlock it yet.
-     * @param account the account (paymaster) to check
-     * @param requiredStake the minimum deposit
-     * @param requiredDelaySec the minimum required stake time.
-     */
-    function isStaked(address account, uint requiredStake, uint requiredDelaySec) public view returns (bool) {
```
yoavw: Why remove this?  Isn't it potentially used as a view call by a batcher to see if a paymaster is sufficiently staked, if that batcher requires a bigger stake or a longer time?

contracts/EntryPoint.sol
drortirosh: the paymaster is charged for postOp: the actualGas (and refund) are calculated just below for all cases (both paymaster and wallet modes), and then the "refundAddress" is paid back (which is either the paymaster or wallet, depends on who originally paid)

contracts/StakeManager.sol
drortirosh: instead of calculation as is here, I added a `staked` boolean: it is set by `addStake()` (which now validates the stake delay/value are above the minimum), and cleared by `unlockStake()`
so to check the stake status, you now do `getDepositInfo(paymaster).staked`

yoavw: Yes, I know.  I meant the case where a certain batcher requires a higher bar for paymasters due to being under attack.  Previously the batcher could call `isStaked(paymaster, higherThanMinimumStake, higherThanMinimumDelay)` to decide.  Now it seems that we're setting the stake parameters as fixed values which everyone on the network must use.  Without granularity, do batchers have a way to differentiate paymasters?

contracts/EntryPoint.sol
```diff
@@ -231,17 +231,19 @@ contract EntryPoint is StakeManager {
...
         }
-        if (paymasterInfo.deposit < requiredPreFund) {
+        if (deposit < requiredPreFund) {
             revert FailedOp(opIndex, paymaster, "paymaster deposit too low");
         }
-        paymasterInfo.deposit = uint112(paymasterInfo.deposit - requiredPreFund);
+        paymasterInfo.deposit = uint112(deposit - requiredPreFund);
 
         uint gas = op.verificationGas - gasUsedByValidateUserOp;
         try IPaymaster(paymaster).validatePaymasterUserOp{gas : gas}(op, requestId, requiredPreFund) returns (bytes memory _context){
```
yoavw, line 252: I noticed that you removed "view" for this function, and realized that paymaster's validation is not explicitly a staticcall.  Does the compiler use staticcall since the interface in IPaymaster says "view", or do we need to make it explicit?

contracts/EntryPoint.sol
```diff
@@ -231,17 +231,19 @@ contract EntryPoint is StakeManager {
...
+        bool staked = paymasterInfo.staked;
+        if (!staked) {
             revert FailedOp(opIndex, paymaster, "not staked");
         }
-        if (paymasterInfo.deposit < requiredPreFund) {
+        if (deposit < requiredPreFund) {
             revert FailedOp(opIndex, paymaster, "paymaster deposit too low");
         }
-        paymasterInfo.deposit = uint112(paymasterInfo.deposit - requiredPreFund);
+        paymasterInfo.deposit = uint112(deposit - requiredPreFund);
```
yoavw, line 249: Oh, so that's why I couldn't find the place where the paymaster pays the prefund.  I was searching for `internalDecrementDeposit` which is what you use for the wallet prepayment.  Didn't realize the functionality is replicated inline here.

contracts/EntryPoint.sol
drortirosh: the validatePaymasterUserOp is still view function. its wrapper internal method also pays the prefund, and thus not view function.

contracts/StakeManager.sol
drortirosh: getDepositInfo is still a public method, to get all info about a paymaster.
Do you think we need to add some version if `isStaked()`
We do have `balanceOf()`, which is equivalent of `getDepositInfo().deposit`, but that because this seems like a common method to access (by either paymaster or wallet).

yoavw: No, if there's already a getter that makes it easy for clients to get the info and calculate, then we don't need to add unused code to the contract.

PR #82, opened 2022-03-16
[N11] IWallet doesn't strongly enforce required functionality

drortirosh:
introduce BaseWallet, which provides basic wallet functionality to
subclasses.


contracts/samples/BaseWallet.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.7;
+
+import "../IWallet.sol";
+import "../EntryPoint.sol";
+import "./ECDSA.sol";
+
+//minimal wallet
```
yoavw: These comments are about SimpleWallet, not about BaseWallet.

contracts/samples/BaseWallet.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.7;
+
+import "../IWallet.sol";
+import "../EntryPoint.sol";
+import "./ECDSA.sol";
```
yoavw: If BaseWallet needs ECDSA then it's probably the wrong abstraction level.  Signatures are supposed to be handled by the wallet itself, so BaseWallet shouldn't know what ECDSA is.

contracts/samples/BaseWallet.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// validate the signature is valid for this message.
+    /// must NOT use the "GAS" opcode while calling the "ecrecover" precompile, which is banned during validateUserOp
+    function _validateSignature(UserOperation calldata userOp, bytes32 requestId) internal virtual view;
+
+    /// validate the current nonce matches the UserOperation nonce.
+    /// also, increment the nonce, to prevent replay of this UserOperation.
+    /// called only if initCode is empty (since "nonce" field is used as "salt" on wallet creation)
+    function _validateAndIncrementNonce(UserOperation calldata userOp) internal virtual;
```
yoavw: Not sure if "Increment" is the right word.  Wallets may implement a different replay protection scheme, so maybe `_validateAndUpdateNonce`?

contracts/samples/BaseWallet.sol
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    function _requireFromEntryPoint() internal virtual view {
+        require(msg.sender == address(entryPoint()), "wallet: not from EntryPoint");
+    }
+
+    /// helper function: recover the signer of this UserOp.
+    /// NOTE: specifically,
+    /// expected to by used by validateSignature
+    function _recoverSigner(UserOperation calldata userOp, bytes32 requestId) internal virtual view returns (address) {
```
yoavw: Doesn't this belong in the wallet itself rather than BaseWallet, since it uses ECDSA?

contracts/samples/BaseWallet.sol
```diff
@@ -0,0 +1,70 @@
...
+        bytes32 hash = requestId.toEthSignedMessageHash();
+        return hash.recover(userOp.signature);
+    }
+
+    /// validate the signature is valid for this message.
```
yoavw: Maybe better to use docstring style comments, since it's used as an interface.

contracts/samples/SimpleWallet.sol
```diff
@@ -53,30 +58,13 @@ contract SimpleWallet is IWallet {
         }
     }
 
-    function updateEntryPoint(EntryPoint _entryPoint) external onlyOwner {
-        emit EntryPointChanged(entryPoint, _entryPoint);
-        entryPoint = _entryPoint;
+    function updateEntryPoint(EntryPoint newEntryPoint) external onlyOwner {
```
yoavw: Shouldn't the entrypoint stuff be in BaseWallet, to ensure that every wallet has a method to upgrade when needed?

yoavw (commented):
Not sure it's the right abstraction.  See comments.

contracts/samples/BaseWallet.sol
drortirosh: yes. will pass through all comment (and also make them natspec..)

contracts/samples/BaseWallet.sol
drortirosh: BaseWallet provide a helper function (recoverSigner) - it doesn't implement the actual verifySignature
the helper is needed, since this is now the "standard practice" to call the precompile, without using the GAS opcode..

contracts/samples/BaseWallet.sol
drortirosh: good idea.

contracts/samples/SimpleWallet.sol
drortirosh: tricky: I wanted basewallet to be storage and owner-model agnostic, and thus should avoid any members.
e.g. 
- proxy construction may try to put owner/entrypoint as "immutable" members, with special updater method in the proxy
- special storage consideration (like our current SampleWallet) which uses a single storage cell for nonce and owner

contracts/samples/BaseWallet.sol
drortirosh: no. see explanation above. also, updated the (incomplete) comment.

contracts/samples/SimpleWallet.sol
yoavw: Got it.  How about wrapping it in BaseWallet and calling a virtual _updateEntryPoint in the wallet itself, so that the wrapper just ensures this function exists and requires that the sender is EntryPoint or address(this), so it can't be accidentally left open?

contracts/samples/BaseWallet.sol
yoavw: Yes, we discussed it on the call.  If it's just a helper for these rare wallets that insist on using ECDSA ;) and can't use ecrecover due to the banned opcode, then we should keep it here.

contracts/BaseWallet.sol
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.7;
+
+import "./IWallet.sol";
+import "./EntryPoint.sol";
+import "./samples/ECDSA.sol";
+
+/**
+ * Basic wallet implmenetation.
```
arbitrary-kdawg: typo: `implmenetation` -> `implementation`

PR #83, opened 2022-03-19
[N01] Suggested EIP changes

shahafn:


eip/EIPS/eip-4337.md
```diff
@@ -171,7 +173,13 @@ While simulating `op`, the client should make sure that:
...
 5.  If `op.initcode.length != 0` ,allow only one `CREATE2` opcode call, otherwise forbid `CREATE2`.
 
-During batching, the client should exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory). After creating the batch, before executing the transaction, the client should run `eth_call` to verify the entire `handleOps` batch transaction.
+During batching, the client should:
+- Exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory).
+- Keep track of paymasters' balances, to verify it doesn't run out of funds during execution.
+
+After creating the batch, before executing the transaction, the client should:
+- Run `eth_estimateGas` with maximum possible gas,to verify the entire `handleOps` batch transaction, and use the estimated gas for the actual transaction execution.
+- Trace `FailedOp` events, and remove the matching ops that fail from the batch, until simulation succeeds.
```
drortirosh: I wonder if we should add this to the EIP: note that such FailedOp are NOT expected to happen during estimate: they should all be handled during simulation. If the fail while estimating the full batch, the paymaster involved should be throttled or banned.

eip/EIPS/eip-4337.md
```diff
@@ -162,7 +164,7 @@ If the call returns an error, the client rejects the `op`.
 
 The simulated call performs the full validation, calling both `wallet.validateUserOp` and (if specified) `paymaster.validatePaymasterUserOp`
 The two operations differ in their opcode banning.
-In order to distinguish between the two, there is a single call to NUMBER opcode (`block.number`), so the client can run "wallet validation" on the first half, 
+In order to distinguish between the two, there is a single call to NUMBER opcode (`block.number`), so the client can run "wallet validation" on the first half,
```
yoavw: This sentence seems chopped.  Shouldn't it be something like this?

In order to distinguish between the two, there is a single call to NUMBER opcode (`block.number`), so that the client can apply `wallet validation` restrictions to the first half of the simulation, and then apply `paymater validation` restrictions in the second half."

@drortirosh @shahafn doesn't this better reflect what we're doing here?

eip/EIPS/eip-4337.md
```diff
@@ -171,7 +173,13 @@ While simulating `op`, the client should make sure that:
 4. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
 5.  If `op.initcode.length != 0` ,allow only one `CREATE2` opcode call, otherwise forbid `CREATE2`.
 
-During batching, the client should exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory). After creating the batch, before executing the transaction, the client should run `eth_call` to verify the entire `handleOps` batch transaction.
+During batching, the client should:
+- Exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory).
+- Keep track of paymasters' balances, to verify it doesn't run out of funds during execution.
```
yoavw: - For each paymaster used in the batch, keep track of the balance while adding UserOps.  Ensure that it has sufficient deposit to pay for all the UserOps that use it.

eip/EIPS/eip-4337.md
```diff
@@ -171,7 +173,13 @@ While simulating `op`, the client should make sure that:
...
+
+After creating the batch, before executing the transaction, the client should:
```
yoavw: before executing the transaction -> before including the transaction in a block

eip/EIPS/eip-4337.md
```diff
@@ -52,31 +52,33 @@ To avoid Ethereum consensus changes, we do not attempt to create new transaction
 
 Users send `UserOperation` objects to a dedicated user operation mempool. A specialized class of actors called **bundlers** (either miners running special-purpose code, or users that can relay transactions to miners eg. through a bundle marketplace such as Flashbots that can guarantee next-block-or-never inclusion) listen in on the user operation mempool, and create **bundle transactions**. A bundle transaction packages up multiple `UserOperation` objects into a single `handleOps` call to a pre-published global **entry point contract**.
 
+To prevent cross-chain replay attacks, `signature` should depend on `chainid`.
```
yoavw: We should be specific and say what the signature covers.  In the code we use `keccak256(abi.encode(userOp.hash(), address(this), block.chainid))`, right?  So the signature covers the hash of the UserOp, the address of EntryPoint, and the chainId.

yoavw (commented):
Suggested some changes.

eip/EIPS/eip-4337.md
yoavw: Yes.  We should mention that FailedOp is not something normal.  Both the wallet and the paymaster that appeared in a FailedOp shouldn't be considered in the current batch, to ensure that the batch is completed without having to go through a lot of bad ops.

As for throttling/banning, this will happen naturally as no ops from this paymaster will get included.  We could additionally penalize a paymaster for FailedOp but currently we don't.  Even without this additional penalty, it'll get banned fairly quickly.

eip/EIPS/eip-4337.md
```diff
@@ -171,7 +173,13 @@ While simulating `op`, the client should make sure that:
...
+
+After creating the batch, before including the transaction in a block, the client should:
+- Run `eth_estimateGas` with maximum possible gas,to verify the entire `handleOps` batch transaction, and use the estimated gas for the actual transaction execution.
+- Trace `FailedOp` events. Remove the failed op that caused the revert from the batch and drop from the mempool. Other ops from the same paymaster should be removed from the current batch, but kept in the mempool. Repeat until simulation succeeds. Note that a failed op is usually detected on single op simulation before, thus failing here is unusual.
 
 If a contract (i) does not contain the `SELFDESTRUCT` or `DELEGATECALL` opcode [except perhaps inside pushdata] and (ii) has nonempty code, _mutable state_ of that contract is defined as storage slots (accessed with SLOAD or SSTORE). If a contract does not satisfy either of the above conditions, mutable state of that contract includes code and storage. Note that balance can not be read in any case because of the forbidden opcode restriction. _Writing_ balance (via value-bearing calls) to any address is not restricted.
```
kristofgazso: Can we remove this too then from the EIP?

kristofgazso: or at least edit to not talk about SELFDESTRUCT and DELEGATECALL

eip/EIPS/eip-4337.md
```diff
@@ -171,7 +173,13 @@ While simulating `op`, the client should make sure that:
...
+During batching, the client should:
+- Exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory).
+- For each paymaster used in the batch, keep track of the balance while adding UserOps. Ensure that it has sufficient deposit to pay for all the UserOps that use it.
+
+After creating the batch, before including the transaction in a block, the client should:
+- Run `eth_estimateGas` with maximum possible gas,to verify the entire `handleOps` batch transaction, and use the estimated gas for the actual transaction execution.
+- Trace `FailedOp` events. Remove the failed op that caused the revert from the batch and drop from the mempool. Other ops from the same paymaster should be removed from the current batch, but kept in the mempool. Repeat until simulation succeeds. Note that a failed op is usually detected on single op simulation before, thus failing here is unusual.
```
drortirosh: you say that after finding FailedOp "other ops from the same paymaster ... should be kept in the mempool"
and later "failing here is unusual"
I suggest rephrasing as:

> Trace "FailOp". Since those FailedOps should be detected on single op simulation before, failing here is unusuall, so paymaster should be throttled down

(and to my understanding this means, removing other ops from the same paymaster too)

eip/EIPS/eip-4337.md
```diff
@@ -142,7 +144,7 @@ When a client receives a `UserOperation`, it must first run some basic sanity ch
 - The paymaster is either the zero address or is a contract which (i) currently has nonempty code on chain, (ii) has registered and has sufficient stake, and (iii) has not been banned
 - The callgas is at least the cost of a `CALL` with non-zero value.
 - The `maxFeePerGas` and `maxPriorityFeePerGas` are above a configurable minimum value that the client is willing to accept
-- The `UserOperation` object is not already present in the pool (or it replaces an existing entry with the same sender, nonce with a higher priority)
+- The sender doesn't have another `UserOperation` already present in the pool (or it replaces an existing entry with the same sender, nonce with a higher priority). Only one `UserOperation` per sender per batch allowed.
```
kristofgazso: these seem to be two seperate ideas, the second one related to batching, and the first one relates to validation. could we separate them?

eip/EIPS/eip-4337.md
```diff
@@ -139,10 +156,10 @@ When a client receives a `UserOperation`, it must first run some basic sanity ch
 
 - Either the `sender` is an existing contract, or the `initCode` is not empty (but not both)
 - The `verificationGas` is sufficiently low (`<= MAX_VERIFICATION_GAS`) and the `preVerificationGas` is sufficiently high (enough to pay for the calldata gas cost of serializing the `UserOperation` plus `PRE_VERIFICATION_OVERHEAD_GAS`)
-- The paymaster is either the zero address or is a contract which (i) currently has nonempty code on chain, (ii) has registered and has sufficient stake, and (iii) has not been banned
+- The paymaster is either the zero address or is a contract which (i) currently has nonempty code on chain, (ii) has registered and staked, (iii) has a sufficient deposit to pay for the UserOperation, and (iv) has not been banned.
```
drortirosh: s/has not been banned/is not currently banned/

eip/EIPS/eip-4337.md
drortirosh: yes, this entire paragraph should be removed.

eip/EIPS/eip-4337.md
yoavw: Fixed

PR #84, opened 2022-03-24
AA-27: Document how client and bundler calculate preVerificationGas in the EIP

drortirosh:


eip/EIPS/eip-4337.md
```diff
@@ -74,6 +74,43 @@ function validateUserOp
...
+
+#### callGas
+
+The gas limit for actual callData execution
+
+#### verificationGas
+
+This value defines the maximum value the bunlder might spend on verifying the UserOperation. It should be enough to run both the `wallet.validateUserOp` and `paymaster.validatePaymasterUserOp`.
+Note that if `initCode` exists, then the verificationGas should also accomodate for creation of the sender contract.
+A bundler MAY drop requests if the verificationGas is too high.
```
yoavw: Not just the bundler.  Any node participating in the mempool, since they also need to simulate the validation once.

eip/EIPS/eip-4337.md
```diff
@@ -74,6 +74,43 @@ function validateUserOp
...
+
+The value also used as a limit on the `postOp()` call of the paymaster, if one is used.
+(and at a worst-cast might be used twice for postOp)
+
+#### preVerificationGas
+
+This value covers all the gas usage that can't be covered by GAS opcode checks.
```
yoavw, line 94: @kristofgazso does the client currently verify this value when creating a bundle?  If it does, it may be worth writing about the test here.

kristofgazso: nope, not currently, we need a defined func to be able to calculate it

PR #90, opened 2022-04-22
new ethers Signer

drortirosh:
An ethers "Signer" class to use an account-abstraction wallet contract.
- first transaction will automatically creates the wallet contract


forshtat:
Closing in favour of issue AA-26

PR #91, opened 2022-05-30
Small gas optimizations...

rodrigoherrerai:


drortirosh:
Thank you for the suggestion.
Indeed, it is more a PR for the solidity compiler optimizer rather than our project.
However, running our "gas usage" checks (do `yarn ci` and then `git diff`, to see the changes in gas-report-output.txt ), I can see this saves 13 gas on average, and 38 gas on worst-case (new contract deployment).
While we do try to save gas, I don't think that difference is enough.

PR #92, opened 2022-06-22
AA-22: Support aggregated signatures and BLS ref. implementation

drortirosh:


eip/EIPS/eip-4337.md
```diff
@@ -30,6 +30,17 @@ This proposal takes a different approach, avoiding any adjustments to the consen
     * Privacy-preserving applications
     * Atomic multi-operations (similar goal to EIP-3074)
     * Pay tx fees with ERC-20 tokens, allow developers to pay fees for their users, and [EIP-3074](./eip-3074.md)-like **sponsored transaction** use cases more generally
+    * Support aggregated signature (e.g. BLS)
+
+## Definitions
+
+* **UserOperation** - a structure that describes a transaction to be sent on behalf of a user. To avoid confusion, it is named named "transaction".
```
yoavw: "named named"?

eip/EIPS/eip-4337.md
```diff
@@ -68,35 +79,55 @@ function simulateValidation
...
+      (UserOperation calldata userOp, bytes32 requestId, uint256 missingWalletFunds)
+      external;
+}
 ```
+The wallet
+* MUST validate the caller is a trusted EntryPoint
+* The requestId is a hash over the userOp (except signature), entryPoint and chainId
+* MUST validate the signature is a valid signature of the `requestId`
+* MUST pay the entryPoint (caller) at least the "missingWalletFunds" 
+* the wallet MAY pay more than this minimum, to cover future transactions
```
yoavw: Mention that "missingWalletFunds" may be zero in some cases, e.g. if wallet previously deposited enough or if a paymaster is paying?

eip/EIPS/eip-4337.md
```diff
@@ -68,35 +79,55 @@ function simulateValidation
...
+* Note that in case the request comes through an aggregator, the signature will be empty.
+
+A Wallet that supports aggregated signature should have the interface:
+```solidity
+interface IAggregatedWallet is IWallet {
+  function getAggregator() view returns (address);
+}
+
+```
+If the wallet supports aggregation, this method should return the aggress of the aggregator.
```
yoavw: aggress -> address

eip/EIPS/eip-4337.md
```diff
@@ -147,7 +178,7 @@ function withdrawTo(address payable withdrawAddress, uint256 withdrawAmount) ext
 
 ```
 
-### Client behavior upon receiving a UserOperation
+### Bundler behavior upon receiving a UserOperation
```
yoavw: Why only Bundler?  The UserOp is checked by every client that participates in the mempool.  If Client is confusing then maybe "Bundler/Node"?

eip/EIPS/eip-4337.md
```diff
@@ -30,6 +30,17 @@ This proposal takes a different approach, avoiding any adjustments to the consen
...
+
+## Definitions
+
+* **UserOperation** - a structure that describes a transaction to be sent on behalf of a user. To avoid confusion, it is named named "transaction".
+  * Like a transaction, it contains "sender", "to", "calldata", "maxFeePerGas", "maxPriorityFee", "signature", "nonce"
+  * unlike transaction, it contains several other fields, desribed below
+  * also, the "nonce" and "signature" fields usage is not defined by the protocol, but by each account implementation
+* **Sender** - the account contract sending a user operation.
+* **EntryPoint** - a singleton contract to execute bundles of UserOperations
+* **Aggregator** - a helper contract trusted by wallets to validate an aggregated signature.
```
yoavw: Should we mention somewhere that aggregators must be whitelisted by bundlers/nodes?  Even though we don't enforce it in EP, there will initially be a whitelist to protect the mempool.  In the future we'll consider replacing it with a staked reputation system like we have for paymasters.

eip/EIPS/eip-4337.md
drortirosh: what's bad about the aggress of the addregator ??!

eip/EIPS/eip-4337.md
drortirosh: removed. though I think we mixed some Bundler and Client in the text. but that's a separate PR .

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
...
+        UserOperation[] calldata ops,
+        AggregatorInfo[] calldata aggregators
+    ) private view returns (uint totalCount) {
+        if (aggregators.length == 0) return 0;
+        uint startPtr = getMemoryPointer();
+        AggregatorInfo[] memory _aggregators = aggregators;
+
+        uint index = 0;
+        uint loopPtr = getMemoryPointer();
+        for (uint i = 0; i < _aggregators.length; i++) {
```
yoavw: Isn't `uint i` allocated in memory in the context of the `for` statement (after loopPtr is set), in which case it gets "freed" in the next line and subsequently reused to store agg?  I guess I somehow misunderstand this, since agg should have overwritten i, and the test would have failed.  Why doesn't it?

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
...
+    }
+
+    //run all aggregators, each on its respective wallets
+    function validateAggregatedSignatures(
+        UserOperation[] calldata ops,
+        AggregatorInfo[] calldata aggregators
+    ) private view returns (uint totalCount) {
+        if (aggregators.length == 0) return 0;
+        uint startPtr = getMemoryPointer();
+        AggregatorInfo[] memory _aggregators = aggregators;
```
yoavw: What's the benefit of copying `aggregators` to memory at this point?  Are we going to modify them?

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
...
+
+        uint index = 0;
+        uint loopPtr = getMemoryPointer();
+        for (uint i = 0; i < _aggregators.length; i++) {
+            restoreMemoryPointer(loopPtr);
+            AggregatorInfo memory agg = _aggregators[i];
+            uint count = agg.count;
+            agg.aggregator.validateSignatures(
+                copyOpsRange(agg.aggregator, ops, index, index + count),
```
yoavw: This would be efficient in the case of many different aggregators (since we reuse memory).  But on the most common case where there's a single aggregator (BLS), it means that we copy all the ops, leading to a lot of memory expansion. We reclaim this memory but this may be the highest watermark during this call, so it still increases the cost.

That is, unless solidity sees that it's already in memory and doesn't re-copy it  for the external `aggregator.validateSignatures` call.  Does solidity make this distinction?  If there's no double-copy then it's fine - we can't avoid having one copy operation for the external call's args.

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
...
+                agg.signature
+            );
+            index += count;
+        }
+        restoreMemoryPointer(startPtr);
+        return index;
```
yoavw: Didn't the line above just free `index`, since it was allocated after `startPtr` was set?  It would actually still work since there's nothing that would overwrite `index` between these two lines, but technically it seems like we're accessing a freed variable.  Luckily EVM is single-threaded ;)

contracts/EntryPoint.sol
```diff
@@ -111,21 +154,26 @@ contract EntryPoint is StakeManager {
...
         uint256 collected = 0;
+        uint256 execOrderLen = execOrder.length;
+        require(execOrderLen == 0 || execOrderLen == ops.length);
 
         for (uint256 i = 0; i < ops.length; i++) {
             uint256 preGas = gasleft();
-            UserOperation calldata op = ops[i];
-            UserOpInfo memory opInfo = opInfos[i];
+            uint256 opIndex = execOrderLen == 0 ? i : execOrder[i];
+            UserOperation calldata op = ops[opIndex];
```
yoavw: When `execOrderLen != 0` we should keep track of executed ops to prevent double-execution of the same op. Probably a bitmap where we set a bit for every op during validation, then test and reset it during execution.  If at least keep such bool in `UserOpInfo`.  Need to ensure that we keep the invariants:

If a validation is successful, there will be exactly one matching execution.
If a validation is not successful, there will be no execution.

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
         require(success);
     }
 
+    struct AggregatorInfo {
+        IAggregator aggregator;
+        uint count;
```
yoavw: Would be nice to have comments on struct members like we have in other places.

contracts/EntryPoint.sol
```diff
@@ -79,16 +81,56 @@ contract EntryPoint is StakeManager {
...
+        UserOperation[] calldata ops,
+        address payable beneficiary,
+        AggregatorInfo[] calldata aggregators,
+        uint[] memory execOrder
+    ) public {
 
         uint256 opslen = ops.length;
         UserOpInfo[] memory opInfos = new UserOpInfo[](opslen);
 
+        uint totalAggregated = validateAggregatedSignatures(ops, aggregators);
```
yoavw: Gas is not measured during `validateAggregatedSignatures` and the aggregator is likely to call wallet functions (e.g. the nonces() call in your sample below, or getPublicKey in the BLS case).  Simulation would have to catch this to prevent on-chain execution and drop the op before it's bundled.  If it happens in a bundle, there's no easy way for the bundler to find the culprit op.

contracts/EntryPoint.sol
```diff
@@ -181,10 +229,22 @@ contract EntryPoint is StakeManager {
...
      * @return preOpGas total gas used by validation (including contract creation)
      * @return prefund the amount the wallet had to prefund (zero in case a paymaster pays)
      */
-    function simulateValidation(UserOperation calldata userOp) external returns (uint256 preOpGas, uint256 prefund) {
+    function simulateValidation(UserOperation calldata userOp, bytes calldata aggregatedSignature) external returns (uint256 preOpGas, uint256 prefund) {
+        if (aggregatedSignature.length != 0) {
+            IAggregator aggregator = IAggregatedWallet(userOp.getSender()).getAggregator();
+            UserOperation[] memory arr = new UserOperation(1);
+            arr[0] = userOp;
+            aggregator.validateSignatures(arr, aggregatedSignature);
```
yoavw: What restrictions should apply during `aggregator.validateSignatures` execution?  Currently we have the `marker` in `_validatePrepayment`, so this part of the simulation runs as "wallet verification" part of the simulation, which is probably the right behavior.  However, it means that the aggregator is not allowed to access its own storage unless we state otherwise.  It can only access the wallet's storage.  Any reason a bundler would need its own storage?

(If there's no good reason for aggregators to have their own storage then it's actually a feature, ensuring that the aggregator is not affected by any state other than the wallet)

contracts/IAggregatedWallet.sol
```diff
@@ -0,0 +1,19 @@
...
+pragma solidity ^0.8.12;
+
+import "./UserOperation.sol";
+import "./IWallet.sol";
+import "./IAggregator.sol";
+
+/**
+ * Aggregated wallet, that support IAggregator.
+ * - the validateUserOp will be called only after the aggregator validated this wallet (with all other wallets of this aggregator).
+ * - the validateUserOp should validate that the signature field
```
yoavw: The comment is incomplete...

contracts/samples/TestAggregator.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.12;
+
+import "../IAggregator.sol";
+import "hardhat/console.sol";
+import "./SimpleWallet.sol";
+
+contract TestAggregator is IAggregator {
```
yoavw: I think the tests that use this are not committed.  They're not in this PR.

eip/EIPS/eip-4337.md
```diff
@@ -69,12 +69,19 @@ The core interface of the entry point contract is as follows:
...
 ```c++
 function handleOps
-    (UserOperation[] calldata ops, address payable beneficiary, address aggregator, bytes aggregatedSignature)
-    public;
+    (UserOperation[] calldata ops, address payable beneficiary
+     AggregatorInfo[] calldata aggregators, uint[] memory execOrder
+    ) public
 
 function simulateValidation
     (UserOperation calldata userOp)
```
yoavw: In the code you changed this interface and added `aggregatedSignature` but not in the ERC.

eip/EIPS/eip-4337.md
```diff
@@ -98,12 +105,14 @@ A Wallet that supports aggregated signature should have the interface:
 ```solidity
 interface IAggregatedWallet is IWallet {
   function getAggregator() view returns (address);
-  function getPublicKey() view returns (bytes memory);
 }
 ```
 * **getAggregator()** returns the aggregator this wallet supports.
 * The wallet is guaranteed to that the EntryPoint will only call its `validateUserOp` (with no signature) after calling the aggregator's validateSignatures
-* **getPublicKey()** is called by the aggregator, to get the wallet-specific public key.
+* The wallet should also support aggregator-specific getter (e.g. `getAggregationInfo()`).
```
yoavw, line 118: I'm not sure we can make it wallet-specific rather than a part of `IAggregatedWallet`.  In the future, we may support staked (unknown, with with reputation) aggregators.  Bundlers may not know anything about these wallets and aggregators.  The bundler needs a way to call such wallet (as a view call during simulation/bundling) and get two pieces of information:
1. The sig to be passed to an aggregation function in the aggregator (e.g. a pure function provided by the BLS aggregator, which takes an array of signatures and returns an aggregated signature which the bundler will use for the corresponding ops.
2. The "leftover".  An optional configuration data that will be passed to the wallet's validation as op.signature.

So I assume the wallet would get something like `getAggregationInfo(UserOperation)` that returns `(bytes sigForAggregator, bytes configForWallet)`.  The bundler would create a UserOperation where `op.sig=configForWallet` and simulate with `simulateValidation(op, sigForAggregator)`.

For bundling, either the bundler is aware of this specific aggregator and knows how to aggregate (e.g. using a BLS library), or it could call a pure `IAggregator.aggregateSignatures(sigs)` which knows how to aggregate the sigs and return a single one.  This generalizes the aggregators, although it may be inefficient to run aggregation in EVM, so common bundlers will probably have code in the bundler.

Does that make sense, or did you have a different idea in mind for communicating this information?

eip/EIPS/eip-4337.md
```diff
@@ -98,12 +105,14 @@ A Wallet that supports aggregated signature should have the interface:
 ```solidity
 interface IAggregatedWallet is IWallet {
```
yoavw, line 108: We have IAggregatedWallet but not IAggregator.  (It's in the code, just not in the ERC).

contracts/EntryPoint.sol
```diff
@@ -181,10 +229,22 @@ contract EntryPoint is StakeManager {
     * @dev The node must also verify it doesn't use banned opcodes, and that it doesn't reference storage outside the wallet's data.
      *      In order to split the running opcodes of the wallet (validateUserOp) from the paymaster's validatePaymasterUserOp,
      *      it should look for the NUMBER opcode at depth=1 (which itself is a banned opcode)
+     * @param userOp the user operation to validate.
+     * @param aggregatedSignature: if the wallet is an IAggregatedSignature, then the simulation of the
+     *      signature is done over a single-operation "bundle". The userOp parameter to simulate is
+     *      as it should be during aggregation (with signature removed).
```
yoavw: When the bundler simulates, how does it know what part of op.signature to use as the bundle's signature, and what part to leave intact for the wallet to use?  We said it doesn't have to be empty and may carry additional configuration.  Will the aggregator provide a pure function that parses the user's op.signature and returns (bundleSig,config)?  (See longer discussion about it in the ERC review below).

contracts/EntryPoint.sol
yoavw: Another issue with not measuring gas at this phase is that if/when we make aggregators staked and permissionless, someone could write a "cost-saving" aggregator.  This aggregator would just call a validation function in the wallet, letting the wallet perform expensive validation at the bundler's expense, and then in the "official" validation it'll just pay and set the nonce.  It means that we may be stuck with a whitelist since someone will surely deploy such aggregator to perform validation at the bundler's expense.

With a whitelist this attack is not possible, but a wallet could still increase the cost to the bundler.  The wallet would return an aggregator and a BLS key, but the "config" part of the sig will be its usual ECDSA sig.  The bundler would spend eth on this redundant BLS validation which the wallet doesn't pay for.

I don't see an easy way to attribute aggregated validation to individual wallets without trusting the aggregator code (which wouldn't be great).  But not being able to attribute this gas, seems to cause some issues.

yoavw (commented):
Long review :)

contracts/EntryPoint.sol
drortirosh: calldata access is more expensive, since it validates boundaries on each access.
memory is cheaper, since the entire validation is done once.
I didn't make the parameter "memory", so I can defer copying it into memory only if needed
(but not I think it is useless, since copy a zero-length array is also nothing..

contracts/EntryPoint.sol
drortirosh: no. local members are on the stack, not memory.
FUTURE versions of solidity might move stack members to memory, to avoid avoid "stack too deep", but even then, highly-used variables (and local vars of small functions) should stay on the stack.

contracts/EntryPoint.sol
drortirosh: yes, batch validation is not calculated per-op. There is no way I can calculate it on-chain - it is called from deep inside the aggregator  itself.
However, since the `aggregator.validate()` function is view, you can estimate it, and divide by op count.
it will to be added to preVerificationGas.
aggregator owner should publish numbers for wallets to put for validation cost.

contracts/EntryPoint.sol
drortirosh: for each aggregator, there should be an off-chain code that take an array of UserOps, and return the needed inputs for `handleOps` - shorten the UserOps and the aggregated signature
Most probably it only removes the signatures from the UserOps.
In case of multisig, as Vitalik suggested, this off-chain code has some more work to do to calculate the aggregated signature, and then this code will also need to create the "metadata" in the signature field (of course, matching the on-chain code to parse it)

The bottom line is that this is not a generic work a bundler can do, but it depends on per-aggregator off-chain library code.

contracts/EntryPoint.sol
drortirosh: again, no. its a stack, not memory item

contracts/EntryPoint.sol
drortirosh: 1. validation is view function. you can waste gas, but not execute anything
2. aggregators are "whitelisted", in a sense that a aggregator's address is tied to the off-chain signature  code in the bundler .

contracts/EntryPoint.sol
drortirosh: exactly. see the `opInfo.executed` few lines above
(yes, I could implement it as a bitfield. I was thinking of re-using another field in the opInfo - either reusing a bit, or just  making it invalid to cause revert on re-execution. Just didn't figure out how.

eip/EIPS/eip-4337.md
drortirosh: 1. aggregators are not a "black box": a bundler must perform off-chain sig aggregation
2. probably have a `aggregator.getInfoForAggregation(UserOp[]) view returns (bytes[])` function, to collect data for a bundle (most importantly is collecting pubkeys, but maybe more). Still, aggregator code might use other mechanism (e.g. off-chain caching)
3. the same method can be used also before simulation

eip/EIPS/eip-4337.md
```diff
@@ -30,6 +30,17 @@ This proposal takes a different approach, avoiding any adjustments to the consen
...
+
+## Definitions
+
+* **UserOperation** - a structure that describes a transaction to be sent on behalf of a user. To avoid confusion, it is named "transaction".
+  * Like a transaction, it contains "sender", "to", "calldata", "maxFeePerGas", "maxPriorityFee", "signature", "nonce"
+  * unlike transaction, it contains several other fields, desribed below
+  * also, the "nonce" and "signature" fields usage is not defined by the protocol, but by each account implementation
+* **Sender** - the account contract sending a user operation.
+* **EntryPoint** - a singleton contract to execute bundles of UserOperations. Bundlers/Clients whitelist the supported entrypoint.
+* **Aggregator** - a helper contract trusted by wallets to validate an aggregated signature. Bundlers/Clients whitelist the supported aggregators.
```
forshtat, line 43: Validator validates. Verifier verifies. Helper helps. And aggregator aggregates. 

Here, `aggregator`, however, helps. And validates. But does not aggregate. 

I suggest we try finding for a better name? 😀

eip/EIPS/eip-4337.md
```diff
@@ -58,45 +69,88 @@ The core interface of the entry point contract is as follows:
...
-    (UserOperation[] calldata ops, address payable beneficiary)
-    public;
+    (UserOperation[] calldata ops, address payable beneficiary
+     AggregatorInfo[] calldata aggregators, uint[] memory execOrder
+    ) public
 
 function simulateValidation
-    (UserOperation calldata userOp)
-    external returns (uint256 preOpGas, uint256 prefund) {
+    (UserOperation calldata userOp, aggregatedSignature)
```
forshtat: Also `aggregatedSignature` here is also an individual signature, not aggregated signature. It is not possible to check a single element of an aggregated signature as far as I understand.

eip/EIPS/eip-4337.md
```diff
@@ -58,45 +69,88 @@ The core interface of the entry point contract is as follows:
...
+* the wallet MAY pay more than this minimum, to cover future transactions
+* Note that in case the request comes through an aggregator, the signature will be empty.
+
+A Wallet that supports aggregated signature should have the interface:
+```solidity
+interface IAggregatedWallet is IWallet {
+  function getAggregator() view returns (address);
+}
+```
+* **getAggregator()** returns the aggregator this wallet supports.
```
forshtat, line 116: We had the same discussion of whether we should have `getTrustedForwarder`, `getTrustedForwarders[]` or `isForwarderTrusted` method in the past, and I think the same consideration should be done here. Do you think only allowing one 'aggregator' for a wallet is going to be ok?

contracts/EntryPoint.sol
```diff
@@ -79,16 +80,62 @@ contract EntryPoint is StakeManager {
...
+        AggregatorInfo[] memory _aggregators = aggregators;
+
+        uint index = 0;
+        // uint loopPtr = getMemoryPointer();
+        for (uint i = 0; i < _aggregators.length; i++) {
+            // restoreMemoryPointer(loopPtr);
+            AggregatorInfo memory agg = _aggregators[i];
+            uint count = agg.count;
+            UserOperation[] memory copyOfOps = copyOpsRange(agg.aggregator, ops, index, index + count);
+            agg.aggregator.validateSignatures(address(this),
```
forshtat, line 109: How do you check that this 'aggregator' is indeed the one trusted by the wallet? I see that you use the wallet's own 'aggregator' in `simulateValidation` but here the 'aggregator' address is just a transaction input.

eip/EIPS/eip-4337.md
```diff
@@ -58,45 +69,88 @@ The core interface of the entry point contract is as follows:
...
+* The wallet is guaranteed to that the EntryPoint will only call its `validateUserOp` (with no signature) after calling the aggregator's validateSignatures
+* The wallet should also support aggregator-specific getter (e.g. `getAggregationInfo()`).
+  This method should export the wallet's public-key to the aggregator, and possibly more info
+* validateUserOp shouldn't validate the signature (since it is verified by the aggregator),
+  but it MUST verify the signature field is shorter than 63 bytes
```
forshtat: Is there a way for a wallet to support both individual ERC-4337 transactions and aggregated-singature ERC-4337 transactions? Because with what you define here the wallet either accepts no signature, or must check signature, but not both.

It can call `IAggregator(this.getAggregator()).validateSignatures(address(msg.sender), [userOp], aggregatedSignature);` internally if there is a signature to avoid duplication.

Also, `shorter than 63 bytes` is super confusing - why not use zero length?

eip/EIPS/eip-4337.md
```diff
@@ -58,45 +69,88 @@ The core interface of the entry point contract is as follows:
...
+
+The core interface required by an aggregator is:
+```solidity
+interface IAggregator {
+
+  function validateSignatures(address entryPoint, UserOperation[] calldata userOps, bytes calldata signature) view external;
```
forshtat: What is the purpose of passing the `entryPoint` in here? The `verifyingContract` (in terms of the ERC712) here is actually the "aggregator". And the `UserOp` **signature** is valid regardless of who verifies it anyway, it is the wallet who should check that EntryPoint is the right one?

contracts/EntryPoint.sol
```diff
@@ -401,5 +482,33 @@ contract EntryPoint is StakeManager {
...
+
+    //restore previously saved memory pointer.
+    // all "memory" allocations since getMemoryPointer are freed, and their memory will be re-uased.
+    function restoreMemoryPointer(uint ptr) private pure {
+        assembly {
+            mstore(0x40, ptr)
+        }
+    }
+
+    function copyOpsRange(
```
forshtat: I can see that you check that passed 'aggregator' is indeed the one trusted by the wallet in the `copyOpsRange` which is a separate loop of external calls, which is kind of wasteful given there is a `validateUserOp` call down the line. 

Even the loop itself is very weird - can't we save gas by switch to a two-dimensional array of UserOps or something like that?

Also, this is a very strange place to have this call anyways - copying is not related to this check.

eip/EIPS/eip-4337.md
```diff
@@ -58,45 +69,88 @@ The core interface of the entry point contract is as follows:
...
+
 ### Required entry point contract functionality
 
 The entry point's `handleOps` function must perform the following steps (we first describe the simpler non-paymaster case). It must make two loops, the **verification loop** and the **execution loop**. In the verification loop, the `handleOps` call must perform the following steps for each `UserOperation`:
-
+* Before the first validation loop, if an aggregator is specified, the entryPoint must call its `validateSignatures(UserOperation[], aggregatedSignature)`, and revert if the validation fails
 * **Create the wallet if it does not yet exist**, using the initcode provided in the `UserOperation`. If the wallet does not exist, _and_ the initcode is empty, or the newly deployed contract address differs from `UserOperation.sender`, the call must fail.
+* in case the UserOperation contains an empty signature, the entryPoint MUST call the wallet's getAggregator, and MUST revert in case the wallet doesn't implement getAggregator or it doesn't match the given **aggregator** parameter
```
forshtat: As I have said above, I consider reverting this so that the wallet checks the aggregator address internally instead of trusting that the EntryPoint did it for him. Wallet should trust the EntryPoint not to lie about the used 'aggregator' address.

eip/EIPS/eip-4337.md
```diff
@@ -162,7 +228,9 @@ If the `UserOperation` object passes these sanity checks, the client must next r
 
 ### Simulation
 
-To simulate a `UserOperation` `op` validation, the client makes an `eth_call` with the following params:
+To simulate a `UserOperation` `op` validation, the client makes a view call to `simulateValidation(userop,aggregatedSgianture)`, with a "from" address set to all-zeros
+
+e.g. makes an `eth_call` with the following params:
```
forshtat: This low-level explainer is redundant and not updated here.

eip/EIPS/eip-4337.md
```diff
@@ -198,6 +272,10 @@ function getSenderStorage(address sender) external view returns (uint256[] memor
 During batching, the client should:
 - Exclude UserOps that access any sender address created by another UserOp on the same batch (via CREATE2 factory).
 - For each paymaster used in the batch, keep track of the balance while adding UserOps. Ensure that it has sufficient deposit to pay for all the UserOps that use it.
+- A batch can include UserOperations of "unaggregated" wallets or wallets that use a single aggregator
+  (that is, a single bundle can't include operations of multiple aggregators)
```
forshtat: I think this is no longer the case, is it?

contracts/EntryPoint.sol
drortirosh: 1. at first I wanted to "slice" the userOp array, but we can't, since it contains dynamic data, and we must copy the items one by one. Also note that we juggle 3 different layouts: the handleOp's incoming "calldata", the memory structure, and the outgoing "validateSignatures" memory (which is structured as the calldata for the called function). Even if these structures are exactly the same (in solidity), their layout differ, and requires one-by-one copy.
2. You're right, we can have a better API by an array of {userOp[], aggregator, signature}. It does add 2 zero-params to a "no aggregator" mode, but I Think that's minimal
3. the loop for call the wallets is required - the wallets are expected the EntryPoint to validate their entrypoint. otherwise, I can take some wallet and pass it with a "do-nothing aggregator" to the EntryPoint, and it will be "validated"
the other alternative is to add `validateUserOp` "aggregator" param, and the wallet will have to validate this field. This is not a replacement for the `getAggregator()` method, since at least off-chain, the bundler must be able to ask the wallet what aggregator it uses. 
Yes, tt will save the extra external call loop, so it seems good idea

eip/EIPS/eip-4337.md
drortirosh: Yes, we do need to think about naming.
My problem is that the "Aggregator" is the off-chain module that creates the aggregated sig.
This component is "de-aggregator", or "aggregatedSignatureValidator" - but I don't like such long names...

eip/EIPS/eip-4337.md
drortirosh: For the simulation, the off-chain aggregator is called with an array of single UserOp and generates an aggregated signature.
Then `simulateValidation` is called with the userOp and the aggregated signature.
Since it received an aggregated signature, it calls the aggregator.validateSignatures() with an array of this single operation.

eip/EIPS/eip-4337.md
drortirosh: yes. I do think we need just one.
This is not an arbitrary trusted caller (which you might argue  that EntryPoint is) - it is the single contract that is tied up to your public-key format.
think of it as an extension of the wallet itself, so should be an immutable member
Changing an aggregator probably also entails changing the public key itself. I think that is something you might do as part of "wallet upgrade"

eip/EIPS/eip-4337.md
drortirosh: We do have 2 modes:
1. ecrecover mode - the wallet validates the signature
2. aggregator mode - the wallet need to validate the request came through aggregator.
3. the reason for a short signature is that Vitalik raised a use-case where we need some per-userOp info for aggregation (e.g. bls multisig). the only field we can pass data in  is the signature. 
4. I used the sig length as indicator we came through aggregator (which also limits the extra data to pass in the above use-case) I think an "aggregator" param to validateUserOp is better.

eip/EIPS/eip-4337.md
drortirosh: the validation need a hash of the UserOp. we currently use hash (entrypoint, hash(userOp), block.number)
I agree this is not required. (most likely, there is a single validator on the chain. the validator can also always put its own address there)

eip/EIPS/eip-4337.md
forshtat: Well, at least for BLS, as far as I understand, signature for array of 1 is in fact an individual signature.  I don't like calling this parameter "aggregatedSignature" if it is known to be a signature of an array of 1.

eip/EIPS/eip-4337.md
forshtat: And there is no reason for some advanced multisig wallet to support more than one implementation of signatures some time in the future, in theory? Say, a Gnosis Safe who has an ECDSA, BLS and other kinds of co-signers?

eip/EIPS/eip-4337.md
forshtat: Re 1 & 2, and 4: 
I understand that you have two modes. I don't see how **single** wallet would support **BOTH** _simultaneously_. The wallet is not "told" that the "aggregator" was called. Wallet only "knows" that this happened if there is no signature. There is no way for the wallet to require signature check **conditionally**. What am I missing?
Re 4 I Totally agree, passing "aggregator" param to `validateUserOp` seems to solve my problem.

Re 3: This is not a good design IMO. We should not have a field called `signatureOrMaybeSomeDataShorterThan63BytesWeDontReallyKnow` in the UserOp struct. Is it time to add `approvalData` to the UserOp? Maybe we should rename `signature` to `approvalData` as this field can contain whatever it wants really. I did not read the Vitalik's use-case, but I assume that if there is a use-case for 63 bytes, there will be a use-case for 64+ bytes soon.

eip/EIPS/eip-4337.md
forshtat: I feel like this parameter is redundant. I am new to the project so I'm not sure, but `entryPoint` here seems unrelated.

contracts/EntryPoint.sol
yoavw: Yes, I meant wasting gas, not executing stuff.  Just griefing.  Since the aggregator has to call the wallet (e.g. to get the pubkey), it can't know the amount of gas the wallet will waste.  So maybe aggregators should put a gas limit on calls it makes to wallets.  And then they can advertise the max cost, and it can be added to preVerificationGas.

Re whitelist, in the future it doesn't have to be.  The aggregator could just be staked, and not have an off-chain signature code in the bundler, but offer a pure function in the aggregator contract, which does the off-chain work in the bundler.  Although in most cases it would make sense to optimize by implementing stuff in the bundler itself.

eip/EIPS/eip-4337.md
yoavw: Why can't an aggregator be a "black box" (before optimization)?  After the data from all the wallets is collected, why can't it use a pure function in the aggregator to aggregate it?  I can see why it'll be optimized for common aggregators, but if someone builds an experimental aggregator it could work without waiting for bundlers to add support.

eip/EIPS/eip-4337.md
drortirosh: it is an aggregated version of all userOps signatures together into a single one.

eip/EIPS/eip-4337.md
drortirosh: we define exactly two distinct modes:
1. each wallet handles its own signatures. EntryPoint trust validateUserOp to do all validations.
2. wallet trusts an external entity first to decide what signature should appear in the final userOp, and then do some aggregation magic.

in your case, if you want a wallet that supports arbitrary number of verification methods, you'd create a single aggregator to manage it.
Technically, you can even have ECDSA signature managed by an aggregator, and on-chain it is not much less efficient (one more call to aggregator.validateSignatures() to validate all ECDSA signatures..)

eip/EIPS/eip-4337.md
drortirosh: fixed with the new API

contracts/EntryPoint.sol
forshtat: Re 1: Is there a way to arrange this data in a way that would avoid this shuffling of bytes? I assume this will save a lot of gas, especially for bigger bundles as they will pay much more for memory allocation. I may be wrong though.

contracts/EntryPoint.sol
```diff
@@ -253,25 +270,26 @@ contract EntryPoint is StakeManager {
...
-    function simulateValidation(UserOperation calldata userOp, bytes calldata aggregatedSignature) external returns (uint256 preOpGas, uint256 prefund) {
-        if (aggregatedSignature.length != 0) {
-            IAggregator aggregator = IAggregator(IAggregatedWallet(userOp.getSender()).getAggregator());
-            UserOperation[] memory arr = new UserOperation[](1);
-            arr[0] = userOp;
-            aggregator.validateSignatures(address(this), arr, aggregatedSignature);
+    function simulateValidation(UserOperation calldata userOp) external returns (uint256 preOpGas, uint256 prefund, address aggregator) {
+
+        aggregator = address(0);
+        bool success = Executor.call(userOp.getSender(), 0, abi.encodeCall(IAggregatedWallet.getAggregator, ()), gasleft());
```
forshtat: This `getAggregator` optional method is barely useful in my opinion. Instead of reading it from the Wallet that may or may not have this method, and also it is an awkward way for the Bundler to discover what aggregator the UserOp uses, this method could accept `aggregator` as a parameter.
The value for the `aggregator` may originally come from the RPC call of `sendUserOperation` as a second parameter.
Or else, the Bundler itself can make a try view call to the `getAggregator` before calling the `simulateValidation`. If you are concerned with the number of view calls you can use a multicall, but there is little value in having this custom try-call implementation in the EntryPoint.

forshtat: P.S. I think regular solidity `try`/`catch` will be good enough here.

forshtat: P.S.S. This comment is a reminder of the issue with calling `simulateValidation` before the Wallet is deployed.

contracts/IAggregator.sol
```diff
@@ -8,9 +8,29 @@ import "./UserOperation.sol";
...
+   * @return sigForUserOp the value to put into the signature field of the userOp when calling handleOps.
+   *    (usually empty, unless wallet and aggregator support some kind of "multisig"
+   * @return sigForAggregation the value to pass (for all wallets) to aggregateSignatures()
+   */
+    function validateUserOpSignature(UserOperation calldata userOp) external view returns (bytes memory sigForUserOp, bytes memory sigForAggregation);
+
+    /**
+     * aggregate multiple signatures into a single value.
+     * This method is called off-chain to calculate the signature to pass with handleOps()
+     * bundler MAY use optimized custom code perform this aggregation
```
forshtat, line 34: Is it legal for an "aggregator" not to implement the aggregation on-chain in solidity and instead **require** that bundlers use native code to perform aggregation? I assume there may be a signature scheme that is not implemented yet or impossible to implement inside the EVM.

forshtat: Also, I believe the same "disclaimer" should be added to the `validateUserOpSignature`

contracts/samples/SimpleWallet.sol
```diff
@@ -114,17 +114,16 @@ contract SimpleWallet is BaseWallet {
     }
 
     /// implement template method of BaseWallet
-    function _validateSignature(UserOperation calldata userOp, bytes32 requestId) internal view virtual override {
+    function _validateSignature(UserOperation calldata userOp, bytes32 requestId, address aggregator) internal view virtual override {
+        (aggregator);
```
forshtat: `(aggregator);` what?

contracts/utils/Executor.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.5 <0.9.0;
+
+// solhint-disable no-inline-assembly
+
+library Executor {
```
forshtat: Is there a library like that in, say, OpenZeppelin collection?

eip/EIPS/eip-4337.md
```diff
@@ -98,45 +97,66 @@ The wallet
...
+* it is called by the bunder just after simulateValidation() returns the aggregator's address.
+* The method should return a replacement value for the UserOp.signature, and a value used for aggregation
+  (the trivial "split" is return "" for the sigForUserOp, and the signature itself for sigForAggregation) 
+* **aggregateSignatures()** must aggregate all "sigForAggregation" into a single value.
+* **validateSignatures()** MUST validate the aggregated signature matches for all UserOperations in the array, and revert otherwise.
+  This method is called on-chain by `handleOps()`
+
+#### Trusting aggregators and off-chain optimization
+
+The aggregators SHOULD stake just like a paymaster. Bundlers MAY throttle down and ban aggregators in case they take too much
```
forshtat, line 148: What is a point of aggregator staking?

contracts/IAggregator.sol
drortirosh: yes, it MAY leave these unimplemented (or revert). But this requires a bundler that aware of this specific signature scheme and has the required off-chain signature validation/aggregation code.

contracts/EntryPoint.sol
```diff
@@ -83,18 +94,81 @@ contract EntryPoint is StakeManager {
...
      * @param beneficiary the address to receive the fees
+     * @param aggregators list of aggregators to validate signatures. each aggregator define a count it validates from userOps
+     * @param execOrder execution order of userOps. either an empty array, or an array of all userOps indices in the correct execution order
      */
-    function handleOps(UserOperation[] calldata ops, address payable beneficiary) public {
+    function handleOps(
+        UserOperation[] calldata ops,
+        address payable beneficiary,
+        AggregatorInfo[] calldata aggregators,
+        uint[] memory execOrder
```
forshtat: Just a question, what is our current calldata compression situation?
Because, if I remember correctly, say, this `uint[] memory` will need 32 bytes per UserOp just to tell in what order to run those operations

drortirosh: That's true. RPC doesn't compress in calldata (only in storage), so even making this uint8[] wouldn't make a difference.
we _could_ make it "bytes", and manually parse it byte by byte, but I'm not sure it worth the trouble

eip/EIPS/eip-4337.md
drortirosh: aggregator can grief a bundler, e.g. by wasting cpu while calculating signature. by requiring a stake, it can be throttled down, so it can do it at most once a day or so.
(at first stage we're going to use whitelisted aggregators, so it is less important to stake)

contracts/utils/Executor.sol
drortirosh: there is, but it seems gas-innefficient.. this one is taken from Gnosis.

contracts/samples/SimpleWallet.sol
drortirosh: unused parameter. do you have a better syntasx?

contracts/EntryPoint.sol
jzaki: If an extra call is no issue, param optimisation can be outside of the EntryPoint.
Eg [this function](https://github.com/web3well/bls-wallet/blob/main/contracts/contracts/BLSExpander.sol#L93) takes some single params if shared amongst all ops, and then expands into a bundle to pass to it's equivalent of the entrypoint/aggregator.

jzaki: Passing the array of structs for aggregators is more costly than passing just one aggregator and signature. Similarly, this single agg case can be optimised in a prior call that then calls this function.

In an different, earlier (and perhaps heavier) implementation we had measured only 31 of equivalent user ops fitting in an Arbitrum block. So just a thought whether the multi-agg case is worth catering for, vs handling in a separate tx bundle.

contracts/EntryPoint.sol
```diff
@@ -103,33 +177,58 @@ contract EntryPoint is StakeManager {
...
                 contextOffset,
-                preGas - gasleft() + op.preVerificationGas
+                preGas - gasleft() + op.preVerificationGas,
+                false
             );
         }
 
+        if (aggregatorsLen > 0) {
+            /// Aggregated Sig loop: validate all aggregated signatures. revert entire transaction on revert.
+            validateAggregatedSignatures(ops, aggregators);
```
jzaki: A bad actor could construct a single griefing aggregator+wallet+userOp that fails all other validated user ops here?
The `_validatePrepayment` would pass, and test runs can pass, but then the attacker could change some state that the malicious aggregator checks, to then fail and take down a bundle.

contracts/EntryPoint.sol
```diff
@@ -208,7 +323,7 @@ contract EntryPoint is StakeManager {
     }
 
     // create the sender's contract if needed.
-    function _createSenderIfNeeded(UserOperation calldata op) internal {
+    function _createSenderIfNeeded(UserOperation calldata op) internal returns (bool){
```
jzaki: If this is reached via an aggregated signature, this might suffer a similar attack vector as `bls-wallet` (rogue key attack detected in audit). If it applies here, it can be resolved by only constructing a wallet with a single non-aggregated signature.

contracts/samples/SimpleWallet.sol
jzaki: Remove the param name?

jzaki (commented):
I've given a few comments where I could, but the goals and hence designs of this (vs [bls-wallet](https://github.com/web3well/bls-wallet)) are slightly different. Eg, emphasis on stricter gas payment considerations, vs just L2 tx cost optimisation.

There are some final fixes coming to `bls-wallet` thanks to an audit, so any fixes relevant to code imported here can also be added.

contracts/EntryPoint.sol
drortirosh: interesting, though 31 seems very small to me. We do need to calculate the overhead of supporting multiple aggregators (and reordering) to understand if it worth the effort.
Its true that in a world (network) where there is a single (dominant) aggregator , there is no need for neither aggregator array, nor reordering.

drortirosh: compression is something that was left out, and can be handled by a separate, wrapper call to EntryPoint. need to check the benefit of such contract, as both Arbitrum and Optimism already employ generic compression on calldata.

contracts/EntryPoint.sol
drortirosh: When creating the bundle, the aggregator is called off-chain (through simulateValidation) to validate the signature of each userOp separately.
Then the aggregateSignature is called, to generate the single signature.
The bundler at this point knows that the signature is good for this bundle.
If the handleOps fails in signature validation, it must be a griefing attempt by the aggregator - so it is banned.
Aggregators must either be whitelisted, or have a stake, so this griefing can't scale. It only cost the bundler some local CPU. If attackers continue to create such malicious aggregators, then the bundlers would raise the stake for aggregators.

contracts/EntryPoint.sol
drortirosh: please DM me with the attack vector.
AFAIK, a proper constructor for a wallet (which contains full signer infromation - e.g. public key, aggregator and entrypoint addresses) protects against such attack.
the attacker can "front-run" the wallet owner and create the wallet (with the real owner's public key) - but only the real owner can use it, since it is only callable via the entrypoint, after  signature validation (validateUserOp validates the aggregator address, and entrypoint calls validateSignatures, which validates this userOp's hash and signature are valid)

forshtat (dismissed):

forshtat (approved):

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
+      external;
+}
+```
+The wallet
+* MUST validate the caller is a trusted EntryPoint
+* The requestId is a hash over the userOp (except signature), entryPoint and chainId
+* MUST validate the signature is a valid signature of the `requestId`
+* MUST pay the entryPoint (caller) at least the "missingWalletFunds" (which might be zero, in case current wallet's deposit is high enough)
+* The wallet MAY pay more than this minimum, to cover future transactions (it can always issue `withdrawTo` to retrieve it)
+* Note that in case the request comes through an aggregator, the signature will be empty.
```
yoavw, line 103: Will it always be empty?  I thought the aggregator splits the signature to sigForAggregation and sigForUserOp, with the latter appearing in the on-chain UserOp.  Did we change that?

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
+    
+struct UserOpsPerAggregator {
+    UserOperation[] userOps;
+    IAggregator aggregator;
+    bytes signature;
+}
 ```
 
-The core interface required for a wallet to have is:
+When a wallet is called through an aggregator, it exposes instead:
```
yoavw, line 89: Why "instead"?  There's no other IWallet interface where we don't have "aggregator", we just make it zero if there's no aggregator.  So it must always expose this interface whether it's aggregated or not, right?

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
+* MUST pay the entryPoint (caller) at least the "missingWalletFunds" (which might be zero, in case current wallet's deposit is high enough)
+* The wallet MAY pay more than this minimum, to cover future transactions (it can always issue `withdrawTo` to retrieve it)
+* Note that in case the request comes through an aggregator, the signature will be empty.
+* The `aggregator` SHOULD be ignored for wallets that don't use an aggregator
+
+A Wallet that works with aggregated signature should have the interface:
+```solidity
+interface IAggregatedWallet is IWallet {
+  function validateUserOp
+  (UserOperation calldata userOp, bytes32 requestId, uint256 missingWalletFunds)
```
yoavw, line 110: In IWallet we require `(UserOperation calldata userOp, bytes32 requestId, address aggregator, uint256 missingWalletFunds)` but here we remove the aggregator since we have a separate `getAggregator()`. Do we require IAggregatedWallet contracts to implement two `validateUserOp` functions with different signatures?

Looking at `IWallet.sol` above, I see the signature was updated.  So I guess the error is here in the EIP, and `IAggregatedWallet` implements the same `validateUserOp` as any `IWallet` would.

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
+
+  function aggregateSignatures(bytes[] calldata sigsForAggregation) external view returns (bytes memory aggregatesSignature);
+
+  function validateSignatures(UserOperation[] calldata userOps, bytes calldata signature) view external;
+}
 ```
 
+* **validateUserOpSignature()** must validate the userOp's signature against the UserOp's hash (the same hash MUST be used for
+  **validateUserOpSignature** and later for **validateSignatures**)
+* it is called from `simulateValidation()`.
```
yoavw, line 137: Maybe mention explicitly that it's only called off-chain as a view function, not used on-chain.

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
+* The method should return a replacement value for the UserOp.signature, and a value used for aggregation
+  (the trivial "split" is return "" for the sigForUserOp, and the signature itself for sigForAggregation)
+* if the **offChainSigCheck** param is true, it should not validate the signature, and instead return **offChainSigInfo**, 
+  which is used by a companion off-chain library code to validate the signature 
+* **aggregateSignatures()** must aggregate all "sigForAggregation" into a single value.
```
yoavw, line 142: Ditto.  Mention that it's a helper function for the bundler to aggregate the signatures, and never called on-chain.

eip/EIPS/eip-4337.md
```diff
@@ -57,46 +68,125 @@ To prevent replay attacks (both cross-chain and multiple `EntryPoint` implementa
...
 In the execution loop, the `handleOps` call must perform the following steps for each `UserOperation`:
 
 * **Call the wallet with the `UserOperation`'s calldata**. It's up to the wallet to choose how to parse the calldata; an expected workflow is for the wallet to have an `execute` function that parses the remaining calldata as a series of one or more calls that the wallet should make.
-* **Refund unused gas fees** to the wallet
 
 ![](../assets/eip-4337/image1.png)
 
-Before accepting a `UserOperation`, bundlers must use an RPC method to locally simulate calling the `simulateValidation` function of the entry point, to verify that the signature is correct and the operation actually pays fees; see the [Simulation section below](#simulation) for details.
+Before accepting a `UserOperation`, bundlers should use an RPC method to locally call the `simulateValidation` function of the entry point, to verify that the signature is correct and the operation actually pays fees; see the [Simulation section below](#simulation) for details.
+A bundler SHOULD drop (and not add to the mempool) `UserOperation` that fails the validation
```
yoavw, line 181: Not just a bundler, but any node/client in the mempool.

eip/EIPS/eip-4337.md
```diff
@@ -162,50 +252,51 @@ If the `UserOperation` object passes these sanity checks, the client must next r
...
-5. No `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` results in an out-of-gas revert.
-6. Any `GAS` opcode is followed immediately by one of { `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` }.
-7. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
-8.  If `op.initcode.length != 0` , allow only one `CREATE2` opcode call, otherwise forbid `CREATE2`.
+2. The first (validateUserOp) call does not access _mutable state_ of any contract except the wallet itself and its deposit in the entry point contract. _Mutable state_ definition includes both storage and balance.
+3. The second (validatePaymasterUserOp) call does not access _mutable state_ of any contract except the paymaster itself.
+4. The third (validateUserOpSignature) view-call doesn't access _mutable state_ of any contract except the aggregator AND wallet itself.
+5. The third (validateUserOpSignature) is the first view call after the "NUMBER" marker. If that aggregator is banned, then this staticcall MUST revert immediately, as if it was called with zero gas.
```
yoavw, line 273: How does this happen in practice?  I thought the actual aggregator is not known ahead of time.  Does the bundler need to trace the execution to catch this?  IIRC we discussed passing an array of accepted aggregator during simulation, and revert if an unauthorized one is used (without calling its `validateUserOpSignature`).  Do we have a better way?

contracts/EntryPoint.sol
```diff
@@ -198,18 +276,31 @@ contract EntryPoint is StakeManager {
...
 
         UserOpInfo memory outOpInfo;
 
-        _validatePrepayment(0, userOp, outOpInfo);
+        actualAggregator = _validatePrepayment(0, userOp, outOpInfo, SIMULATE_NO_AGGREGATOR);
         prefund = outOpInfo.prefund;
         preOpGas = preGas - gasleft() + userOp.preVerificationGas;
 
+        numberMarker();
+        if (actualAggregator != address(0)) {
```
yoavw, line 301: Shouldn't we also check a whitelist of aggregators here (if provided)?  See question below in the EIP part.

contracts/IAggregator.sol
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * validate signature of a single userOp
+     * This method is called by EntryPoint.simulateUserOperation() if the wallet has an aggregator.
+     * First it validates the signature over the userOp. then it return data to be used when creating the handleOps:
+     * @param userOp the userOperation received from the user.
+     * @param offChainSigCheck if true, don't check signature, and leave it for the Bundler to use an off-chain native library.
+     * @return sigForUserOp the value to put into the signature field of the userOp when calling handleOps.
+     *    (usually empty, unless wallet and aggregator support some kind of "multisig"
+     * @return sigForAggregation the value to pass (for all wallets) to aggregateSignatures()
+     * @return offChainSigInfo in case offChainSigCheck is true, this value should be used by the off-chain signature code (e.g. it contains the sender's publickey)
```
yoavw, line 26: Why does the off-chain optimized implementation need its own return value rather than use `sigForAggregation` to give it what it needs?

yoavw (commented):
9 comments on this already merged PR :)

eip/EIPS/eip-4337.md
drortirosh: Right... Forgot to update the EIP text with that.

eip/EIPS/eip-4337.md
drortirosh: Damn. Yet another leftover from the old implementation. Will fix

contracts/IAggregator.sol
drortirosh: off-chain signature validation requires the userOp, signature AND the public-key (on-chain validation can simply calls `wallet.getPublicKey()`

eip/EIPS/eip-4337.md
drortirosh: yes, a single validateUserOp (IAggregatedWallet just inherits the validateUserOp from IWallet)

eip/EIPS/eip-4337.md
drortirosh: ok (though we need a general term for "node/bundler"

eip/EIPS/eip-4337.md
drortirosh: this flow uses opcode banning: the 1st STATICCALL after the 3rd marker IS the call to the aggregator.
the address should be validated as per the bundler's rules.
see AA-51 #112 for a whitelist params (though it allows for whitelist: blacklisting can only be done using opcode banning)

contracts/IAggregator.sol
yoavw: Got it, thanks.

eip/EIPS/eip-4337.md
yoavw: Usually it's called "client" but I find this terminology confusing in our context because client often refers to the user sending the UserOp.

PR #98, opened 2022-07-07
AA-32: enforce-checkin-gas-reports

drortirosh:


reports/gas-used-output.txt
```diff
@@ -5,47 +5,47 @@
 ·······················|······················|············|··············|·············|···············|··············
 |  Contract            ·  Method              ·  Min       ·  Max         ·  Avg        ·  # calls      ·  eur (avg)  │
```
forshtat, line 6: Captain Europe for the rescue! 

Seriously though, if we enforce it, let's use USD, or hide the column altogether if it is not filled if possible.

Also I am not sure about how useful this information is without extra context. Like, `handleOps` is more or less useless  as it may accept whatever operations we make it. What do we learn from it?

forshtat (dismissed):

reports/gas-used-output.txt
drortirosh: our current gas checks are not too useful. maybe the minimum of handleOps.
This one becomes useful once #93  is merged.

drortirosh:
part of AA-40

PR #100, opened 2022-07-12
AA-36: Support deterministic mutli-chain addresses

drortirosh:
initCode as deployer+data, instead of constructor code
- supports any deployer contract
- initCode doesn't have to include entire CREATE2 constructor code (it is only a method call to the deployer contract


contracts/EntryPoint.sol
```diff
@@ -319,34 +311,38 @@ contract EntryPoint is StakeManager {
...
-                address(create2factory),
-                salt,
-                keccak256(initCode)
-            )
-        );
+    function _createSender(bytes calldata initCode) internal returns (address sender) {
+        address initAddress = address(bytes20(initCode[0 : 20]));
+        bytes memory initCallData = initCode[20 :];
+        bool success;
+        assembly {
```
forshtat, line 332: Does this have to be in assembly? Can't it be `(success, ret) = initAddress.call(initCallData)` kind of situation?

forshtat (approved):

contracts/EntryPoint.sol
```diff
@@ -319,34 +311,38 @@ contract EntryPoint is StakeManager {
...
+            success := call(gas(), initAddress, 0, add(initCallData, 0x20), mload(initCallData), 0, 32)
```
yoavw, line 333: Calling a user-provided address address from EntryPoint (which isn't the sender) seems risky.  Wallets trust EntryPoint to only call their execution functions after they approve it in validation.  Could an attacker abuse this call to perform unauthorized operations on 3rd party wallets?

Maybe something along these lines:
* Attacker creates a batch meant for a wallet that has internal batching.  The batch contains two operations.  One is a transfer of all the wallet's assets, and the other is a call to a factory contract that will create the expected `sender` (which doesn't exist yet).
* Attacker sends a UserOp with this non-existent `sender` and with an initcode where the first 20 bytes are the victim wallet's address and the rest of the initcode is the batch above.
* `_createSenderIfNeeded` gets called, verifies that `sender` has no code yet, and calls `_createSender`.
* `_createSender` calls the victim wallet, giving it the malicious batch with the two operations.
* The wallet accepts the call because it trusts EntryPoint.  It transfers all of its assets (or even changes its owner), and then also calls the factory to generate `sender`.
* When the wallet returns, `sender` has code and the creation is considered successful.
* `sender`'s validation is called, and since it's an attacker-controlled contract, it of course succeeds.
* The rest of the op is delivered to `sender` but it doesn't matter at this point.  The attack was already successful.

I haven't checked whether this particular flow will work, but this call seems risky.  Maybe we should move it to a separate helper contract that gets deployed from EntryPoint's constructor, so that the call doesn't come directly from EntryPoint.

drortirosh: ok.

PR #105, opened 2022-08-18
AA-45 Refactor handle ops

drortirosh:


contracts/EntryPoint.sol
```diff
@@ -83,6 +83,27 @@ contract EntryPoint is StakeManager {
...
+     * @param opIndex into into the opInfo array
+     * @param userOp the userOp to execute
+     * @param opInfo the opInfo filled by validatePrepayment for this userOp.
+     * @return collected the total amount this userOp paid.
+     */
+    function _executeUserOp(uint256 opIndex, UserOperation calldata userOp, UserOpInfo memory opInfo) private returns (uint256 collected) {
+        uint256 preGas = gasleft();
+        uint256 contextOffset = opInfo.contextOffset;
+        bytes memory context;
+        assembly {context := contextOffset}
```
forshtat: Can we convert these assembly pieces into utility functions of
`function getUintMemoryPointer(bytes memory) private view returns (uint)`
and
`function getBytesFromMemoryPointer(uint point) private view returns (bytes memory)`
?
Like, Solidity won't copy the `bytes memory` to pass it into a function, right?

Not that it is necessary, just weird one-liner assembly insertions in the middle of regular function seems weird.

contracts/EntryPoint.sol
```diff
@@ -83,6 +83,27 @@ contract EntryPoint is StakeManager {
...
+
+        try this.innerHandleOp(userOp, opInfo, context) returns (uint256 _actualGasCost) {
+            collected = _actualGasCost;
+        } catch {
+            uint256 actualGas = preGas - gasleft() + opInfo.preOpGas;
```
forshtat, line 95: I understand that `preOpGas` is 'gas used before execution'?

Let me just point out that we have in this code:

- `verificationGas` which is a limit on gas for verification
- `preVerificationGas` which is an exact value for gas overhead to be added
- `preOpGas` which is a partial gas used value

I *strongly suggest* we spend some more letters in the names and call these `XxxLimit`, `XxxUsed`, `XxxOverhead` etc.

drortirosh: ok. but its a shotgun renaming. I'll add it in a separate PR.

contracts/EntryPoint.sol
```diff
@@ -139,7 +132,7 @@ contract EntryPoint is StakeManager {
     struct UserOpInfo {
         bytes32 requestId;
         uint256 prefund;
-        PaymentMode paymentMode;
+        address paymaster;
```
forshtat: `PaymentMode` is not used any more

contracts/EntryPoint.sol
```diff
@@ -191,30 +185,35 @@ contract EntryPoint is StakeManager {
...
-    function _getPaymentInfo(UserOperation calldata userOp) internal view returns (uint256 requiredPrefund, PaymentMode paymentMode) {
-        requiredPrefund = userOp.requiredPreFund();
-        if (userOp.hasPaymaster()) {
-            paymentMode = PaymentMode.paymasterDeposit;
-        } else {
-            paymentMode = PaymentMode.walletDeposit;
-        }
+    function _getPaymentInfo(UserOperation calldata userOp) internal view returns (uint256 requiredPrefund, address paymaster) {
+    unchecked {
+        paymaster = userOp.paymaster;
```
forshtat: There seems to be no point in returning `address paymaster` unmodified here.

contracts/EntryPoint.sol
```diff
@@ -191,30 +185,35 @@ contract EntryPoint is StakeManager {
...
+        //when using a Paymaster, the verificationGas is used also to as a limit for the postOp call.
+        // our security model might call postOp eventually twice
+        uint256 mul = paymaster != address(0) ? 3 : 1;
```
forshtat: I am having my doubts that implicit `postOpGasLimit := verificationGasLimit` is such a great assumption.
IIRC, wallet deployment cost is counted towards the verification gas, isn't it? 
This potentially means much bigger then expected `requiredGas`, and way too much gas available to the `postOp`

contracts/EntryPoint.sol
```diff
@@ -281,9 +280,8 @@ contract EntryPoint is StakeManager {
      * revert with proper FailedOp in case paymaster reverts.
      * decrement paymaster's deposit
      */
-    function _validatePaymasterPrepayment(uint256 opIndex, UserOperation calldata op, bytes32 requestId, uint256 requiredPreFund, uint256 gasUsedByValidateWalletPrepayment) internal returns (bytes memory context) {
+    function _validatePaymasterPrepayment(uint256 opIndex, address paymaster, UserOperation calldata op, bytes32 requestId, uint256 requiredPreFund, uint256 gasUsedByValidateWalletPrepayment) internal returns (bytes memory context) {
```
forshtat: What is the point of doing it this way?

contracts/EntryPoint.sol
forshtat: Also, I don't think there is a single function here that has `UserOpInfo` but does not have a corresponding `UserOperation`, making this field a duplicate.

contracts/EntryPoint.sol
```diff
@@ -310,24 +308,27 @@ contract EntryPoint is StakeManager {
      * also make sure total validation doesn't exceed verificationGas
      * this method is called off-chain (simulateValidation()) and on-chain (from handleOps)
      */
-    function _validatePrepayment(uint256 opIndex, UserOperation calldata userOp, bytes32 requestId) private returns (uint256 requiredPreFund, PaymentMode paymentMode, bytes memory context){
+    function _validatePrepayment(uint256 opIndex, UserOperation calldata userOp, UserOpInfo[] memory opInfos) private {
```
forshtat: Just a question - can we pass an individual `UserOpInfo` here? I feel worried that this function accepts the entire array just to set a few value in a single element. Even if gas-wise this is cheap.

contracts/EntryPoint.sol
drortirosh: yes, it is not needed: paymaster!=address(0) means "paymaster pays".
(it used to be a 3-state, but now the mode is boolean)

drortirosh: it isn't. its an in-memory collected data about a userOp (context, prefund, etc)
it also caches "paymaster" address.

contracts/EntryPoint.sol
drortirosh: I use the return value to cache into the opInfo

contracts/EntryPoint.sol
drortirosh: no, you can't.
even making the assignment as:
```
        UserOpInfo memory opInfo = opInfos[opIndex];
        opInfo = UserOpInfo(...)
```
causes a revert on geth (and interestingly, hardhat is fully stuck..)
a "reference" into an array is not assignable

contracts/EntryPoint.sol
drortirosh: avoiding reading the paymaster again from the calldata. I read it once (to know the payment mode) and then use it from memory. cheaper to pass it as parameter than reading it again from calldata. (solidity manages to cost ~200gas for each calldata member. it does a lot of useless boundary checks)

contracts/EntryPoint.sol
forshtat: I mean, you should also remove the `enum` declaration.

contracts/EntryPoint.sol
forshtat: Do you need to assign to 'reference'? It is already pointing somewhere, can't you just assign every field of a struct?

```
UserOpInfo memory opInfo = opInfos[opIndex];
opInfo.paymaster = paymaster;
opInfo.prepaid = prepaid;
...
```
?

contracts/EntryPoint.sol
drortirosh: we do need SOME limit on postOp - we need verify the paymaster has "max possible gas" to pay for it.
The options are:
- constant value
- add new `UserOp.postOpGasLimit` parameter (which is used only if there's a paymaster)
- verificationGas
- callGas
- a factor (e.g. half) of one of the above.

using verificationGas is not ideal, but better than the alternatives. The only downside is that the paymaster needs a higher balance (which is never used)

PR #112, opened 2022-09-12
AA-51: allowedAggregators simulation parameter

drortirosh:
`simulateValidation()` always reverts - successful result is error `SimulationResulut`
return aggregator address, and never calls it.
bundler should either reject UserOP with aggregator or uuuse 
`aggregator.validateUserOpSignature()` (or an equivalent native library code)


contracts/core/EntryPoint.sol
```diff
@@ -240,8 +239,34 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+    function simulateValidation(UserOperation calldata userOp)
     external returns (uint256 preOpGas, uint256 prefund, address actualAggregator, bytes memory sigForUserOp, bytes memory sigForAggregation, bytes memory offChainSigInfo) {
+        address[] memory allowedAggregators = new address[](1);
+        allowedAggregators[0] = address (0);
+        return simulateValidationWithAggregators(userOp, allowedAggregators, false);
+    }
+
+    /**
+    * Simulate a call to wallet.validateUserOp and paymaster.validatePaymasterUserOp.
+    * allow using a sighnature aggregator.
```
yoavw: sighnature => signature

contracts/core/EntryPoint.sol
```diff
@@ -240,8 +239,34 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+    * Validation succeeds if the call doesn't revert.
+    * @dev The node must also verify it doesn't use banned opcodes, and that it doesn't reference storage outside the wallet's data.
+     *      In order to split the running opcodes of the wallet (validateUserOp) from the paymaster's validatePaymasterUserOp,
+     *      it should look for the NUMBER opcode at depth=1 (which itself is a banned opcode)
+     * @param userOp the user operation to validate.
+     * @param allowedAggregators list of aggregators we specifically allow for this userOp. An empty list means "allow any".
+     *          a list of a single "address(0)" entry means "do not allow any other aggregator", effectively means blocking aggregated signatures.
+     * @param offChainSigCheck if the wallet has an aggregator, skip on-chain aggregation check. In thus case, the bundler must
```
yoavw: Why do we need this?  We do need off-chain aggregation, for optimization purposes.  But why do we need off-chain verification?  There's always an on-chain component that verifies the signature, so why not always use it in simulation?

contracts/core/EntryPoint.sol
```diff
@@ -240,8 +239,34 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+     *          perform the equivalent check using an off-chain library code
+     * @return preOpGas total gas used by validation (including contract creation)
+     * @return prefund the amount the wallet had to prefund (zero in case a paymaster pays)
+     * @return actualAggregator the aggregator used by this userOp. if a non-zero aggregator is returned, the bundler must get its params using
+     *      aggregator.
+     * @return sigForUserOp - only if has actualAggregator: this value is returned from IAggregator.validateUserOpSignature, and should be placed in the userOp.signature when creating a bundle.
+     * @return sigForAggregation  - only if has actualAggregator:  this value is returned from IAggregator.validateUserOpSignature, and should be passed to aggregator.aggregateSignatures
+     * @return offChainSigInfo - if has actualAggregator, and offChainSigCheck is true, this value should be used by the off-chain signature code (e.g. it contains the sender's publickey)
```
yoavw: Why do we need `offChainSigInfo` rather than just reuse `sigForAggregation`?  Does the off-chain aggregation implementation need different information than the on-chain aggregation code?

contracts/interfaces/IEntryPoint.sol
```diff
@@ -89,13 +94,36 @@ interface IEntryPoint is IStakeManager {
...
+     * @return sigForUserOp - only if has actualAggregator: this value is returned from IAggregator.validateUserOpSignature, and should be placed in the userOp.signature when creating a bundle.
+     * @return sigForAggregation  - only if has actualAggregator:  this value is returned from IAggregator.validateUserOpSignature, and should be passed to aggregator.aggregateSignatures
+     * @return offChainSigInfo - if has actualAggregator, and offChainSigCheck is true, this value should be used by the off-chain signature code (e.g. it contains the sender's publickey)
+     */
+    function simulateValidation(UserOperation calldata userOp)
+    external returns (uint256 preOpGas, uint256 prefund, address actualAggregator, bytes memory sigForUserOp, bytes memory sigForAggregation, bytes memory offChainSigInfo);
+
     /**
     * Simulate a call to wallet.validateUserOp and paymaster.validatePaymasterUserOp.
+    * allow using a sighnature aggregator.
```
yoavw: sighnature => signature again

contracts/core/EntryPoint.sol
```diff
@@ -225,13 +225,12 @@ contract EntryPoint is IEntryPoint, StakeManager {
```
yoavw, line 228: Question related to line 129 above (can't comment there because it hasn't changed in this PR)

Potential DoS vector on aggregators:
If the aggregated signature validation fails for any op, we catch it and revert with `SignatureValidationFailed(address(aggregator))`.  We don't give the aggregator an opportunity to report the `opIndex`.
The aggregator may access wallet storage during validation, so let's assume an aggregator that does. What if an attacker sends a valid aggregated op, then frontruns with a transaction that invalidates it by changing a storage slot in the wallet?
If someone does this to non-aggregated validation, the bundler identifies it by opIndex and drops the bad op.  But if it happens during `aggregator.validateSignatures(ops, opa.signature)`, the bundler doesn't know what to drop. It'll have to drop the aggregator itself, censoring all the valid aggregated ops.

Should we change the `validateSignatures` interface to return `(success,opIndex)` instead of just reverting?  So if it returns `success==false` we revert with `SignatureValidationFailed(address(aggregator),opIndex)`.  If the aggregator knows who the culprit, it returns it in `opIndex`.  If the signature verification fails outside the context of any specific op, it returns `opIndex=-1` to indicate that.  Or is there a better way to do this?

yoavw (commented):
Wrote some comments and questions.  And let's discuss the DoS vector I mentioned.

contracts/core/EntryPoint.sol
yoavw: Never mind, I just saw you replied this question on the other PR.

contracts/core/EntryPoint.sol
drortirosh: I don't think we can add "opIndex" to help here: BLS aggregation merely sum the signature points - no per-op info.
The bundler calls simulate (which calls wallet.validateUserOperation and aggregator.valdiateUserOpSignature, which internally calls wallet.getPublicKey) just before creating the bundle.
So the way to make this attack, the wallet has to make getPublicKey return one valid off-chain, and then different value on-chain.
both calls to getPublicKey appear in the same sequence (just after validateUserOp) and called by the same sender (aggregator)
Do you think you can detect a difference? 

The other alternative is make a better mechanism to prevent modification of public-key:

One option is to raise the public-key into the protocol, and drop the UserOp if it gets modified (currently it is a private protocol between the aggregator and wallet).

Another optioin is to move the public-key storage into the aggregator itself, and add "time-lock" mechanism.

However, both of these solutions require deeper understanding of the aggregation by the core protocol (not sure that it is that a bad thing: a "wallet-specific-publickey" or equivalent is a concept that will exist in any aggregation system.

contracts/core/EntryPoint.sol
drortirosh: we do add aggregation into the protocol, but some networks (e.g. Ethereum mainnet) will never use it. I felt uneasy to force complex API for network that only want to disable the mechanism...

yoavw: On networks where aggregation is not used, there won't be an aggregator and therefore offChainSigCheck also won't be relevant.  I'm trying to understand the use case where there *is* an aggregator but it doesn't verify the signature during simulation and requires off chain sig check.

contracts/core/EntryPoint.sol
yoavw: The way is indeed to change the result of getPublicKey, or of getAggregator.  And there's no need to detect the difference in order to perform the attack.  The attacker puts the op in mempool while it's still valid, and then invalidates it using another transaction with higher gas, which changes the result of getPublicKey.

I guess it should normally be caught by the 2nd simulation while the op is still alone, so it won't happen in a batch.  But we do lose one line of defense which we otherwise have.  opIndex is meant to catch ops that somehow slipped through the 2nd simulation.

I'm not sure about adding the complexity of moving the key into the aggregator.  And in some cases it can be a complex "key" - something that isn't just a key but multiple keys with roles, although in this case the aggregator needs to know its structure anyway since it's the one parsing it.

In the flow you described above, where aggregator.valdiateUserOpSignature internally calls wallet.getPublicKey, it would be nice if we could somehow reflect the internal state if signature verification fails.  If I understand correctly, upon signature failure it can't detect which of the public keys caused it to fail.  But if it could somehow tell the bundler what was the public key it got from each wallet, the bundler could compare it to what it saw before, and find the culprit.  I don't see an easy way to export so much information though.  If it becomes an issue, clients might need to trace the execution and catch these changes if they ever occur.

contracts/core/EntryPoint.sol
drortirosh: the scenario is:
- the network does support aggregation. There is one commonly-used aggregator, where the bundler has a native library to handle it
- the bundler probably bans any other aggregators (through userOp), as on-chain validation is heavy
- setting the "offChainSigCheck" makes the EVM code simpler (but requires the bundler to call the native library to validate the signature)
- the bundler probably uses a native library  replacement for "aggregateSignatures".

yoavw: But it means that we have two different implementations for verification.  One used during simulation (but not calling the EVM verification function) and another used on-chain.  Any difference between the two could result in on-chain DoS.  The two implementations must behave 100% identically, with no corner cases.  Not always easy to achieve.

drortirosh: - There is always an on-chain validation, called by handleOps.
- the aggregated signature is created by the bundler. This code is traditionally an optimized, off-chain library. So offchain/onchain dilemma is always there...
- I only added a "reference implementation" view method to perform these operations
- (also note that I can't even "prove" that sequence of the view function "validateSignature", and then aggregation is equivalent to the on-chain validateSignatures() method.. - it was tested for some specific scenarios)

contracts/core/EntryPoint.sol
drortirosh: - the client can pull all public-keys of the ops it put into a bundle,
- in case the execution fails (meaning, one op managed to to modify it after the 2nd validation, but before putting it on chain), the client will pull all public-keys again, and compare the lists

contracts/core/EntryPoint.sol
yoavw: Yes, there's always on-chain validation.  I was just concerned that someone exploits a difference between off-chain and on-chain signature verification, causing it to succeed off-chain and then revert in handleOps.  This actually shouldn't happen on-chain because the bundler simulates the entire handleOps during estimateGas, so the on-chain verification code will revert at that point.  But it'll be hard for the bundler to know which opIndex is responsible.

Unlike the other DoS case I mentioned, where the key changes, this one can't be detected by pulling the keys from the wallets and comparing.  The key remains the same.  Verification of one op just succeeds in simulation (with the native code implementation) and then fails in handleOps simulation for the entire bundle.  The only way to find the culprit is to simulate the ops with non-native code (creating single-op bundles and simulating handleOps for each), which negates the benefit of having native verification code to begin with.

drortirosh: 1. there is even a greater issue, since we validate specific userOp using one function, but eventually on-chain only validate the aggregated sig.
2. A node can always call manually the validateUserOpSignature() after simulationValidation returns (simply giving an empty array to allowedAggregators prevents a call to it on-chain, and then it gets the aggregator's address in the return value from simulationValidation)

contracts/core/EntryPoint.sol
```diff
@@ -227,37 +228,50 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
         prefund = outOpInfo.prefund;
         preOpGas = preGas - gasleft() + userOp.preVerificationGas;
 
         numberMarker();
-        if (actualAggregator != address(0)) {
-            (sigForUserOp, sigForAggregation, offChainSigInfo) = IAggregator(actualAggregator).validateUserOpSignature(userOp, offChainSigCheck);
+        if (aggregator != address(0)) {
+            bool found = false;
+            if (allowedAggregators.length>0) {
+                for (uint i=0; i< allowedAggregators.length; i++) {
```
forshtat: I know I am coming late to this PR, but can I get a real quick TLDR why do we need to check the `aggregator` against the list of allowed ones in Solidity? Like, if this method is called in view mode and returns an aggregator address, what difference does it make if the check if the aggregator is allowed is done off-chain? 
Yes, you run some code of a non-whitelisted aggregator in a view mode then, but how much damage can this do?
Thanks!

yoavw: Previously the simulation did run the aggregator code, so we needed a way to prevent execution of unknown/unstaked aggregators to avoid DoS (e.g. aggregators that cause a lot of work, succeed in 1st simulation and fail the 2nd one in large quantities through a single storage slot change).  With the split to two steps, and the simulation only returning the aggregator and not executing it, this is no longer needed.  The bundler can look at the returned aggregator address and decides if it wants to proceed to the next step or drop the UserOp.  I think we can remove this.

contracts/bls/BLSSignatureAggregator.sol
```diff
@@ -98,19 +119,14 @@ contract BLSSignatureAggregator is IAggregator {
...
+        uint256[4] memory pubkey = getUserOpPublicKey(userOp);
         uint256[2] memory message = userOpToMessage(userOp);
 
-        if (offChainSigCheck) {
-            offChainSigInfo = abi.encode(signature, pubkey, getRequestId(userOp));
-        } else {
-            require(BLSOpen.verifySingle(signature, pubkey, message), "BLS: wrong sig");
-        }
+        require(BLSOpen.verifySingle(signature, pubkey, message), "BLS: wrong sig");
         sigForUserOp = "";
```
forshtat: Unrelated to this PR, but what is `sigForUserOp` and why is it hard-coded to be empty string? This should at least be explained in the comment (which I did read and understood nothing)

contracts/core/EntryPoint.sol
```diff
@@ -25,8 +26,8 @@ contract EntryPoint is IEntryPoint, StakeManager {
 
     SenderCreator private immutable senderCreator = new SenderCreator();
 
-    // internal value used during simulation: need to query aggregator if wallet is created
-    address private constant SIMULATE_NO_AGGREGATOR = address(1);
+    // internal value used during simulation: need to query aggregator.
+    address private constant SIMULATE_FIND_AGGREGATOR = address(1);
```
forshtat, line 30: Unrelated but I suggest using a completely random address like `0xfefefefe...` because `address(1)` or something like that is actually used for precompiles which may be confusing.

contracts/core/EntryPoint.sol
```diff
@@ -232,32 +233,25 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
         preOpGas = preGas - gasleft() + userOp.preVerificationGas;
 
-        numberMarker();
-        if (actualAggregator != address(0)) {
-            (sigForUserOp, sigForAggregation, offChainSigInfo) = IAggregator(actualAggregator).validateUserOpSignature(userOp, offChainSigCheck);
+        if (aggregator != address(0)) {
+            revert UnverifiedSignatureAggregator(preOpGas, prefund, deadline, aggregator);
         }
-        require(msg.sender == address(0), "must be called off-chain with from=zero-addr");
+        require(tx.origin == address(0), "must be called off-chain with from=zero-addr");
```
forshtat: I think this line should be in the first line as it checks the pre-condition that the method is called from a right place and as a view call only.

drortirosh: The line is last, so that we can see other revert reasons.
in solidity, there is no side-effect for running reverts at the end (except gas usage, which we don't care about in view function)
anyway, it was removed, as we now always revert, for simplified API.

forshtat (approved):

contracts/core/EntryPoint.sol
```diff
@@ -284,12 +268,13 @@ contract EntryPoint is IEntryPoint, StakeManager {
     }
 
     /**
-     * helper: make a "view" call to calculate the sender address.
-     * must be called from zero-address.
+     * Get counterfactual sender address.
```
derekchiang, line 271: What was the rationale for returning the address in an error?

drortirosh: This "getter" has side effect (creating the wallet), so it has to revert to make sure no one attempt to call it on chain. Yes, we're rewriting it as a function returning address wrapping an internal function that always reverts...

PR #113, opened 2022-09-12
internal audit fixes

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -270,40 +272,23 @@ contract EntryPoint is IEntryPoint, StakeManager {
     }
 
     // create the sender's contract if needed.
-    function _createSenderIfNeeded(MemoryUserOp memory mUserOp, bytes calldata initCode) internal {
+    function _createSenderIfNeeded(uint opIndex, MemoryUserOp memory mUserOp, bytes calldata initCode) internal {
```
yoavw: Usually nicer to specify the int size, e.g. uint256.  (and good catch re the inconsistent revert method that didn't expose the offending opIndex)

eip/EIPS/eip-4337.md
```diff
@@ -96,28 +97,25 @@ interface IWallet {
 The wallet
 * MUST validate the caller is a trusted EntryPoint
 * The requestId is a hash over the userOp (except signature), entryPoint and chainId
-* MUST validate the signature is a valid signature of the `requestId`
+* If the wallet does not support signature aggregation, it MUST validate the signature is a valid signature of the `requestId`
```
yoavw, line 100: I'm nitpicking a bit, but it's not about whether the wallet supports aggregation in general, but whether it's using aggregation for the current UserOp. There could be a wallet that supports both ECDSA and BLS, in which case it MUST validate in non-aggregated mode (i.e. aggregator==0), but not when the UserOp is aggregated.

eip/EIPS/eip-4337.md
```diff
@@ -96,28 +97,25 @@ interface IWallet {
 The wallet
 * MUST validate the caller is a trusted EntryPoint
 * The requestId is a hash over the userOp (except signature), entryPoint and chainId
-* MUST validate the signature is a valid signature of the `requestId`
+* If the wallet does not support signature aggregation, it MUST validate the signature is a valid signature of the `requestId` 
 * MUST pay the entryPoint (caller) at least the "missingWalletFunds" (which might be zero, in case current wallet's deposit is high enough)
 * The wallet MAY pay more than this minimum, to cover future transactions (it can always issue `withdrawTo` to retrieve it)
-* Note that in case the request comes through an aggregator, the signature will be empty.
```
yoavw, line 102: MAY be empty?

eip/EIPS/eip-4337.md
```diff
@@ -96,28 +97,25 @@ interface IWallet {
...
 }
 ```
 * **getAggregator()** returns the aggregator this wallet supports.
-* **validateUserOp()** MUST verify the `aggregator` parameter is valid and the same as `getAggregator`
+* **validateUserOp()** (inherited from IWallet interface) MUST verify the `aggregator` parameter is valid and the same as `getAggregator`
 * The wallet should also support aggregator-specific getter (e.g. `getAggregationInfo()`).
   This method should export the wallet's public-key to the aggregator, and possibly more info
   (note that it is not called directly by the entryPoint) 
-* validateUserOp MAY completely ignore the signature field
+* validateUserOp SHOULD ignore the signature field
```
yoavw: Why is it SHOULD and not MAY?  Couldn't it contain config data such as the role of the signer, which `validateUserOp` could use to determine if this signer is allowed to perform the requested operation?  (I mean in wallets like Argent where there can be different roles assigned to different signers).

eip/EIPS/eip-4337.md
drortirosh: Currently, the way for a wallet to signal if it supports aggregation or not is by having getAggregator() method returning a non-zero address.
So currently, a wallet can change its state from aggregated to non-aggregated mode, but it can't arbitrarily accept both.
One way to achieve this is to change it to getAggregator(UserOperation), or at least getAggregator(signature), so that it can decide based on the signature what aggregation scheme to use.
We do need more feedback as whether this is needed.

eip/EIPS/eip-4337.md
drortirosh: the line is removed..

eip/EIPS/eip-4337.md
drortirosh: ok

PR #117, opened 2022-10-23
AA-60 validate timestamp

drortirosh:
validateUserOp to return "deadline" - last valid timestamp.


contracts/bls/BLSWallet.sol
```diff
@@ -21,9 +21,12 @@ contract BLSWallet is SimpleWallet, IBLSWallet {
         aggregator = anAggregator;
     }
 
-    function _validateSignature(UserOperation calldata userOp, bytes32 requestId, address userOpAggregator) internal override view {
+    function _validateSignature(UserOperation calldata userOp, bytes32 requestId, address userOpAggregator)
+    internal override view returns (uint deadline) {
```
yoavw: uint256?  I try to be specific about sizes.

contracts/core/EntryPoint.sol
```diff
@@ -357,7 +360,11 @@ contract EntryPoint is IEntryPoint, StakeManager {
         }
         paymasterInfo.deposit = uint112(deposit - requiredPreFund);
         uint256 gas = mUserOp.verificationGasLimit - gasUsedByValidateWalletPrepayment;
-        try IPaymaster(paymaster).validatePaymasterUserOp{gas : gas}(op, opInfo.requestId, requiredPreFund) returns (bytes memory _context){
+        try IPaymaster(paymaster).validatePaymasterUserOp{gas : gas}(op, opInfo.requestId, requiredPreFund) returns (bytes memory _context, uint256 deadline){
+            // solhint-disable-next-line not-rely-on-time
+            if (deadline != 0 && deadline < block.timestamp) {
+                revert FailedOp(opIndex, address(0), "expired");
```
yoavw: Any reason to use address(0) in this case and not attribute the expiry to the paymaster if it was a paymaster expiry?  A paymaster could abuse it to invalidate many ops by returning an inconsistent value based on a storage slot, just like other invalidations by a paymaster.

contracts/core/EntryPoint.sol
```diff
@@ -316,8 +316,11 @@ contract EntryPoint is IEntryPoint, StakeManager {
             uint256 bal = balanceOf(sender);
             missingWalletFunds = bal > requiredPrefund ? 0 : requiredPrefund - bal;
         }
-        // solhint-disable-next-line no-empty-blocks
-        try IWallet(sender).validateUserOp{gas : mUserOp.verificationGasLimit}(op, opInfo.requestId, aggregator, missingWalletFunds) {
+        try IWallet(sender).validateUserOp{gas : mUserOp.verificationGasLimit}(op, opInfo.requestId, aggregator, missingWalletFunds) returns (uint256 deadline) {
```
yoavw: `deadline` is locally defined here (and in _validatePaymasterPrepayment) and not propagated up.  How will the bundler learn what the deadline is, in order to defend itself against short-lived ops and in order to drop expired ops efficiently without an additional simulation?  The simulation must return it and the bundler must be aware of it.

contracts/test/TestExpiryWallet.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.12;
+
+import "../samples/SimpleWallet.sol";
+
+/**
+ * A test wallet, for testing expiry.
+ * add "temporary" owners, each with a deadlline time.
```
yoavw: "deadlline" (typo)

eip/EIPS/eip-4337.md
```diff
@@ -91,7 +92,7 @@ The core interface required for a wallet to have is:
 interface IWallet {
   function validateUserOp
       (UserOperation calldata userOp, bytes32 requestId, address aggregator, uint256 missingWalletFunds)
-      external;
+      external returns (uint256 deadline);
```
yoavw, line 95: Should probably explain what this return value means.

And then also return this value during simulation so that the bundler can learn this value and drop expired ops without re-simulating them.  Maybe even mention that bundlers should defend themselves against expirations that are too short and unlikely to make it into a block.  And that they should keep track and drop expired ops without simulating again.

Also, before merging the "private" version of the EIP to the public one, be sure to merge with the public one.  There were some minor changes there since we last pushed from here.

drortirosh: ok. didn't update the EIP comments yet, only the sol files..

contracts/core/EntryPoint.sol
drortirosh: it was. but then I copy-and-pasted the line...

eip/EIPS/eip-4337.md
```diff
@@ -76,7 +77,7 @@ function handleAggregatedOps(
 
 function simulateValidation
     (UserOperation calldata userOp, bool offChainSigCheck)
-    external returns (uint256 preOpGas, uint256 prefund, address actualAggregator, bytes memory sigForUserOp, bytes memory sigForAggregation, bytes memory offChainSigInfo) {
+    external returns (uint256 preOpGas, uint256 prefund, uint256 deadlne address actualAggregator, bytes memory sigForUserOp, bytes memory sigForAggregation, bytes memory offChainSigInfo) {
```
shahafn: deadline (typo)

eip/EIPS/eip-4337.md
```diff
@@ -101,6 +102,7 @@ The wallet
 * MUST pay the entryPoint (caller) at least the "missingWalletFunds" (which might be zero, in case current wallet's deposit is high enough)
 * The wallet MAY pay more than this minimum, to cover future transactions (it can always issue `withdrawTo` to retrieve it)
 * The `aggregator` SHOULD be ignored for wallets that don't use an aggregator
+* The return value `deadline` is either zero, or the last timestamp this request is deemed valid.
```
shahafn: Zero value should be explained to mean indefinitely, I think.

eip/EIPS/eip-4337.md
yoavw: And also missing , between deadline and address

PR #118, opened 2022-10-23
Clarify wallet factory behavior when the wallet already exists

derekchiang:
Currently, when a wallet already exists, it's not clear if [`entryPoint.getSenderAddress()`](https://github.com/eth-infinitism/account-abstraction/blob/699ecca1770f7e561bbc01e4e15e30ce2a50ebcb/contracts/core/EntryPoint.sol#L289-L292) will revert or not, since that depends on the internal logic of the wallet factory.  For example, if the factory is [`SimpleWalletDeployer`](https://github.com/eth-infinitism/account-abstraction/blob/699ecca1770f7e561bbc01e4e15e30ce2a50ebcb/contracts/samples/SimpleWalletDeployer.sol), `getSenderAddress()` reverts because `SimpleWalletDeployer` reverts if the wallet already exists.

I'm updating the spec to clarify that the wallet factory should return the address even if the wallet already exists, since that's more friendly to clients who don't want to keep track of deployed wallet addresses, such as the [`SimpleWallet` client](https://github.com/eth-infinitism/bundler/blob/d5071dbfb22206f013b7b28ee188afa546265897/packages/sdk/src/SimpleWalletAPI.ts).


drortirosh:
👍

PR #119, opened 2022-10-26
AA-63 remove paymaster stake value from EntryPoint

drortirosh:


contracts/interfaces/IEntryPoint.sol
```diff
@@ -100,14 +100,15 @@ interface IEntryPoint is IStakeManager {
      *          perform the equivalent check using an off-chain library code
      * @return preOpGas total gas used by validation (including contract creation)
      * @return prefund the amount the wallet had to prefund (zero in case a paymaster pays)
+     * @return paymasterStake if current UserOperation uses a paymaster, then returns the current stake of this paymaster
```
yoavw: Maybe mention that the bundler should check if it meets its minimum requirement, and drop the op if it doesn't.

eip/EIPS/eip-4337.md
```diff
@@ -184,6 +184,8 @@ We extend the entry point logic to support **paymasters** that can sponsor trans
 
 ![](../assets/eip-4337/image2.png)
 
+First, a paymaster must issue `addStake()` to lock some eth for a period of time. This period is defined by the EntryPoint. Note that the amount staked is not defined on-chain, and MUST be
+validated by nodes during the UserOperation validation phase. The staked value MUST be above 1000 USD in native tokens (and SHOULD be larger, to compensate for native token price fluctuations)
```
yoavw: I'm not sure if we can really define it that way.  Fluctuations can be a major issue for autonomous (unmanaged) paymasters.  E.g. a TokenPaymaster that remains balanced by swapping in Uniswap, and has no owner.  Someone deposited the stake once, and is no longer around.  Suddenly the native currency price plunges, and the paymaster becomes unreachable.

Therefore I don't think the stake can be denominated in USD and tracked automatically.  It has to be a value that makes sense for the current chain, but should remain constant unless there's an extreme price drop that warrants a manual change.  E.g. make it 1 eth, but adjust it manually if eth drops to $500 or reaches $5000.

We could mitigate the issue by keeping a long term reputation for paymasters rather than just the sliding scale.  If a paymaster has been active for a year with no issues, and then its stake becomes insufficient due to eth price dropping, it makes sense to let it keep running.  However, this would complicate things.  Probably safer to just have a default value denominated in the chain's native currency, and adjust it manually in extreme situations.  

At the very least, I wouldn't use the word "MUST", but advise that the network should use a value equivalent to that.

WDYT?

eip/EIPS/eip-4337.md
```diff
@@ -312,6 +313,19 @@ Exceptions to the forbidden opcodes:
 
 ### Reputation scoring and throttling/banning for paymasters
 
+#### Rationale
+
+Paymasters might influence many UserOperations in the mempool.
+To prevent abuse, we "penalize" a paymaster that causes invalidation of large number of UserOperations in the mempool.
+To prevent paymasters from "sybil-attack", we require a paymaster to stake with the system, and thus make such DoS attack very expensive.
```
yoavw, line 326: Trying an alternative wording, let me know if it's easier to understand:
A paymaster could invalidate many UserOperations in the mempool with a single state change. This opens a denial-of-service attack vector against bundlers, which would have to re-simulate all of these UserOperations in order to drop them.
To prevent such abuse, the system throttles paymasters that have a high invalidation rate.
This mitigation won't work if the paymaster can be easily redeployed at a new address, aka. "Sybil attack".
To make throttling effective, we require a paymaster to lock a stake, which it can only withdraw after a long delay.  Redeploying to a new address requires either staking an additional amount or waiting a long time.  The stake makes DoS attacks by paymasters expensive and ensures that they don't scale.
This stake is never slashed, and can be withdrawn any time (after unstake delay).

eip/EIPS/eip-4337.md
```diff
@@ -312,6 +313,19 @@ Exceptions to the forbidden opcodes:
...
+Note that this stake is never slashed, and can be withdrawn any time (after unstake delay)
+
+The stake value is not enforced on-chain, but specifically by each node while simulating a transaction.
+The stake is expected to be above PAYMASTER_STAKE_VALUE (1000 USD in native token)
```
yoavw: See comment above re 1000 USD.  We could have a suggested value but it shouldn't be pegged to USD in an automated way.  The number (denominated in native tokens) should remain steady except during extreme situations.

eip/EIPS/eip-4337.md
drortirosh: added this.

eip/EIPS/eip-4337.md
drortirosh: redefined with a table of actual native token values

contracts/core/BasePaymaster.sol
```diff
@@ -70,10 +70,10 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     /**
      * add stake for this paymaster.
      * This method can also carry eth value to add to the current stake.
-     * @param extraUnstakeDelaySec - set the stake to the entrypoint's default unstakeDelay plus this value.
+     * @param unstakeDelaySec - the unstake delay for this paymaster. Can only be increased.
      */
-    function addStake(uint32 extraUnstakeDelaySec) external payable onlyOwner {
-        entryPoint.addStake{value : msg.value}(entryPoint.unstakeDelaySec() + extraUnstakeDelaySec);
+    function addStake(uint32 unstakeDelaySec) external payable onlyOwner {
+        entryPoint.addStake{value : msg.value}(unstakeDelaySec);
```
forshtat, line 76: I think having the `addStake_and_or_increaseUnstakeDelay` method is unnecessary coupled. I think it is better of as two separate methods.

eip/EIPS/eip-4337.md
```diff
@@ -186,6 +193,8 @@ We extend the entry point logic to support **paymasters** that can sponsor trans
 
 ![](../assets/eip-4337/image2.png)
 
+First, a paymaster must issue `addStake()` to lock some eth for a period of time. This period is defined by the EntryPoint. Note that the amount staked is not defined on-chain, and MUST be
```
forshtat: What does `This period is defined by the EntryPoint.` mean here?

eip/EIPS/eip-4337.md
```diff
@@ -243,7 +251,7 @@ When a client receives a `UserOperation`, it must first run some basic sanity ch
 
 - Either the `sender` is an existing contract, or the `initCode` is not empty (but not both)
 - The `verificationGasLimit` is sufficiently low (`<= MAX_VERIFICATION_GAS`) and the `preVerificationGas` is sufficiently high (enough to pay for the calldata gas cost of serializing the `UserOperation` plus `PRE_VERIFICATION_OVERHEAD_GAS`)
-- The paymaster is either the zero address or is a contract which (i) currently has nonempty code on chain, (ii) has registered and staked, (iii) has a sufficient deposit to pay for the UserOperation, and (iv) is not currently banned.
+- The paymaster is either the zero address or is a contract which (i) currently has nonempty code on chain, (ii) has registered and staked, (iii) has a sufficient deposit to pay for the UserOperation, (iv) has sufficient stake value and (v) is not currently banned.
```
forshtat: `ii` and `iv` are synonymous. Just say `has registered and has sufficient stake`?

eip/EIPS/eip-4337.md
```diff
@@ -317,6 +324,28 @@ Exceptions to the forbidden opcodes:
    (that is, making calls is allowed, using `gasleft()` or `gas` opcode directly is forbidden)
 
 ### Reputation scoring and throttling/banning for paymasters
+#### Rationale
+
+Paymasters might influence many UserOperations in the mempool.
+To prevent abuse, we "penalize" a paymaster that causes invalidation of large number of UserOperations in the mempool.
```
forshtat: `penalize` usually means `slash` some stake, which we don't, so I guess... `Throttle`?

eip/EIPS/eip-4337.md
```diff
@@ -317,6 +324,28 @@ Exceptions to the forbidden opcodes:
...
+To prevent paymasters from "sybil-attack", we require a paymaster to stake with the system, and thus make such DoS attack very expensive.
+Note that this stake is never slashed, and can be withdrawn any time (after unstake delay)
+
+The stake value is not enforced on-chain, but specifically by each node while simulating a transaction.
+The stake is expected to be above PAYMASTER_MIN_STAKE_VALUE, and unstake delay above PAYMASTER_MIN_UNSTAKE_DELAY
+The value of PAYMASTER_MIN_UNSTAKE_DELAY is 84600 (one day)
+
+The value of PAYMASTER_MIN_STAKE_VALUE is determined per chain, as roughly 1000 USD:
```
forshtat: This codifies 5 exchange rates as-of-this-writing as part of the ERC.
This does not specify a mechanism to change the values, which will be required to keep the peg.
This does not specify a mechanism to add new `min stake` values to the list of supported alt-chains.
This does not provide a rationale for the `1000 USD` number which seems to be pulled out of a hat.
And I think an **_E_**-RC (Ethereum something something) cannot (and shouldn't) specify parameters for chains other than mainnet for now. And sure not list a subset of such networks by name.

src/Create2Factory.ts
```diff
@@ -54,7 +54,7 @@ export class Create2Factory {
         21000
 
       // deployer requires some extra gas
-      gasLimit = Math.floor(gasLimit * 64 / 63)
+      gasLimit = Math.floor(gasLimit * 65 / 63)
```
forshtat: ??

test/entrypoint.test.ts
```diff
@@ -102,11 +102,11 @@ describe('EntryPoint', function () {
...
 
     describe('without stake', () => {
-      it('should fail to stake too little value', async () => {
-        await expect(entryPoint.addStake(2, { value: ONE_ETH })).to.revertedWith('stake value too low')
+      it('should fail to stake without value', async () => {
+        await expect(entryPoint.addStake(2)).to.revertedWith('no stake specified')
       })
-      it('should fail to stake too little delay', async () => {
-        await expect(entryPoint.addStake(1)).to.revertedWith('stake delay too low')
+      it('should fail to stake with delay', async () => {
```
forshtat: without

forshtat (changes requested):

eip/EIPS/eip-4337.md
drortirosh: do you have a better suggestion?
we do need to specify some values. at first, I put a $1000 value, saying you need "equivalent". But this force the bundler to contact some exchange to keep it up to date, which is a burden we don't want.
Worse, it means a valid permissionless paymaster today might become under-staked in the future. So instead we thought we can use fixed native values, and only change them on huge price changes.

contracts/core/BasePaymaster.sol
drortirosh: It is a single method to set the stake values.
In most cases, the "modify" capability will not be used - but it is annoying to require 2 separate calls.

src/Create2Factory.ts
drortirosh: have no idea why it broke. The difference is now I don't pass a constructor param. I understand that this Create2Factory is problematic (it REQUIRES manual gas and not "estimate", since it doesn't revert on failure). It is now used only in tests. should switch to Arachnid's deterministic deployer

eip/EIPS/eip-4337.md
yoavw: I'm also not a fan of setting it to a fixed USD denominated value.  I think it should be a network-wide value (or a mempool-wide value since a chain could have more than one mempool), which reflects the value and use cases of that network.  E.g. a gaming chain where censorship and frontrunning is not that much of an issue might select a lower value.

Maybe the ERC itself is not the right place to set the values for the different networks, but we do need one place that reflects the consensus for each network.  My current thinking is that the bundler specs and compliance tests suite is the right place for this, since all bundlers should (hopefully) run these tests to ensure compatibility.

If it's in the bundler specs/tests, anyone is free to send a PR to change it, but hopefully it'll involve some process in each network, where their community (or owner - in some networks) decides on it.

The question is what to write about it here in the ERC.  "A chain-specific amount agreed by the mempool nodes, sufficient to mitigate DoS and censorship attacks?"  - the problem is that determining a sufficient amount requires understanding the attack and its economics, which is a bit long to add here.

src/Create2Factory.ts
drortirosh: I'm not trying to fix it now (the alternative is to put a fixed gas on the tests that use it.

eip/EIPS/eip-4337.md
drortirosh: removed actual values. still we need to find a place to communicate them.

yoavw: Yes.  In the bundler specs, reference bundler, and bundlers test suite.  Any bundler that passes these tests can be used by paymaster developers to ensure that the paymaster is sufficiently staked.

Bundlers should also communicate it to users/devs somehow.  I suggest a `getMinimumPaymasterStake(chainid)` RPC which returns the minimum stake value and unstake time for that chain.

drortirosh: will add to the RPC pr..

PR #121, opened 2022-11-02
AA-67 relax storage rules in opcode banning

drortirosh:
allow wallet to access wallet-specific storage in other contracts.


eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,27 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
-Since the wallet is allowed to access its own entry point deposit in order to top it up when needed, the client must know the storage slot in order to whitelist it. The entry point therefore implements the following view function:
-
-```c++
-function getSenderStorage(address sender) external view returns (uint256[] memory senderStorageCells)
-```
+2. The first (validateUserOp) call is allowed to access storage slot of the wallet itself (see Storage access by Slots, below), 
+   but only on contracts that are not part of the current bundle (that is, not in a paymaster or any other wallet)
+   Note that the wallet is allowed (and expected) to deposit to its own balance in the EntryPoint. That is allowed by this rule.
+3. The second (validatePaymasterUserOp), the paymaster may access storage slots of itself AND of the wallet, above.
+   A bundler MAY whitelist specific paymaster addresses, and not enforce this storage access limitation.
```
yoavw: It individual bundlers whitelist specific paymasters, they become incompatible with the public mempool  These operations won't get propagated by other participants, and if the bundler attempts to propagate it, other peers will disconnect it.

We do intend to support such whitelisting, but as part of a separate mempool where everyone agrees on the whitelist.  (or a standalone bundler where sizeof(mempool)==1)

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,27 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
 While simulating `userOp` validation, the client should make sure that:
 
 1. Neither call's execution trace invokes any **forbidden opcodes**
-2. The first (validateUserOp) call does not access _mutable state_ of any contract except the wallet itself and its deposit in the entry point contract. _Mutable state_ definition includes both storage and balance.
-3. The second (validatePaymasterUserOp) call does not access _mutable state_ of any contract except the paymaster itself. The paymaster is also not allowed to make any state-change calls (SSTORE)
-   A bundler MAY whitelist specific paymaster addresses, and not enforce the above storage limitations
-4. The third (validateUserOpSignature) view-call doesn't access _mutable state_ of any contract except the aggregator AND wallet itself.
-5. The third (validateUserOpSignature) is the first view call after the "NUMBER" marker. If that aggregator is banned, then this staticcall MUST revert immediately, as if it was called with zero gas.
```
yoavw: I see this rule was removed, and is unrelated to the storage rule changes.  Is that intentional because it's no longer needed due to the aggregation API changes, or was it removed here by mistake?

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,27 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
+5. Any `CALL` or `CALLCODE` during validation has `value=0`, except for the transfer from the wallet to the entry point.
+6. No `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` results in an out-of-gas revert.
+7. No referenced address (`CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL`, `EXTCODESIZE`, `EXTCODEHASH` to addresses with `EXTCODESIZE=0`.
+8. Any `GAS` opcode is followed immediately by one of { `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` }.
+9. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
+10. If `op.initcode.length != 0` , allow only one `CREATE2` opcode call (in the validateUserOp block), otherwise forbid `CREATE2`.
+
+#### Storage access by slots
+
+A contract `A` may reference storage of
```
yoavw: During validation, ...
(yes, it should be clear from the context, but let's make it clear to avoid confusion)

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,27 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
+1. Slots of contract `A` address itself. 
+2. Slot `A` on any other address. 
+3. Slots of type `keccak256(A || X)` on any other address. (to cover `mapping(address => value)`, which is usually used for balance in ERC20 tokens). 
+4. Slots of type `keccak256(X || OWN)` on any other address, where OWN is some slot of the previous (third) type
```
yoavw: `OWN` should be recursive.  So not just the previous (third) type but also this one.

I'd define a recursion, such that any address created by a keccak that includes a previously allowed address, is also allowed.

Question: should we limit the size of the `OWN` whitelist?  At the cost of keccak, a validation could waste a few tens of kilobytes in memory by keccak(ctr++ || X).  Is that even an issue or can we ignore it?

eip/EIPS/eip-4337.md
drortirosh: do we need to complicate the EIP with these alternate mempool? maybe its better to remove that clause, and document the "non-standard" running more elsewhere.

yoavw: We shouldn't complicate the EIP, but we can't contradict outselves and suggest that a bundler MAY change the rules, because this bundler becomes incompatible with the primary mempool.

I don't want to complicate the EIP with the whole design of an alternative mempool.  Just to mention in a single paragraph that it'll be possible to have additional mempools where paymasters can be whitelisted, through agreement between the participants of that mempool.  And explicitly say that this is beyond the scope of the EIP, and that such mempools should be audited separately.

drortirosh: ok. adding separate "alternate mempool" section..

eip/EIPS/eip-4337.md
drortirosh: added "recursive", and a limit of 30 entries (no real reason for this number. any suggestion?
should be large enough above all normal cases, and small enough to prevent DoS

yoavw: 30 is too low, considering the modules use case we discussed with Gnosis Safe yesterday.  I'm now unsure whether we should limit it to a fairly high number or not limit it at all (and let the verification gas limit take care of it, which could amount to a few tens of kilobytes in memory).

The use case we discussed is a module (a separate contract) called during validation, which is using its own storage (doesn't delegatecall to the wallet).  Such module can't work as-is, but could be "ported" to 4337 by moving all of its storage to a mapping, in one of two ways:
1. Convert every variable to a mapping.  So `uint256 bla` becomes `mapping(address=>uint256) bla` and is accessed as `bla[msg.sender]` instead of `bla`.
2. Create a struct `Vars` with all the variables and have a `mapping(address=>Vars) vars`.  So `uint256 bla` is now accessed as `vars[msg.sender].bla`.

It enables converting any contract to something that can be used safely during validation, as long as it doesn't use banned opcodes and stays within the verification gas limit.

To support this method, the limit should be high or non-existent.

Am I right that the maximum memory that can be wasted with a reasonable verification gas limit capped at just tens of KB?

eip/EIPS/eip-4337.md
```diff
@@ -252,6 +252,15 @@ If the `UserOperation` object passes these sanity checks, the client must next r
 
 ### Simulation
 
+#### Rationale
+
+In order to add a UserOperation into the mempool (and later to add it into a bundle) we need to "simulate" it make sure it is valid, and that it is capable of paying for its own execution.
+In addition, we need to verify that the same will hold true when executed on-chain.
+For this purpose, a UserOperation is not allowed to access any information that might change between simulation and execution, such as current block time, number, hash etc.
+In addition, a UserOperation is only allowed to reference data related to this sender address: Multiple UserOperations should not reference the same storage, so that it is impossible to invalidate a large number of UserOperations with a single state change.
```
forshtat: `reference` is an overloaded word. Why not say `access` or `read`? Also, I would specify that we are talking about "storage slots".

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,39 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
+7. No referenced address (`CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL`, `EXTCODESIZE`, `EXTCODEHASH` to addresses with `EXTCODESIZE=0`.
+8. Any `GAS` opcode is followed immediately by one of { `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` }.
+9. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
+10. If `op.initcode.length != 0` , allow only one `CREATE2` opcode call (in the validateUserOp block), otherwise forbid `CREATE2`.
+
+#### Storage access by slots
+
+During validation, a contract `A` may reference storage of
+1. Slots of contract `A` address itself. 
+2. Slot `A` on any other address.
```
forshtat, line 302: What is `slot A`?

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,39 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
+
+#### Storage access by slots
+
+During validation, a contract `A` may reference storage of
+1. Slots of contract `A` address itself. 
+2. Slot `A` on any other address. 
+3. Slots of type `keccak256(A || X)` on any other address. (to cover `mapping(address => value)`, which is usually used for balance in ERC20 tokens). 
+4. Slots of type `keccak256(X || OWN)` on any other address, where OWN is some slot of the previous (third) type, or recursively of this (forth) type
+  (to cover `mapping(address ⇒ mapping(address ⇒ uint256)`) that are usually used for `allowances` in ERC20 tokens).
+NOTE: the total slots per wallet is limited to 30. above that, the UserOperation is rejected
```
forshtat: Why introduce a separate limit for storage access count? Why set it to `30`?
The slot is either valid or not, and there is gas limit.

eip/EIPS/eip-4337.md
```diff
@@ -271,23 +280,39 @@ In order to distinguish between them, there is a call to the NUMBER opcode (`blo
...
+8. Any `GAS` opcode is followed immediately by one of { `CALL`, `DELEGATECALL`, `CALLCODE`, `STATICCALL` }.
+9. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
+10. If `op.initcode.length != 0` , allow only one `CREATE2` opcode call (in the validateUserOp block), otherwise forbid `CREATE2`.
+
+#### Storage access by slots
+
+During validation, a contract `A` may reference storage of
+1. Slots of contract `A` address itself. 
+2. Slot `A` on any other address. 
+3. Slots of type `keccak256(A || X)` on any other address. (to cover `mapping(address => value)`, which is usually used for balance in ERC20 tokens).
```
forshtat, line 303: How does the bundler know the slot is accessed this way? Is there a specific pattern of opcodes to detect? If so, do we become reliant on Solidity implementation details?

forshtat (changes requested):

eip/EIPS/eip-4337.md
yoavw: Yes.  See what I wrote about it in my previous review: https://github.com/eth-infinitism/account-abstraction/pull/121#discussion_r1017117460

If my back-of-an-envelope calculation is correct (please verify it), then gas limit should be enough.  An additional 50KB memory used during validation is unlikely to be an issue.  If it were in the order of MBs I'd think differently.  But in any case 30 is too low and prevents the use case I described above.

eip/EIPS/eip-4337.md
yoavw: It does rely on the solidity pattern of using keccak to handle mapping, although it could be generalized a bit to whitelist (recursively) any keccak result that included `A` (or a previously whitelisted address) in its input, so it would cover `keccak(A||X)` but also `keccak(X||A||Const)` or whatever.

Keep in mind that these rules are implemented by bundlers, not enforced on-chain by EntryPoint.  The community can always add more methods, so when EVM gets a new precompile to support `shinyNewHash` and then the `newShinyLanguage` compiler determines slots address using `newShinyHash(🌞||A||X)`, we just add it to the specs and bundlers can add it to their implementation.

eip/EIPS/eip-4337.md
yoavw: PUSH32 value
ADDRESS
SLOAD

Not a common pattern, but safe from our perspective.  Anything that has the address or a keccak that had it as an input, is fine.

eip/EIPS/eip-4337.md
drortirosh: validateUserOpSignature is removed. its true that from another PR...

eip/EIPS/eip-4337.md
drortirosh: removed the limit.

eip/EIPS/eip-4337.md
drortirosh: these 2 keccak rules are the way currently solidity (and I think vyper?) use mappings. 
in theory, we should block any `keccak( * || A || *)` and `keccak( * || OWN || * )`.
These are harder to implement, and we'll add them in case there is a language that actually uses a more complex reference-address-inside-a-keccak format.

yoavw: That's fine, but I'm curious - why is it harder to implement?  We know the memory range which keccak is going to consume.  Why can't we mark the entire keccak operation as "safe" if any word-aligned offset contains `A` or `OWN`, and then whitelist the result?  Is this implementation really more complicated?

drortirosh: yes it is possible. scanning the memory block (can be as you suggest, in word-aligned jumps), and each slot verify against the hashset we keep of valid values (A, and collected "OWN" values) it is a loop for each keccak, but a small one (currently the block size is 64 (2 words). can be limit it to <5 words. each extra word is needed to define a new hash dimension, and EVM does not support dynamic languages with run-time dimensions...

PR #122, opened 2022-11-02
fix deposit withdrawal

andrewwahid:
Fix a vulnerability where any withdrawn deposits are not actually reflected in the deposit info storage.
The vulnerability allows any wallet or paymaster to withdraw all funds deposited in an Entrypoint contract.


drortirosh:
can you please also add a test for it?
also, you need to rebase it latest main, and update gas calc with
`arn ci-gas-calc`

livingrockrises:
I made this change and deployed EntryPoint and Paymaster on goerli. calling withdrawTo still does not update getDeposit state from paymaster. 

we're checking paymaster code as well but it just calls entryPoint.withdrawTo(withdrawAddress, amount); for this part

drortirosh:
we want to add this fix, but please add a test for it.

PR #123, opened 2022-11-03
for loops can be gas optimised

livingrockrises:
using ++i costs less gas per iteration than += (in for loops)
added unchecked which can reduce gas usage on contracts where underflow/overflow is unrealistic


drortirosh:
please recheck.
when I tested it locally (after applying the "via-ir" optimization) this modification improve some cases by ~60 gas, but worsen others, again by fractional value.
I would like to improve gas usage, but not by such small values, especially not when it is inconclusive, and makes the code less readable.

PR #124, opened 2022-11-03
minor refactor : remove unused interface from entry point

livingrockrises:
ICreate2Deployer is not used anymore in entry point as initcode now contains factory and data.


drortirosh:
thank you. please update the gas calculations, so all CI tests pass correctly 
(its very strange how such a benign change affect solidity's code generation..)
please run `yarn ci-gas-calc`
That script runs the gas-calc tests against a local "geth".
If you have local geth installed, you can run it on localhost:8545. If not, that's ok, as the above script will start a docker image.

livingrockrises:
updated gas calculations

livingrockrises:
Note: unless i use local ganache or geth, docker image fails with below

`localgeth  | ERROR[11-03|12:57:51.727] Failed to check db for legacy receipts   err="this operation is not supported"`

`gascalc    |   0 passing (3s)
gascalc    |   2 failing
gascalc    | 
gascalc    |   1) gas calculations
gascalc    |        warmup:
gascalc    |      Error: fatal: failed to deploy Eip2470factory
gascalc    |       at Create2Factory.deployFactory (src/Create2Factory.ts:102:13)`

`gascalc    |   2) "after all" hook in "{root}":
gascalc    |      TypeError: Cannot read properties of undefined (reading 'doneTable')
gascalc    |       at Context.<anonymous> (gascalc/GasChecker.ts:356:26)`

PR #126, opened 2022-11-03
Minor refactor : remove unused interface

livingrockrises:
ICreate2Deployer in entry point is not used anymore as initcode now contains factory and data.


drortirosh:
please re-check gas calc. afaik, it shouldn't be modified, but I can't merge it while it fails.

drortirosh:
thats, but merged through another PR.

PR #132, opened 2022-11-10
AA-68 rpc calls

drortirosh:
better definition of the required RPC methods:

* eth_sendUserOperation
* eth_estimateUserOperationGas
* eth_getUserOperationReceipt
* eth_supportedEntryPoints


eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
+* **aggregator** if set, the signature-aggregator used by the wallet.
 
-The entry point-based approach allows for a clean separation between verification and execution, and keeps wallets' logic simple. The alternative would be to require wallets to follow a template where they first self-call to verify and then self-call to execute (so that the execution is sandboxed and cannot cause the fee payment to revert); template-based approaches were rejected due to being harder to implement, as existing code compilation and verification tooling is not designed around template verification.
+**Error Codes:**: just like `eth_sendUserOperation`
 
-### Paymasters
+### * eth_getUserOperationTransactionByHash
 
-Paymasters facilitate transaction sponsorship, allowing third-party-designed mechanisms to pay for transactions. Many of these mechanisms _could_ be done by having the paymaster wrap a `UserOperation` with their own, but there are some important fundamental limitations to that approach:
+Return a transaction information based on a UserOperation hash (requestId) returned by `eth_sendUserOperation`
```
drortirosh: TBD: should we use the term `requestId` ? it is badly named, as we don't call it "request" anywhere, and it is actually UserOpHash

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+
+The entry point-based approach allows for a clean separation between verification and execution, and keeps wallets' logic simple. The alternative would be to require wallets to follow a template where they first self-call to verify and then self-call to execute (so that the execution is sandboxed and cannot cause the fee payment to revert); template-based approaches were rejected due to being harder to implement, as existing code compilation and verification tooling is not designed around template verification.
```
forshtat: This entire paragraph is convoluted and redundant.

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+
+### Paymasters
+
+Paymasters facilitate transaction sponsorship, allowing third-party-designed mechanisms to pay for transactions. Many of these mechanisms _could_ be done by having the paymaster wrap a `UserOperation` with their own, but there are some important fundamental limitations to that approach:
```
forshtat, line 414: Please do not go into details of alternative approaches in the ERC that discusses the current one.

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+
+* No possibility for "passive" paymasters (eg. that accept fees in some ERC-20 token at an exchange rate pulled from an on-chain DEX)
```
forshtat, line 416: This is redundant.

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+* Paymasters run the risk of getting griefed, as users could send ops that appear to pay the paymaster but then change their behavior after a block
```
forshtat, line 417: This is redundant.

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
 
-`eth_sendUserOperation`
+The paymaster scheme allows a contract to passively pay on users' behalf under arbitrary conditions. It even allows ERC-20 token paymasters to secure a guarantee that they would only need to pay if the user pays them: the paymaster contract can check that there is sufficient approved ERC-20 balance in the `validatePaymasterUserOp` method, and then extract it with `transferFrom` in the `postOp` call; if the op itself transfers out or de-approves too much of the ERC-20s, the inner `postOp` will fail and revert the execution and the outer `postOp` can extract payment (note that because of storage access restrictions the ERC-20 would need to be a wrapper defined within the paymaster itself).
```
forshtat, line 419: You have introduced a PR for the Relaxed Slot Access AA-67 here (https://github.com/eth-infinitism/account-abstraction/pull/121) that should allow accessing the ERC-20 tokens unmodified, am I missing something?

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+### First-time wallet creation
 
-The result `SHOULD` be set to true if and only if the request passed simulation and was accepted in the client's User Operation pool. If the validation, simulation, or User Operation pool inclusion fails, `result` `SHOULD NOT` be returned. Rather, the client `SHOULD` return the failure reason.
+It is an important design goal of this proposal to replicate the key property of EOAs that users do not need to perform some custom action or rely on an existing user to create their wallet; they can simply generate an address locally and immediately start accepting funds.
 
+The wallet creation itself is done by a "factory" contract, with wallet-specific data.
+The factory is expected to use CREATE2 (not CREATE) to create the wallet, so that the order of creation of wallets doesn't interfere with the generated addresses.
+The `initCode` field (if non-zero length) is parsed as a 20-byte address, followed by "calldata" to pass to this address.
+This method call is expected to create a wallet and return its address.
```
forshtat: `is expected` is a very soft statement for an ERC. Why not say `must`?

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+If the factory does use CREATE2 or some other deterministic method to create the wallet, it's expected to return the wallet address even if the wallet has already been created.  This is to make it easier for clients to query the address without knowing if the wallet has already been deployed, by simulating a call to `entryPoint.getSenderAddress()`, which calls the factory under the hood.
+When `initCode` is specified, if either the `sender` address points to an existing contract, or (after calling the initCode) the `sender` address still does not exist,
+then the operation is aborted.
+The `initCode` MUST NOT be called directly from the entryPoint, but from another address.
```
forshtat, line 432: You control the EntryPoint source code, what do you mean `MUST NOT`?

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+
+Wallets are encouraged to be DELEGATECALL forwarding contracts for gas efficiency and to allow wallet upgradability. The wallet code is expected to hard-code the entry point into their code for gas efficiency. If a new entry point is introduced, whether to add new functionality, improve gas efficiency, or fix a critical security bug, users can self-call to replace their wallet's code address with a new code address containing code that points to a new entry point. During an upgrade process, it's expected that two mempools will run in parallel.
+
+## RPC methods
+
+### * eth_sendUserOperation
+
+eth_sendUserOperation submits a User Operation object to the User Operation pool of the client. The client MUST validate the UserOperation, and return a result accordingly.
+
+The result `SHOULD` be set to the **requestId** if and only if the request passed simulation and was accepted in the client's User Operation pool. If the validation, simulation, or User Operation pool inclusion fails, `result` `SHOULD NOT` be returned. Rather, the client `SHOULD` return the failure reason.
```
forshtat: I think `requestId` is a confusing name as the struct is no longer called `RelayReqeust` but is a `UserOperation`, so maybe rename it to `userOpID` project-wide?

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
+#### Example failure responses:
+```json
 {
   "jsonrpc": "2.0",
   "id": 1,
-  "method": "eth_supportedEntryPoints",
-  "params": []
+  "error": {
+    "message": "Wallet didn't pay prefund",
+    "code": -33000
```
forshtat: Codes don't match. Just saying.

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
+Simulate the validation of the user operation.
 
-In this proposal, we expect wallets to have a `validateUserOp` method that takes as input a `UserOperation`, and verify the signature and pay the fee. This method is required to be almost-pure: it is only allowed to access the storage of the wallet itself, cannot use environment opcodes (eg. `TIMESTAMP`), and can only edit the storage of the wallet, and can also send out ETH (needed to pay the entry point). The method is gas-limited by the `verificationGasLimit` of the `UserOperation`; nodes can choose to reject operations whose `verificationGasLimit` is too high. These restrictions allow miners and network nodes to simulate the verification step locally, and be confident that the result will match the result when the operation actually gets included into a block.
+**Parameters**: just like `eth_sendUserOperation`
```
forshtat: Copy-paste here?

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
+**Return Value**:
+* **preOpGas** gas used by validation (including preVerificationGas)
+* **prefund** amount the wallet need to pre-pay (above preOpGas and callGasLimit)
+* **deadline** if set, until what time this userOp is valid
+* **paymasterInfo** if set, current paymasterStake and paymasterUnstakeDelay
+* **aggregator** if set, the signature-aggregator used by the wallet.
 
-The entry point-based approach allows for a clean separation between verification and execution, and keeps wallets' logic simple. The alternative would be to require wallets to follow a template where they first self-call to verify and then self-call to execute (so that the execution is sandboxed and cannot cause the fee payment to revert); template-based approaches were rejected due to being harder to implement, as existing code compilation and verification tooling is not designed around template verification.
+**Error Codes:**: just like `eth_sendUserOperation`
```
forshtat: No need to mention that, I believe

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
-* No possibility for "passive" paymasters (eg. that accept fees in some ERC-20 token at an exchange rate pulled from an on-chain DEX)
-* Paymasters run the risk of getting griefed, as users could send ops that appear to pay the paymaster but then change their behavior after a block
+An equivalent of `eth_getTransactionByHash`
 
-The paymaster scheme allows a contract to passively pay on users' behalf under arbitrary conditions. It even allows ERC-20 token paymasters to secure a guarantee that they would only need to pay if the user pays them: the paymaster contract can check that there is sufficient approved ERC-20 balance in the `validatePaymasterUserOp` method, and then extract it with `transferFrom` in the `postOp` call; if the op itself transfers out or de-approves too much of the ERC-20s, the inner `postOp` will fail and revert the execution and the outer `postOp` can extract payment (note that because of storage access restrictions the ERC-20 would need to be a wrapper defined within the paymaster itself).
+**Parameters**
+* **hash** a requestId value returned by `eth_sendUserOperation`
 
-### First-time wallet creation
+**Return value**:
```
forshtat: I suggest creating a wrapped object that has original UserOperation, EntryPoint transaction, `requestId` and other relevant information here. Otherwise, this call may end up being quite useless.

forshtat (changes requested):
Overall, please give the text a check on Grammarly. Thanks!

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
 
 To help make sense of these params, note that a malicious paymaster can at most cause the network (only the p2p network, not the blockchain) to process `BAN_SLACK * MIN_INCLUSION_RATE_DENOMINATOR / 24` non-paying ops per hour.
 
-### RPC methods
+## Rationale
+
+The main challenge with a purely smart contract wallet based account abstraction system is DoS safety: how can a miner including an operation make sure that it will actually pay fees, without having to first execute the entire operation? Requiring the miner to execute the entire operation opens a DoS attack vector, as an attacker could easily send many operations that pretend to pay a fee but then revert at the last moment after a long execution. Similarly, to prevent attackers from cheaply clogging the mempool, nodes in the P2P network need to check if an operation will pay a fee before they are willing to forward it.
```
yoavw: "miner" -> "block builder" (everywhere, not just in this line)

eip/EIPS/eip-4337.md
yoavw: With the new rule in AA-67 we can access the user's balance in 3rd party ERC20 contracts, as long as the `balanceOf` function of that token doesn't access other storage slots not associated with the user's address.  However, this doesn't make TokenPaymaster completely safe because it can't check the token's exchange rate on Uniswap or anything like it.  It could use different strategies to protect itself:
* If it's a frequently used paymaster, it can save a copy of the price during `postOp`, and in each validation it requires the user to have at least 2x the required amount based on the saved price, to account for price fluctuations.
* It could be a DepositPaymaster, requiring the user to lock a security deposit once, and ask the user to provide the Uniswap price of the token as part of the paymasterData.  If the paymaster discovers (during `postOp` that the user lied and doesn't have sufficient balance+allowance, it forfeits the security deposit to cover the damage.

eip/EIPS/eip-4337.md
yoavw: The EIP defines what EntryPoint should do.  I.e. if someone else were to implement EntryPoint, they'd have to follow this rule in order to protect their EntryPoint.  (Otherwise if initCode runs with msg.sender==EntryPoint then it could tell any wallet to do anything, without calling its validation first).

eip/EIPS/eip-4337.md
```diff
@@ -360,16 +360,76 @@ Stated in simpler terms, we expect at least `1 / MIN_INCLUSION_RATE_DENOMINATOR`
...
+2. **EntryPoint** the entrypoint address the request should be sent through. this MUST be one of the entry points returned by the `supportedEntryPoints` rpc call.
+
+#### Return value:
+* If the UserOperation is valid, the client MUST return the calculated **requestId** for it
+* in case of failure, MUST return an `error` result object, with `code` and `message`. The error code and message SHOULD be set as follows:
+  * **code: -32602** - invalid UserOperation struct/fields
+  * **code: -32500** - transaction rejected by entryPoint's simulateValidation.
+    * The `message` field SHOULD be set to the FailOp error message
+    * The `data` field SHOULD contain a `paymaster` value, if set by this FailedOp
+  * **code: -32501** - transaction rejected because paymaster is throttled/banned
```
yoavw: Missing a code for "paymaster is not sufficiently staked".  This one is for throttled/banned.  In the example below you use -33000 but it doesn't appear here.

eip/EIPS/eip-4337.md
yoavw: Right.  Is this one a -32500 (failed simulation)?  Or a separate code that we don't currently have?

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
   "jsonrpc": "2.0",
   "id": 1,
-  "result": [
-    "0xcd01C8aa8995A59eB7B2627E69b40e0524B5ecf8",
-    "0x7A0A0d159218E6a2f407B99173A2b12A6DDfC2a6"
-  ]
+  "error": {
+    "message": "not staked",
+    "paymaster": "0x123456789012345678901234567890123456790",
+    "code": -33000
```
yoavw: -33000 is not in the list and is also duplicated in an above example (not paying prefund).
I think the list above doesn't have a code for "paymaster not sufficiently staked", so this could be it.

eip/EIPS/eip-4337.md
```diff
@@ -391,78 +451,104 @@ The result `SHOULD` be set to true if and only if the request passed simulation
...
+An equivalent of `eth_getTransactionByHash`
 
-The paymaster scheme allows a contract to passively pay on users' behalf under arbitrary conditions. It even allows ERC-20 token paymasters to secure a guarantee that they would only need to pay if the user pays them: the paymaster contract can check that there is sufficient approved ERC-20 balance in the `validatePaymasterUserOp` method, and then extract it with `transferFrom` in the `postOp` call; if the op itself transfers out or de-approves too much of the ERC-20s, the inner `postOp` will fail and revert the execution and the outer `postOp` can extract payment (note that because of storage access restrictions the ERC-20 would need to be a wrapper defined within the paymaster itself).
+**Parameters**
+* **hash** a requestId value returned by `eth_sendUserOperation`
 
-### First-time wallet creation
+**Return value**:
+Transaction object (note that the returned transaction object does not contain the requestId itself)
+In case the UserOperation is not yet mined, return `null`
```
yoavw: "mined" -> "included in a block".  There's no mining :)
(also appears later in the doc)

eip/EIPS/eip-4337.md
drortirosh: Look at other rpc APIs. It's but a high level language, and they don't introduce objects just because. A UserOp is an object, like tx, but other structures are not added just for the sake of the api

eip/EIPS/eip-4337.md
drortirosh: Ok.
(Codes are 32500, since I'm not sure if they deliberately selected numbers just close to type(int16).min ..
I didn't want to get below -32768..

eip/EIPS/eip-4337.md
drortirosh: not new, only moved up.
The rationale section does need cleanup, I agree

eip/EIPS/eip-4337.md
drortirosh: yes. entire rationale needs work. but not in the RPC calls PR.
it deserves a separate one.
it only appears here as a change since I moved the "Rationale" section up, above the RPC calls.

yoavw (commented):
Looks good to me.  Do the new `AA??` revert messages correspond to RPC error numbers?

drortirosh:
> Looks good to me. Do the new `AA??` revert messages correspond to RPC error numbers?

No. they are just unique values in EntryPoint, so the rest of the string is not really needed.
informally, they are split into AA1x (init) , AA2x (account validation), AA3x (paymaster validation) and AA9x (system/should-never-happen errors)

yoavw (commented):
The change makes sense, but needs to be clearly communicated to existing projects. The type remains the same but the meaning of the number is very different, so existing code needs to be changed to extract the price from gasCost and gasUsed.

drortirosh:
checked again - the only other change to this event was the name of userOpHash.
So in order to make a new topichash (and slightly reduce gas cost...) we can remove the "indexed" from paymaster, which is completely useless.

yoavw:
> checked again - the only other change to this event was the name of userOpHash. So in order to make a new topichash (and slightly reduce gas cost...) we can remove the "indexed" from paymaster, which is completely useless.

Why is it useless?  If you're operating an owned (centralized) paymaster, wouldn't it be useful to get all the UserOps that use this paymaster in order to monitor for abuse or to calculate costs for off-chain charging?

shahafn (changes requested):
getUserOperationTransactionByHash is not documented

drortirosh:
> getUserOperationTransactionByHash

it is called `eth_getUserOperationReceipt`

shahafn:
> > getUserOperationTransactionByHash
> 
> it is called `eth_getUserOperationReceipt`

This is another function

Issue #135, opened 2022-11-17
Cannot add two values with different units in DepositPaymaster and TokenPaymaster


Jammyaa:
Hey eth-infinitism contributors,
I discovered in the source code of sample/DepositPaymaster.sol and sample/TokenPaymaster.sol, when calculating the actualTokenCost, you guys have added actualGasCost ( = actualGas * gasPrice, so its unit is wei) and COST_OF_POST (its unit is gas unit). 
```sol 
// DepositPaymaster.sol - Line 153
uint256 actualTokenCost = (actualGasCost + COST_OF_POST) * maxTokenCost / maxCost; // COST_OF_POST = 35000
```
and
```sol
// TokenPaymaster.sol - Line 109
uint256 charge = getTokenValueOfEth(actualGasCost + COST_OF_POST); // COST_OF_POST = 15000
```
This is wrong math logic when you guys try to add two numbers with different units.
Could anyone review this issue? Thank you guys very much.

Jammyaa:
I fix it in PR https://github.com/eth-infinitism/account-abstraction/pull/137

PR #137, opened 2022-11-18
fix error add two numbers with different units

Jammyaa:
Fix issue #135


yoavw (commented):
Thanks.  Looks good to me.
@drortirosh any comment?

PR #143, opened 2022-12-02
Cosmetic changes

plusminushalf:


eip/EIPS/eip-4337.md
```diff
@@ -55,10 +55,10 @@ To avoid Ethereum consensus changes, we do not attempt to create new transaction
 | `callData` | `bytes` | The data to pass to the `sender` during the main execution call |
 | `callGasLimit` | `uint256` | The amount of gas to allocate the main execution call |
 | `verificationGasLimit` | `uint256` | The amount of gas to allocate for the verification step |
-| `preVerificationGas` | `uint256` | The amount of gas to pay for to compensate the bundler for pre-verification execution and calldata |
+| `preVerificationGas` | `uint256` | The amount of gas to compensate the bundler for pre-verification execution and sending userOp parameters like calldata, paymasterAndData and signature to blockchain |
```
drortirosh: not exactly. the value is the amount to pay for entire userOp. some of this value is fixed, and some depends on actual userOp, like the content of the dynamic fields paymasterAndData, signature, callData.

plusminushalf: I wanted to convey the same that some portion of it is fixed and the variables include the parameters: paymasterAndData, signature, callData. Hence I used the word: "Like".

I feel we can be more explicit about this by wording it like this:
The amount of gas to compensate the bundler for pre-verification execution and sending userOp to the blockchain. UserOp has a fixed cost and a variable cost (depended on dynamic fields - paymasterAndData, signature and callData)

yoavw: Yes, I think this last paragraph describes it well.  @drortirosh ?

drortirosh: yes, this parameter deserves better description (probably its own section in the document),
maybe the right wording right now should be 
"the value to compensate the bundler for gas used by this UserOperation that is not calculated directly on-chain by EntryPoint's validation"

plusminushalf: I personally feel that `not calculated directly on-chain by EntryPoint's validation` is something that can not be understood at a glance. If we can be a little bit more specific about things that can't be calculated then it would make things easier.

Also I agree, we should have a dedicated area for this. A lot of people have asked for it's meaning in the discord server.

drortirosh: The major things it covers is `calldata` cost, fixed transaction cost (the 21000 gas stipend) and some overhead of the `handleOps` method.
The tricky part is that this overhead changes depends on the actual bundle size (e.g. the 21000 stipend should be split among all bundle users) - but at the time of submission, we can't tell what will be the bundle size..

yoavw: Right, but I think it's correct that most readers wouldn't know what can or cannot be calculated on-chain.  It may be obvious to us, but not to most solidity developers unless they spend a lot of time thinking about it.  So a concrete example is useful.

plusminushalf: I have added the minimal definition by @drortirosh and also added a section which explains this in more detail. Let me know if this would suffice as a better explanation.

PR #144, opened 2022-12-02
AA 73 unify reputation

drortirosh:
unified reputation mechanism for paymaster, deployer and aggregator.


contracts/core/EntryPoint.sol
```diff
@@ -285,23 +301,25 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
         uint256 preGas = gasleft();
-        MemoryUserOp memory mUserOp = opInfo.mUserOp;
-        _createSenderIfNeeded(opIndex, mUserOp, op.initCode);
+        _createSenderIfNeeded(opIndex, opInfo, op.initCode);
         if (aggregator == SIMULATE_FIND_AGGREGATOR) {
-            try IAggregatedAccount(mUserOp.sender).getAggregator() returns (address userOpAggregator) {
-                aggregator = actualAggregator = userOpAggregator;
+            //simulation marker after account creation,
+            // before calling account-specific methods (getAggregator, validateUserOp)
+            numberMarker();
```
yoavw, line 297: You added a 2nd marker to the code and the test below, but it's also mentioned in the EIP "In order to distinguish between them, there is a call to the NUMBER opcode (block.number), used as a delimiter between the validation functions." so I guess it must be also added there.

eip/EIPS/eip-4337.md
```diff
@@ -343,19 +343,21 @@ Exceptions to the forbidden opcodes:
...
    (that is, making calls is allowed, using `gasleft()` or `gas` opcode directly is forbidden)
 
-### Reputation scoring and throttling/banning for paymasters
+### Reputation scoring and throttling/banning for paymasters, deployers and aggregators
 #### Rationale
 
-Paymasters might influence many UserOperations in the mempool.
-To prevent abuse, we throttle down (or completely ban for a period of time) a paymaster that causes invalidation of large number of UserOperations in the mempool.
-To prevent paymasters from "sybil-attack", we require a paymaster to stake with the system, and thus make such DoS attack very expensive.
+UserOperation's memory access rules prevent them from interfere with each other.
```
yoavw: You mean storage access rules, not memory.

eip/EIPS/eip-4337.md
```diff
@@ -343,19 +343,21 @@ Exceptions to the forbidden opcodes:
...
+But "global" entities - paymasters, deployers and aggregators are accessed by multiple UserOperations, and thus might influence them.
```
yoavw: Maybe make it a bit more specific - that it could actually invalidate these previously-valid ops.
"...and thus might invalidate multiple previously-valid UserOperations."

eip/EIPS/eip-4337.md
```diff
@@ -343,19 +343,21 @@ Exceptions to the forbidden opcodes:
...
+To prevent abuse, we throttle down (or completely ban for a period of time) an entity that causes invalidation of large number of UserOperations in the mempool.
+To prevent such entities from "sybil-attack", we require them to stake with the system, and thus make such DoS attack very expensive.
 Note that this stake is never slashed, and can be withdrawn any time (after unstake delay)
+The only exemption from staking is if the entity doesn't use global storage
+(it is allowed to use storage specific to the account, such as mapping keyed by account address)
```
yoavw: "It is allowed to use storage specific to the account - any slots within the account, as well as any mapping keyed by the account address."

btw, this actually enables some useful unstaked paymasters.  E.g. a VerifyingPaymaster (which seems to be the most widely used use case so far) doesn't use storage at all, and even a TokenPaymaster could be written without accessing any global storage not keyed by the account address.  It's great if we can allow this safely.  Most paymasters won't need to stake anymore.

It does complicate the bundler a bit.  The bundler can no longer drop ops due to having an unstaked paymaster.  It must simulate the op, see that its paymaster validation stage doesn't access global storage, and only check for the stake if it does.

A paymaster could still invalidate many ops by depleting or withdrawing its deposit.  But luckily it's a cheap drop that doesn't require 2nd simulation.  These ops won't be included in the same bundle because the bundler already keeps track while building the bundle, and they'll be dropped the next time the bundler looks at them and sees that they require more gas than the paymaster can afford.

The paymaster could mess with the bundler's calculation by withdrawing its own deposit during validation (since `validatePaymasterUserOp` is no longer `view), but to do this it must access a storage slot in EntryPoint associated with itself (and not the account), so this paymaster will have to be staked and therefore would get itself banned.

Any other way an unstaked paymaster could invalidate other ops (other than storage access which we block for unstaked paymasters)?  If there aren't any, then we just improved paymasters significantly.

One thing we may need to clarify (since `validatePaymasterUserOp` is no longer `view`), is that `validatePaymasterUserOp` can read storage but not write to it.  We said the bundler will emulate the behavior of `staticcall` but I don't see where we wrote it in the EIP.  What we wrote is "The second (validatePaymasterUserOp), the paymaster may access storage slots of itself AND of the account, above." since we wrote it while it was still `view` and couldn't write.  Should we clarify that it's not allowed to write to storage?

yoavw: Another benefit of the above (if it's indeed safe) is that it'll remove some objections to high stake requirements.  If most paymasters don't need to stake, then maybe it's ok to require a higher stake from the rare ones who do.

yoavw: @drortirosh please see the questions in the above comment.  I think today's text changes still just use "storage access" and not make a distinction between reading and writing.  Should we clarify that entities are not allowed to write to storage outside their own, just read it, or are we ok with them writing as well?

contracts/core/EntryPoint.sol
drortirosh: Yes, I added a 2nd marker: this is the marker after creation, and the 2nd marker is after validateUserOp.
I don't really understand the comment.

yoavw: The comment was about having it in the code but not in the EIP.  Your commit today ([4714176](https://github.com/eth-infinitism/account-abstraction/pull/144/commits/47141764a634a09fd893bafc312386c1586d17d4)) fixed that.

contracts/core/EntryPoint.sol
```diff
@@ -230,16 +230,18 @@ contract EntryPoint is IEntryPoint, StakeManager {
         (address aggregator, uint256 deadline) = _validatePrepayment(0, userOp, outOpInfo, SIMULATE_FIND_AGGREGATOR);
         uint256 prefund = outOpInfo.prefund;
         uint256 preOpGas = preGas - gasleft() + userOp.preVerificationGas;
-        DepositInfo memory depositInfo = getDepositInfo(outOpInfo.mUserOp.paymaster);
-        PaymasterInfo memory paymasterInfo = PaymasterInfo(depositInfo.stake, depositInfo.unstakeDelaySec);
+        StakeInfo memory paymasterInfo = getStakeInfo(outOpInfo.mUserOp.paymaster);
+        StakeInfo memory senderInfo = getStakeInfo(outOpInfo.mUserOp.sender);
+        bytes calldata initCode = userOp.initCode;
+        address factory = initCode.length>=20 ? address(bytes20(initCode[0 : 20])) : address(0);
+        StakeInfo memory factoryInfo = getStakeInfo(factory);
```
forshtat, line 238: Both `factory` and `aggregator` can be `address(0)` and yet the code to handle them is very different. Also, only aggregator code is wrapped in another struct, making it nested. Is there a reason why this is done this way? I though the idea was to "unify" the code.

forshtat: Actually, the `paymaster` can also be `address(0)`, can it not?

drortirosh: the reason for the difference is that paymaster and factory are specified in the userOp, but aggregator is a method call on the wallet.
so when simulate returns, it can tell what paymaster the stake relates to, but can't easily tell what aggregator was used (with deployed wallet it can make another eth_call to determine the aggregator, but with newly deployed wallet, it can't do even that) - so we must return it.

PR #147, opened 2022-12-10
AA85 simplify verificationGas

drortirosh:


yoavw (commented):
This PR removes all validation gas limits (except their use in postOp). So now we're only checking the gas only at the end of validation. Doesn't it open an easy DoS vector on nodes?

Attacker will keep sending UserOps with very high `callGas`, normal `verificationGas`, running an infinite loop in `validateUserOp`.

These ops won't be propagated since they'll `FailOp` in the gas check at the end of `_validatePayment`, but they will waste more than `verificationGas` during 1st simulation.

I guess nodes can defend themselves by limiting the simulation gas to `verificationGas`+constant.  A bundler that doesn't, might find itself spending CPU on 10M-gas FailedOps.

Another attack vector is against staked entities.  We bundle the unstaked account and the staked paymaster together, reverting them in the gas check in line 409.  Two possible issues there:

1. Currently this FailedOp blames the paymaster, so an account can waste too much gas in `validateUserOp` - a little over verificationGas, but still below the slack that the bundler leaves for the EntryPoint overhead during simulation.  The paymaster will get blamed for the FailedOp in line 409, losing reputation and not being able to defend itself.

2. If we remove `mUserOp.paymaster` from this FailedOp, a staked paymaster will be able to DoS the 2nd simulation and invalidate O(n) UserOps without losing reputation.  This paymaster would have a flag, that when raised, causes it to waste a little over `verificationGas` (still within the slack given to the simulation).  The UserOps will pass 1st simulation, then the flag will be raised during the first included op, and all the other ops will revert in line 409 during their 2nd simulation.

So how do we distinguish between over-verificationGas spending by the account, by paymasters, and by factories?

drortirosh:
> 2\. If we remove `mUserOp.paymaster` from this FailedOp, a staked paymaster will be able to DoS the 2nd simulation

Ok. so you say we can't do the simplification.
At first I thought there isn't a need to distinguish between different OOG scenarios, but now you describe why we should.

PR #148, opened 2022-12-13
AA85 check verificationGas

drortirosh:
check verificationGas on all operations (create, validate, paymaster)
report descriptive error on out-of-gas in validateUserOp, validatePaymasterUserOp


test/entrypoint.test.ts
```diff
@@ -265,12 +277,22 @@ describe('EntryPoint', function () {
...
     })
 
+    it('should report failure on insufficient verificationGas (OOG) for creation', async () => {
+      const op1 = await fillAndSign({
+        initCode: getAccountInitCode(accountOwner1.address, simpleAccountFactory),
+        sender: '0x'.padEnd(42, '1'),
+        verificationGasLimit: 1e5
+      }, accountOwner1, entryPoint)
+      await expect(entryPoint.callStatic.simulateValidation(op1, { gasLimit: 1e6 }))
+        .to.revertedWith('AA11 initCode failed')
```
shahafn: How do we actually assert that this reverts on OOG? this is a catch-all kind of message that doesn't really check insufficient verificationGasLimit, i.e. the test will falsely pass when failing on any other reason it seems.

test/entrypoint.test.ts
```diff
@@ -441,6 +463,15 @@ describe('EntryPoint', function () {
...
         await calcGasUsage(rcpt, entryPoint, beneficiaryAddress)
       })
+
+      it('should report failure on insufficient verificationGas after creation', async () => {
+        const op1 = await fillAndSign({
+          sender: account.address,
+          verificationGasLimit: 10000
+        }, accountOwner, entryPoint)
+        await expect(entryPoint.callStatic.simulateValidation(op1))
+          .to.revertedWith('AA23 reverted (or OOG)')
```
shahafn, line 494: Same issue as above it seems

test/entrypoint.test.ts
drortirosh: right. but the same initCode succeeds if gas is enough. EVM code doesn't give you a good way to detect OOG compared to other revert reasons, and I don't want to calculate (fixing revert string to "say failed or OOG", like other items.

shahafn (approved):

PR #149, opened 2022-12-14
AA-74: sample account cleanup part 1 - using ERC1967Proxy instead of instance

forshtat:


contracts/samples/SimpleAccountFactory.sol
```diff
@@ -1,15 +1,23 @@
...
 /**
  * A sample factory contract for SimpleAccount
  * A UserOperations "initCode" holds the address of the factory, and a method call (to createAccount, in this sample factory).
  * The factory's createAccount returns the target account address even if it is already installed.
  * This way, the entryPoint.getSenderAddress() can be called either before or after the account is created.
  */
 contract SimpleAccountFactory {
+    address private immutable accountImplementation;
+
+    constructor(address _accountImplementation){
```
drortirosh: do we need the factory to get the impl object?
it is tied to a specific implementation (or at least, specific signature for initialize function)
this makes the factory class reusable (because a new instance is needed for every implementation anyway)
I think its simpler to have the implementation constructed internally
the constructor parameter you will have is entryPoint, since the implementation is created for specific EP

contracts/samples/SimpleAccountFactory.sol
```diff
@@ -23,16 +31,22 @@ contract SimpleAccountFactory {
         if (codeSize > 0) {
             return SimpleAccount(payable(addr));
         }
-        ret = new SimpleAccount{salt : bytes32(salt)}(entryPoint, owner);
+        ret = SimpleAccount(payable(new ERC1967Proxy{salt : bytes32(salt)}(
+                accountImplementation,
+                abi.encodeWithSelector(SimpleAccount.initialize.selector, entryPoint, owner)
```
drortirosh: I prever abi.encodeCall - type-safe, and slightly shorter (no ".selector")

contracts/samples/SimpleAccount.sol
```diff
@@ -31,16 +35,11 @@ contract SimpleAccount is BaseAccount {
 
     IEntryPoint private _entryPoint;
```
drortirosh: the idea was that entryPoint is immutable (and requires "setImplementation" ) to replace.

contracts/samples/SimpleAccount.sol
```diff
@@ -80,9 +79,14 @@ contract SimpleAccount is BaseAccount {
...
      */
-    function _updateEntryPoint(address newEntryPoint) internal override {
-        emit EntryPointChanged(address(_entryPoint), newEntryPoint);
-        _entryPoint = IEntryPoint(payable(newEntryPoint));
+    function initialize(IEntryPoint anEntryPoint, address anOwner) public virtual initializer {
+        _initialize(anEntryPoint, anOwner);
+    }
+
+    function _initialize(IEntryPoint anEntryPoint, address anOwner) internal virtual {
+        _entryPoint = anEntryPoint;
```
drortirosh: initializer function can only be called once - even when replacing implementation.
This means that your wallet can't change neither its entryPoint nor its owner.
Please add samples on how to use your wallet to change owner and entryPoint

drortirosh: I think that:
- entryPoint should be immutable member of the impl.
- the factory should have a method getImpl(entryPoint),  (and a hash of impl by entryPoint)
- this also means factory calls "new Impl" by itself, and doesn't receive it as constructor param.
- anyone can call addImpl() to create a new entry (whoever calls it pays for the gas to add it to the hash. its not "official" or "endorsed" entryPonit)
- setEntryPoint in SimpleAccount can be implemented as
    setImplementation(factory.getImpl(newEntryPoint))
   (this means that implementation has 2 immutable members (and constructor params): factory and entrypoint. the factory is interface, since you can't make circular reference to the class that creates you.

contracts/samples/SimpleAccountFactory.sol
forshtat: We can have:
1) both `implementation` and `entrypoint` be immutable members of the factory
2) one of the addresses be immutable members of the factory (i.e. same for all created proxies) and the other be configurable in constructor
3) all parameters be configurable in constructor

I have judged that the FACTORY creates instances of the same ACCOUNT, therefore the implementation used cannot be configurable. 

The other sane alternative is a FACTORY that creates ANY kind of proxy account.

Having a Factory that deploys proxies with an arbitrary implementation but immutable `EntryPoint` does not seem valid to me.

PR #152, opened 2022-12-21
AA 92 estimate execution

drortirosh:
1. A new API for an account (or paymaster) to report signature check validation: instead of revert, they should return `SIG_VALIDATION_FAILED` (1) to mark signature validation.  checking for `tx.origin` as zero is no longer in use.
2. ValidateUserOp returns the SIG_VALIDATION_FAILED value (like any other deadline) for a bundler to check.
3. `handleOps`  reverts on such signature failure.
4. a new `simulateHandleOp()` helper method, which executes a full UserOp (validation+execution).
   like `validateUserOp`, it reverts so it never commits changes on-chain, only used for off-chain simulation. This way, a userOp can be "estimated", even if depends on the deployed account (e.g. it is a method call on a newly deployed account, or depends on the account's state such as its balance)


contracts/core/EntryPoint.sol
```diff
@@ -24,8 +23,13 @@ contract EntryPoint is IEntryPoint, StakeManager {
 
     SenderCreator private immutable senderCreator = new SenderCreator();
 
+    /**
+     * for simulation purposes, account should accept signature
+     * failure if this is the given aggregator.
```
yoavw: Maybe add that EntryPoint guarantees that this UserOperation will revert without side effects if called on-chain.

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +145,27 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+     * This way, this method can be used to estimate entire execution without requiring the user
+     * to sign the request.
+     * Note that in order to collect the the success/failure of the target call, it must be executed
+     * with trace enabled to track the emitted events.
+     */
+    function estimateExecution(UserOperation calldata op) external {
+
+        UserOpInfo memory opInfo;
+
+        _validatePrepayment(0, op, opInfo, NO_VALIDATION_AGGREGATOR);
```
yoavw: We execute _validatePrepayment+_executeUserOp but not the rest of the overhead of handleOps.  Therefore it can be used for simulation but not for gas estimation (unless the fixed overhead is added).

yoavw (commented):
Looks good (added a couple of comments), but no tests.  One important test (aside from the ones for the new functionality) is that an attempt to call a normal handleOps transaction gets reverted when calling the address(1) aggregator. Solidity takes care of that, but it might change in the future so let's test for it.

contracts/core/EntryPoint.sol
drortirosh: It does almost all the same on-chain gas calculations as handleOps (there is a slight difference, since there is no loop over an array)
It does differ in preVerificationGas:
- it doesn't account for bundle size, can modify greatly preVerificationGas.
- it doesn't account for aggregator gas (again, it can't be calculated, and must be added to preVerificationGas)

contracts/core/EntryPoint.sol
```diff
@@ -24,8 +22,16 @@ contract EntryPoint is IEntryPoint, StakeManager {
 
     SenderCreator private immutable senderCreator = new SenderCreator();
 
-    // internal value used during simulation: need to query aggregator.
-    address private constant SIMULATE_FIND_AGGREGATOR = address(1);
+    uint256 private constant SIMULATION_INDEX = 99999;
+
+    /**
+     * for simulation purposes, validatePaymasterUserOp must return this value
```
yoavw: only validatePaymasterUserOp?  Not validateUserOp?

contracts/core/EntryPoint.sol
```diff
@@ -24,8 +22,16 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+     * in case of signature failure, instead of revert.
+     * failure if this is the given aggregator.
+     * (it is guaranteed that this aggregator cannot be used on-chain, only
```
yoavw: What aggregator?  :)
I assume you mean to change it so that it is guaranteed that if an account or a paymaster return this value, the UserOp will not be executed on-chain.

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +151,28 @@ contract EntryPoint is IEntryPoint, StakeManager {
         _compensate(beneficiary, collected);
     }
 
+    /**
+     * perform estimation of enture UserOperation.
+     * this method will always revert. it performs full validation of the UserOperation,
+     * but passing NO_VALIDATION_AGGREGATOR to the account.validateUserOperation, to signal it
```
yoavw: Seems to no longer be the case.

contracts/core/EntryPoint.sol
```diff
@@ -375,6 +401,31 @@ contract EntryPoint is IEntryPoint, StakeManager {
     }
     }
 
+    /**
+     * revert if either account deadline or paymaster deadline is expired
+     */
+    function _validateDeadline(uint opIndex, uint256 deadline, address paymaster, uint256 paymasterDeadline) internal view {
+        //we want to treat "zero" is "maxint", so we subtract one, ignoring overflow
```
yoavw: "zero" **as** "maxint".
And I assume underflow rather than overflow.

contracts/core/EntryPoint.sol
```diff
@@ -375,6 +401,31 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+    unchecked {
+        // solhint-disable-next-line not-rely-on-time
+        if ((deadline - 1) < block.timestamp) {
```
yoavw: Deliberate underflow as a feature - it's going to be fun explaining during the audit ;)

contracts/core/EntryPoint.sol
```diff
@@ -405,14 +456,10 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
         if (mUserOp.paymaster != address(0)) {
-            uint paymasterDeadline;
             (context, paymasterDeadline) = _validatePaymasterPrepayment(opIndex, userOp, outOpInfo, requiredPreFund, gasUsedByValidateAccountPrepayment);
-            if (paymasterDeadline != 0 && paymasterDeadline < deadline) {
-                deadline = paymasterDeadline;
-            }
         } else {
             context = "";
-
+            paymasterDeadline = 0;
```
yoavw: Any way paymasterDeadline would not already be zero at this point?

contracts/core/EntryPoint.sol
```diff
@@ -24,8 +22,16 @@ contract EntryPoint is IEntryPoint, StakeManager {
 
     SenderCreator private immutable senderCreator = new SenderCreator();
 
-    // internal value used during simulation: need to query aggregator.
-    address private constant SIMULATE_FIND_AGGREGATOR = address(1);
+    uint256 private constant SIMULATION_INDEX = 99999;
```
yoavw: opIndex is a uint256.  Any reason to use 99999 rather than maxint256-1?  Who knows, maybe some chain with **really** large blocks will be able to run 100k bundles.

Unlikely on a conventional chain but may be possible on a zkEVM built in a way that doesn't require submitting the transaction - just proving its execution.

On a chain that enables such a huge transaction it'll be possible to bypass the aggregator and call aggregated accounts by padding 99999 UserOps (all calling the same minimal account to keep execution as cheap as possible), then calling the aggregated account.

contracts/core/EntryPoint.sol
```diff
@@ -375,6 +401,31 @@ contract EntryPoint is IEntryPoint, StakeManager {
     }
     }
 
+    /**
+     * revert if either account deadline or paymaster deadline is expired
+     */
+    function _validateDeadline(uint opIndex, uint256 deadline, address paymaster, uint256 paymasterDeadline) internal view {
```
yoavw: opIndex is defined as uint256 everywhere but here.

contracts/interfaces/IAccount.sol
```diff
@@ -8,6 +8,8 @@ interface IAccount {
     /**
      * Validate user's signature and nonce
      * the entryPoint will make the call to the recipient only if this validation call returns successfully.
+     * signature failure should be reported by returning SIG_VALIDATION_FAILED (1), and not by revert.
```
yoavw: Maybe mention that non-signature failures that should fail simulation (e.g. bad nonce) should still revert rather than return a value.  Simulation is meant to test before user-signature, but the rest of the UserOp must be valid.

contracts/interfaces/IAccount.sol
```diff
@@ -20,6 +22,7 @@ interface IAccount {
      *      can be withdrawn anytime using "entryPoint.withdrawTo()"
      *      In case there is a paymaster in the request (or the current deposit is high enough), this value will be zero.
      * @return deadline the last block timestamp this operation is valid, or zero if it is valid indefinitely.
+     *      signature failure is returned as SIG_VALIDATION_FAILED value (1)
```
yoavw, line 26: Ditto - other types of failure should still revert.

contracts/core/EntryPoint.sol
```diff
@@ -110,7 +118,9 @@ contract EntryPoint is IEntryPoint, StakeManager {
             IAggregator aggregator = opa.aggregator;
             uint256 opslen = ops.length;
             for (uint256 i = 0; i < opslen; i++) {
-                _validatePrepayment(opIndex, ops[i], opInfos[opIndex], address(aggregator));
+                UserOpInfo memory opInfo = opInfos[opIndex];
+                (uint256 deadline, uint256 paymasterDeadline,) = _validatePrepayment(opIndex, ops[i], opInfo, address(aggregator));
+                _validateDeadline(i, deadline, opInfo.mUserOp.paymaster, paymasterDeadline);
                 opIndex++;
```
yoavw, line 122: Worth the gas to explicitly check that it never reaches SIMULATION_INDEX?  Or should we just change SIMULATION_INDEX to maxint256-1 and assume it can't be reached?

(Reaching SIMULATION_INDEX can call an aggregated account without calling its aggregator, bypassing its validation).

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +151,28 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+     * This way, this method can be used to estimate entire execution without requiring the user
+     * to sign the request.
+     * Note that in order to collect the the success/failure of the target call, it must be executed
+     * with trace enabled to track the emitted events.
+     */
+    function estimateExecution(UserOperation calldata op) external {
+
+        UserOpInfo memory opInfo;
+
+        _validatePrepayment(SIMULATION_INDEX, op, opInfo, address(0));
```
yoavw: Why do we need to use SIMULATION_INDEX in estimateExecution?  It is used to trigger the code that finds the aggregator instead of calling the aggregator in handleAggregatedOps.  It is not used for skipping the signature check (instead, we just don't call validateDeadline here).  estimateExecution doesn't care about aggregators, so what's the reason for not using zero here?

reports/gas-checker.txt
```diff
@@ -8,28 +8,28 @@
...
+║ simple                         │     1 │         78098 │                │                     ║
 ╟────────────────────────────────┼───────┼───────────────┼────────────────┼─────────────────────╢
-║ simple - diff from previous    │     2 │               │          43377 │               12332 ║
+║ simple - diff from previous    │     2 │               │          43600 │               12555 ║
 ╟────────────────────────────────┼───────┼───────────────┼────────────────┼─────────────────────╢
-║ simple                         │    10 │        468533 │                │                     ║
+║ simple                         │    10 │        470511 │                │                     ║
 ╟────────────────────────────────┼───────┼───────────────┼────────────────┼─────────────────────╢
-║ simple - diff from previous    │    11 │               │          43573 │               12528 ║
+║ simple - diff from previous    │    11 │               │          43724 │               12679 ║
```
yoavw: The gas increase seems fairly consistent.  Do we know what caused it?  The switch to _validateDeadline doesn't seem that wasteful.

yoavw (commented):
Long review...

contracts/core/EntryPoint.sol
yoavw: If I understand correctly, the only real use of SIMULATION_INDEX is what SIMULATE_FIND_AGGREGATOR used to do - find the aggregator during off-chain validation check.  It's not really used for wallet simulation (see my question below in estimateExecution).  So why is this better than using a special aggregator value, whether it's address(1) or an address that is known to revert if used as an aggregator, such as EntryPoint?  I think `aggregator==address(EntryPoint)` is a fairly safe marker and we'll have a test that ensures handleAggregatedOps reverts on it.

(@drortirosh please pay attention to this addition - seems like github hides it for some reason)

contracts/core/EntryPoint.sol
drortirosh: fixed..

contracts/core/EntryPoint.sol
drortirosh: we handle specific value (zero) this way. its not a general case... should we remove it?

contracts/core/EntryPoint.sol
drortirosh: We report the index (in failop), so I didn't want to use type(uint).max ...
Can we assume that 1 million UserOps in a bundle is beyond the possible capacity of ethereum evm ? (raised from 100,000)

contracts/core/EntryPoint.sol
drortirosh: SIMULATION_INDEX has several purposes:
-  instead of receiving an aggregator, we query the wallet for its own aggregator
- it add a NUMBER marker
- revert on proper FailOp for missing contracts for wallet/paymaster

reports/gas-checker.txt
drortirosh: I think we've crossed some glass ceiling in the stack, and the compiler (we use via-ir) now automatically uses memory (which is more expensive).

drortirosh: > I think we've crossed some glass ceiling in the stack
no. that's not the case. the code compiles without via-ir. 
its the cost of bubbling up return values from inner method.
see latest commit (pass `opinfo` instead of `info.mUserOp.paymaster` into the method, and it saves 30 gas for each entry.

contracts/core/EntryPoint.sol
yoavw: Still seems to say `treat "zero" is "maxint"` rather than "as"

contracts/core/EntryPoint.sol
yoavw: I think it's fine, just a bit unusual.  The comment explains it but I assume it'll come up.

contracts/core/EntryPoint.sol
yoavw: I mean that it may be a (small) waste of gas if it is already guaranteed to be zero.   Can it somehow become non-zero, requiring this line?

contracts/core/EntryPoint.sol
yoavw: @drortirosh both comments became hidden with no reply, so making sure it doesn't "fall off".

contracts/core/EntryPoint.sol
yoavw: It is marked as "outdated" but the current code still has "uint" rather than "uint256".

contracts/core/EntryPoint.sol
```diff
@@ -375,15 +394,41 @@ contract EntryPoint is IEntryPoint, StakeManager {
     }
     }
 
+    /**
+     * revert if either account deadline or paymaster deadline is expired
+     */
+    function _validateDeadline(uint opIndex, UserOpInfo memory opInfo, uint256 deadline, uint256 paymasterDeadline) internal view {
```
yoavw, line 400: uint256 opIndex

contracts/core/EntryPoint.sol
yoavw: See my other comment about this number.
1. On a normal chain a million is well beyond reach.  But what about a zkEVM where blocks are not submitted, and the prover just proves the validity of the state transition?  I'm no ZK expert, but some of these chains prove huge state changes every few hours, spanning across a many blocks that happened during that time.  They could theoretically support very large blocks.
2. Why is this number even needed?  See my suggestion regarding aggregator==address(EntryPoint).

contracts/core/EntryPoint.sol
yoavw: And any of them is relevant during estimateExecution?  I think they're all relevant to the bundler (during validation check) but not to a user during wallet-simulation.  If there's value in it, we can leave it here.  I'm just wondering if there is.

contracts/core/EntryPoint.sol
yoavw: uint256 opIndex

contracts/core/EntryPoint.sol
drortirosh: yes, they are.
- we can't call validateUserOp with wrong aggregator (it reverts), so we must first query for it.
- we do want proper exception strings instead if revert("")
- this is the core of estimateUserOperation, so you do need the marker to determine the boundary for gas usage (actually, it is possible also without it, by full tracking of method signatures and return values)
-

contracts/interfaces/IAccount.sol
drortirosh: I doin't like repeating myself too much. I do mention SIG_VALIDATION_FAILED value here, since it is a valid return value.

contracts/core/EntryPoint.sol
drortirosh: didn't find this aggregator==address(entrypoint).
the index appears in FailedOp, and dumped in logs.
Do you think `handleOps(ops[])` can be executed on a zkEVM with an array of over million entries?

contracts/core/EntryPoint.sol
yoavw: @drortirosh here's one for example, hidden a link to it doesn't seem to work.

contracts/core/EntryPoint.sol
yoavw: I don't know why you can't see my comments.  It appears like you hid them (unanswered) and I can only unhide them on my end.  I'll send you separately.

I don't know whether a zkEVM prover will handle an array with over a million entries or not, and whether it can't do it now but will be able to do it in a couple of years.  zkEVM is making rapid progress.  I'm pretty sure it won't be able to do 2**256-1 but I'm not so sure about 1M.  They already prove a few hours of state transitions in a single proof, so why not a single large block?

But regardless, I'm not sure why this number is needed.  I wrote about it in one of the hidden comments.

contracts/core/EntryPoint.sol
yoavw: Got it.  I thought all these things are only done in validation simulation (validateUserOp) but not when doing full execution simulation (estimateExecution).  Do we also use the marker when simulating the entire op?

contracts/core/EntryPoint.sol
drortirosh: yes, both

contracts/core/EntryPoint.sol
drortirosh: fixed

contracts/core/EntryPoint.sol
drortirosh: We are not talking about proving hours of data - we specifically talk about a single handleOps call, and a batch of UserOps in it - and this batched (single) transaction we have index for. it can't span between different transactions.

contracts/interfaces/IAccount.sol
drortirosh: done

contracts/core/EntryPoint.sol
drortirosh: fixed

contracts/core/EntryPoint.sol
drortirosh: stated elsewhere - there are few things that we do during simulation and not in normal execution, and now I use SIMULATION_INDEX to mark it.

contracts/core/EntryPoint.sol
drortirosh: added marker so that if we will make a traceCall on the simulateExecution, it will still not garble the results.
(but anyway, it is not based on INDEX anymore, but back on SIMULATE_FIND_AGGREGATOR....

drortirosh:
> One important test (aside from the ones for the new functionality) is that an attempt to call a normal handleOps transaction gets reverted when calling the address(1) aggregator

added the test.

contracts/core/EntryPoint.sol
forshtat:
```
if (deadline == SIG_VALIDATION_FAILED){
 revert();
} else if (deadline == 0) {
  // no-op
} else if (deadline < block.timestamp){
  revert();
}
```

Would this not work? Is it in any way worse then `(deadline - 1) < block.timestamp)`? 

This logic repeats twice - for Account and for Paymaster. Won't it be wise to extract the repeating part to a utility method?

contracts/core/BaseAccount.sol
```diff
@@ -56,6 +59,7 @@ abstract contract BaseAccount is IAccount {
      *          (also hashes the entrypoint and chain-id)
      * @param aggregator the current aggregator. can be ignored by accounts that don't use aggregators
      * @return deadline the last block timestamp this operation is valid, or zero if it is valid indefinitely.
+     *      return SIG_VALIDATION_FAILED in case signature fails
```
forshtat, line 62: Well, I don't know about you, but I believe variables have names for a reason. And functions have return value for a reason. In this case I don't see a benefit in reusing the `deadline` variable as a `validation success` flag.

Solidity can totally return two values: a boolean success flag and an integer value.

Please do explain why did you choose to merge the values into one as I don't see it and this seems to me to make this trivial code confusing.

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +149,24 @@ contract EntryPoint is IEntryPoint, StakeManager {
         _compensate(beneficiary, collected);
     }
 
+    function simulateExecution(UserOperation calldata op) external override {
+
+        UserOpInfo memory opInfo;
+
+        (uint256 deadline, uint256 paymasterDeadline,) = _validatePrepayment(0, op, opInfo, SIMULATE_FIND_AGGREGATOR);
```
forshtat, line 156: You do pass the `SIMULATE_FIND_AGGREGATOR` into the `_validatePrepayment`, so the method is, so to speak, "simulation-aware".
However, for the purpose of `deadline == SIG_VALIDATION_FAILED` check, it is not aware it is being run in a simulation and returns a failure. I think this is inconsistent.
Maybe if you did not have to reset the `deadline` and `paymasterDeadline` in the `simulateExecution` you would not have to separate the `_validatePrepayment` and `_validateDeadline`? This separation seems to me a bit arbitrary anyway.

contracts/interfaces/IEntryPoint.sol
```diff
@@ -160,5 +173,15 @@ interface IEntryPoint is IStakeManager {
...
+
+    /**
+     * simulate full execution of a UserOperation (including both validation and target execution)
+     * this method will always revert. it performs full validation of the UserOperation.
+     * Note that in order to collect the the success/failure of the target call, it must be executed
+     * with trace enabled to track the emitted events.
+     */
+    function simulateExecution(UserOperation calldata op) external;
+
+    error ExecutionComplete(uint256 preOpGas, uint256 paid, uint256 deadline, uint256 paymasterDeadline);
```
forshtat: "Simulation Complete"? What's wrong with the old "Simulation Result"? Reverts seem to be named XxxResult in other places.

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +149,24 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+        //ignore signature check failure
+        if (deadline == SIG_VALIDATION_FAILED) {
+            deadline = 0;
+        }
+        if (paymasterDeadline == SIG_VALIDATION_FAILED) {
+            paymasterDeadline = 0;
+        }
+        _validateDeadline(0, opInfo, deadline, paymasterDeadline);
+        numberMarker();
```
forshtat, line 165: It is hard to keep track of where these `numberMarker`s belong and why. Why is this one here? Maybe it should accept some string so it looks like `numberMarker("paymaster context");` or something.

contracts/core/EntryPoint.sol
```diff
@@ -141,6 +149,24 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+        uint256 paid = _executeUserOp(0, op, opInfo);
+        revert ExecutionComplete(opInfo.preOpGas, paid, deadline, paymasterDeadline);
```
forshtat: You don't know if the original `deadline` was `0` or `SIG_VALIDATION_FAILED` here. If that is not important there is no point even emitting those values, is there?

contracts/core/EntryPoint.sol
```diff
@@ -219,29 +245,27 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
-        uint256 prefund = outOpInfo.prefund;
-        uint256 preOpGas = preGas - gasleft() + userOp.preVerificationGas;
+        (uint256 deadline, uint256 paymasterDeadline, address aggregator) = _validatePrepayment(0, userOp, outOpInfo, SIMULATE_FIND_AGGREGATOR);
         StakeInfo memory paymasterInfo = getStakeInfo(outOpInfo.mUserOp.paymaster);
         StakeInfo memory senderInfo = getStakeInfo(outOpInfo.mUserOp.sender);
         bytes calldata initCode = userOp.initCode;
         address factory = initCode.length >= 20 ? address(bytes20(initCode[0 : 20])) : address(0);
         StakeInfo memory factoryInfo = getStakeInfo(factory);
 
+        ReturnInfo memory returnInfo = ReturnInfo(outOpInfo.preOpGas, outOpInfo.prefund, deadline, paymasterDeadline, getMemoryBytesFromOffset(outOpInfo.contextOffset));
```
forshtat, line 263: Is there any explanation why `_validatePrepayment` returns `deadline`, `paymasterDeadline` and `aggregator` by value but fills in `preOpGas`, `prefund`, `contextOffset` etc. on a memory-pointer `UserOpInfo` struct?
Note that if there is no rational explanation that means that it is just a very very bad code, in my opinion.

forshtat (changes requested):

contracts/core/BaseAccount.sol
drortirosh: 1. solidity is expensive when handling return variables. 
2. a low number is "invalid". I just define a special low value as special invalid

contracts/core/EntryPoint.sol
drortirosh: yes, the method is has limited knowledge of simulation, but I wanted to keep that to a minimum, since it is executed on-chain for each transaction.
small deadline values revert (on expiration) . so using "1" value will always revert.  that's the rationale of using a "deadline" value to mark a (different) failure (signature)
if you look at validateUserOp, it only calls validatePrepayment, NOT validateDeadline (since it returns the deadlines, for the bundler to check off-chain)

contracts/core/EntryPoint.sol
drortirosh: factory
<marker>
validateUserOp
<marker>
validatePaymasterUserOp
<marker>
exec 
maybe adding comments would help (as helpful as a string, since we're not going to do anything with it anyway)

contracts/core/EntryPoint.sol
drortirosh: I don't know what is the value, but I do want the bundler (external code) to know it.

contracts/core/EntryPoint.sol
drortirosh: 1. some of it is of course historic reasons.
2. parameters that are used directly are return values on the stack. parameters that are used later (sometimes even multiple times, in different contexts) or return multiple stack depths are in memory.

contracts/core/EntryPoint.sol
drortirosh: separate if's are possible (it was actually that way to begin with). but indeed, 0 is treated as "infinite" value. its so hard to pass such value in solidity, that I let the users put "0" instead..

the logic is repeated twice, on different fields. I did want to merge them to one (we care only about min(deadline, paymasterDeadline), after treating zero as infinite..), but since either can be SIG separately, I couldn't, and the logic to "merge" these values has to be off-chain.

contracts/interfaces/IEntryPoint.sol
drortirosh: right. will fix.

forshtat (dismissed):

contracts/core/EntryPoint.sol
forshtat: Re №2, I looked at the code and I don't see that difference. The `aggregator` stack variable is usually not used (makes sense), and `outOpInfo` is immediately unwrapped into the `ReturnInfo` (this line).

I can't block the entire PR on this, but I would be less sad if the `_validatePrepayment` could return only an instance of `OutOpInfo` that is allocated inside the method.
Or, if that is expensive in gas, if it could pack all fields into the struct.

We may do these kinds of changes in a separate PR for a later revision of the EntryPoint.

drortirosh: return a struct is inefficient in solidity - it requires in-memory copy. (it actually started that way, before I switched to an "out" reference).
please look at all calls to the methods.
UserOpInfo is the state that is kept per userOp between validation and execution. we need an array of those, since we perform all validations before all executions.
In simulateExecution (and in validateUserOp) we handle a single userOp, and thus we don't really need that, but we do want to use same inner functions.

drortirosh: UserOpInfo contains exactly the parameters we later move into innerHandleOp, and thus we don't want to expand it (it requires an actual copy in to the inner context)
I did experiment with a larger struct, wrapping it, that contains parameters returned by the validation functions that are not required by innerHandleOp, instead of using return values.
it turned out it took even more gas - since stack is more efficient than memory (and as a side-effect this change also obfuscated the code even more and made it even less comprehensible, since with a referenced struct you can't tell which parameter is input and which is output)

forshtat (approved):

PR #153, opened 2022-12-26
AA-93: Adding debug RPC APIs for the Bundler to use

shahafn:


eip/EIPS/eip-4337.md
```diff
@@ -453,26 +453,26 @@ For security reasons, it is important that the generated contract address will d
...
 it should make a static call to the `entryPoint.getSenderAddress()`
 
 ### Entry point upgrading
 
 Accounts are encouraged to be DELEGATECALL forwarding contracts for gas efficiency and to allow account upgradability. The account code is expected to hard-code the entry point into their code for gas efficiency. If a new entry point is introduced, whether to add new functionality, improve gas efficiency, or fix a critical security bug, users can self-call to replace their account's code address with a new code address containing code that points to a new entry point. During an upgrade process, it's expected that two mempools will run in parallel.
 
 ## RPC methods
-
-### * eth_sendUserOperation
+### debug Namespace
```
yoavw: There are two "debug Namespace" titles.  I think this one should have been something else, as it precedes the "eth_" ones.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+# Response
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": "0x1"
+}
+```
+
+### debug Namespace
+
```
yoavw, line 663: Maybe a few words here about why bundlers need to implement the debug namespace (required by the tests suite) and that these APIs shouldn't be available to users in production?

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+        callData, // bytes
+        callGasLimit, // uint256
+        verificationGasLimit, // uint256
+        preVerificationGas, // uint256
+        maxFeePerGas, // uint256
+        maxPriorityFeePerGas, // uint256
+        paymasterAndData, // bytes
+        signature // bytes
+      },
+      "prefund": "0x1234abcd",
```
yoavw: I think we're missing `entrypoint` here.  During periods of network upgrades, bundlers will participate in two mempools, one with the old entrypoint and one with the new one.  This is inevitable because the old one needs to be supported as long as some accounts haven't upgraded.  The `entrypoint` is also specified during eth_sendUserOperation, and is hashed into userOpHash.

Alternatively, add an optional `entrypoint` parameter to this call, since different entrypoints have different mempools.  This call should then only return the mempool of the requested entrypoint.  If not specified, use the primary one (the first one returned by `eth_supportedEntryPoints` as the EIP defines it to be the preferred one).

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+`prefund` - The calculated prefund required for the user operation execution.
+`aggregator` - (optional) the aggregator used by the senderaccount.
+
+```json=
+# Request
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "debug_bundler_dumpMempool",
+  "params": []
```
yoavw: Consider `entrypoint` param.  See explanation below.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+      "prefund": "0x1234abcd",
+      "aggregator": "0x7A0A0d159218E6a2f407B99173A2b12A6DDfC2a6"
+    }
+  ]
+}
+```
+
+#### * debug_bundler_sendBundleNow
+
+Forces the bundler to execute the entire current mempool as `handleOps()` transaction.
```
yoavw: It's not necessarily the entire bundle, as it may be impossible due to conflicts between ops, paymaster having insufficient funds for all ops, etc.  I think it triggers building and sending the maximum bundler possible.  That is, the most profitable bundle that can be built from the current mempool.  E.g. if there are two conflicting ops, the one with the highest priority fee will be simulated first and therefore included instead of the other one.

Some tests may actually rely on this behavior, and dump the mempool before and after, to see which ones were left behind.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+
+```json=
+# Request
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "debug_bundler_sendBundleNow",
+  "params": []
```
yoavw, line 742: If we add the `entrypoint` param to `dump` as I suggested above, we should add it to all the APIs that operate on a mempool, including this one.  If the bundler participates in multiple pools of different entrypoints, this should only act on the desired one (or the primary one if unspecified).

yoavw (commented):
Left some comments.  Mainly about entrypoints.

eip/EIPS/eip-4337.md
shahafn: Yes it should be eth namespace

eip/EIPS/eip-4337.md
shahafn: Ok, I thought it was implied by the name itself, but I can succinctly describe it.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+# Response
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": "ok"
+}
+```
+
+#### * debug_bundler_dumpMempool
+
```
drortirosh, line 688: dump the current **UserOperation**s mempool

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+Returns:
+
+`array` - Array of mempool entries objects, or an empty array.
+
+where each mempool entry has the following fields:
+
+`userOp` - UserOperation object. See `eth_sendUserOperation` for more details.
+`prefund` - The calculated prefund required for the user operation execution.
```
drortirosh: these are implementation-specific fields. I think the "dump" should only dump the raw UserOp entries

eip/EIPS/eip-4337.md
drortirosh: yes, I agree an entryPoint is needed as parameter (like in sendUserOperation)

eip/EIPS/eip-4337.md
drortirosh: yes, but we currently define methods require entryPoint, and there is no "default value" (the first one returned by supportedEntryPoints is the suggested default, but it is still required to be specified by rpc calls)

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": "ok"
+}
+```
+
+#### * debug_bundler_setBundleInterval
+
+Sets bundling mode and max mempool size.
```
drortirosh: this is a method I added to our implementation, but it is more complex than what is required just for the tests.
we require setManualMode(true) (so that sendBundleNow will work)
but the interval mode is not required.
also, the maxPoolSize is a parameter to force bundling when the mempool reaches this number. this is not something that is used by testing either.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+  "result": "ok"
+}
+```
+
+#### * debug_bundler_setReputation
+
+Sets reputation of given addresses. parameters:
+
+
+`array` - Aray of reputation objects, each with the following fields:
```
drortirosh: s/Aray/Array/
the rest of the items should be indented.

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+
+`address` - The address to set the reputation for.
+
+`opsSeen` - number of user operations seen linked to this address.
```
drortirosh: the number of user operations using this address that were added into the mempool

eip/EIPS/eip-4337.md
```diff
@@ -624,6 +624,219 @@ Returns an array of the entryPoint addresses supported by the client. The first
...
+
+`opsIncluded` - number of user operations included  in a mined bundle linked to this address.
```
drortirosh: number of userOps linked to this address that were included in past bundles (by any bundler)

eip/EIPS/eip-4337.md
drortirosh: yes. this is more of "trigger the logic to build the next bundle from the mempool" rather than "force"

shahafn: This behavior is only expected on test mode, so I assumed the bundler already dropped conflicting ops. So it shouldn't be in the mempool at that point anyway, no?

eip/EIPS/eip-4337.md
shahafn: Then don't return it in the bundler reference implementation... I went by the code.

eip/EIPS/eip-4337.md
shahafn: Added

drortirosh (approved):

PR #155, opened 2022-12-28
AA-94 update keccak rules.

drortirosh:
seems that allowing keccak(X||OWN) opens an attack surface (shared storage between accounts)
rule 3 alone is enough to access balances and allowances of 2 entitys (paymaster and account) - which is OK.
added also slot+offset

### DIFF
--- a/eip/EIPS/eip-4337.md
+++ b/eip/EIPS/eip-4337.md
@@ -313,9 +313,9 @@ In solidity, this includes all storage of the contract itself, and any storage o
 An address `A` is associated with:
 1. Slots of contract `A` address itself.
 2. Slot `A` on any other address.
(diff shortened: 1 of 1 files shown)


yoavw (commented):
LGTM.
After the release we should reach out to zksync to alert them of the potential vulnerability.  Not urgent because they [haven't implemented](https://v2-docs.zksync.io/dev/developer-guides/aa.html#building-custom-accounts) these rules yet:

`The verification rules are not fully enforced right now. Even if your custom account works right now, it might stop working in the future if it does not follow the rules below.`

We should also implement this rule in the reference bundler and add tests for accessing balances, allowances (from a staked paymaster), and a struct.  They should all work.  And then tests for accessing +129 and multi-level mapping, which should fail.

yoavw (approved):

PR #158, opened 2023-01-01
fix comment:uint112 is 10^15 eth rather than 2^15

arc0035:
uint112 can hold type(uint112).max / 10**18 = 5192296858534827 eth, which is 10^15, rather than 2^15


yoavw (approved):
Right

drortirosh:
thanks. already pushed in through another PR (can't merge this directly, because even this minor chnage affect gas usage..)

PR #159, opened 2023-01-01
AA-99 time range

drortirosh:


contracts/core/BaseAccount.sol
```diff
@@ -16,9 +16,20 @@ import "../interfaces/IEntryPoint.sol";
 abstract contract BaseAccount is IAccount {
     using UserOperationLib for UserOperation;
 
-    //return value in case of signature failure.
+    //return value in case of signature failure, with no time-range.
+    // equivalent to packSigTimeRange(true,0,0);
     uint256 constant internal SIG_VALIDATION_FAILED = 1;
 
+    /**
+     * helper to pack the return valid for validateUserOp
```
shahafn: "return value" not "return valid"

contracts/core/EntryPoint.sol
```diff
@@ -260,7 +260,9 @@ contract EntryPoint is IEntryPoint, StakeManager {
         address factory = initCode.length >= 20 ? address(bytes20(initCode[0 : 20])) : address(0);
         StakeInfo memory factoryInfo = getStakeInfo(factory);
 
-        ReturnInfo memory returnInfo = ReturnInfo(outOpInfo.preOpGas, outOpInfo.prefund, deadline, paymasterDeadline, getMemoryBytesFromOffset(outOpInfo.contextOffset));
+        bool sigFailed = uint8(deadline | paymasterDeadline) != 0;
+        //TODO: extract validBefore, not only validAfter
```
shahafn: Todo when?

contracts/core/EntryPoint.sol
```diff
@@ -397,29 +396,52 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
+        // solhint-disable-next-line not-rely-on-time
+        outOfTimeRange = block.timestamp > validUntil || block.timestamp < validFrom;
+    }
+
+    //extract sigFailed, validFrom, validUntil.
+    // also convert zero validUntil to type(uint64).max
+    function _parseSigTimeRange(uint sigTimeRange) internal pure returns (bool sigFailed, uint64 validFrom, uint64 validUntil) {
+        sigFailed = uint8(sigTimeRange) != 0;
+        // subtract one, to explicitly treat zero as max-value
+        validUntil = uint64(int64(int(sigTimeRange >> 8) - 1));
```
yoavw, line 434: Nice casting trick to hide the underflow without "unchecked" :)
Maybe one day Solidity will start catching casts of negative ints to uints, and our tests will catch it.

contracts/test/TestExpiryAccount.sol
```diff
@@ -22,17 +22,18 @@ contract TestExpiryAccount is SimpleAccount {
         addTemporaryOwner(anOwner, type(uint256).max);
     }
 
-    function addTemporaryOwner(address owner, uint256 deadline) public onlyOwner {
-        ownerDeadlines[owner] = deadline;
+    function addTemporaryOwner(address owner, uint256 sigTimeRange) public onlyOwner {
+        ownerDeadlines[owner] = sigTimeRange;
```
yoavw: It's no longer ownerDeadlines, but time range.

test/entrypoint.test.ts
```diff
@@ -1009,7 +1008,7 @@ describe('EntryPoint', function () {
             paymasterAndData: hexConcat([paymaster.address, expireTime])
           }, ethersSigner, entryPoint)
           const ret = await entryPoint.callStatic.simulateValidation(userOp).catch(simulationResultCatch)
-          expect(ret.returnInfo.paymasterDeadline).to.eql(now - 60)
+          expect(ret.returnInfo.validUntil).to.eql(now - 60 - 1)
```
yoavw, line 1010: We have tests for expiration `validUntil` for both Account and Paymaster, but no tests for `validAfter` for either, as far as I can tell.  Or a test that includes both for that matter (an op that becomes valid in a future block, and invalid in a later block).

yoavw (commented):
Left comments.  I think we lack tests for validAfter and for a combination of (validBefore,validAfter).

contracts/test/TestExpiryAccount.sol
drortirosh: yes, this test only checks for deadline, not for validAfter, so ownerDeadlines is ok, but sigTimeRange is wrong (it should be "deadline"
another test will check for validAfter, since it is not this use-case

test/entrypoint.test.ts
drortirosh: yes, I need to add an explicit test for validAfter.

contracts/core/BaseAccount.sol
drortirosh: fixed.

contracts/core/EntryPoint.sol
drortirosh: removed

yoavw:
@drortirosh I reviewed the new tests.  Looks good, but as far as I can tell we're only testing simulation, not attempting an actual `handleOps` with an out-of-time-range UserOp.  Shouldn't we have one to ensure that it behaves similarly to `simulateValidation`?

yoavw (approved):

PR #163, opened 2023-01-02
AA-101: validate gasLimit to leave more than callGasLimit  to call

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -56,6 +59,14 @@ contract EntryPoint is IEntryPoint, StakeManager {
         try this.innerHandleOp(userOp.callData, opInfo, context) returns (uint256 _actualGasCost) {
             collected = _actualGasCost;
         } catch {
+            bytes32 innerRevertCode;
+            assembly {
+                returndatacopy(0, 0, 32)
+                innerRevertCode := mload(0)
+            }
+            // handleOps was called with gas limit too low. abort entire bundle.
+            require(innerRevertCode != INNER_OUT_OF_GAS, "AA95 out of gas");
```
yoavw: A malicious paymaster could revert with 0xdeaddead from its postOp.  It would revert the entire bundle without a FailedOp, making it impossible to know what op caused the revert, and which paymaster to ban.

Since the op itself can't cause this OOG situation (both the account and the paymaster's postOp are called with gas limits), then it's either caused by the bundler, or by a malicious paymaster.  Therefore, why not revert it with a FailedOp, reporting the paymaster?  If the bundler is malicious then paymaster reputation held by it, is irrelevant.  And if the bundler is not malicious than it can only be the paymaster's fault.  No way for the account to cause this, right?

test/entrypoint.test.ts
```diff
@@ -433,6 +433,69 @@ describe('EntryPoint', function () {
...
+        await calcGasUsage(rcpt, entryPoint, beneficiaryAddress)
+
+        // check that the state of the counter contract is updated
+        // this ensures that the `callGasLimit` is high enough
+        // therefore this value can be used as a reference in the test below
+        console.log('  == offset after', await counter.offset())
+        expect(await counter.offset()).to.equal(offsetBefore.add(iterations))
+      })
+
+      it('account should not pay if too low gas limit was set', async function () {
```
yoavw: I assume this test passes now with the above fix?

yoavw (commented):
Looks good, except my comment about malicious paymasters.  But also a general question about the implementation:

I was also wondering if it would be more gas-efficient (and equally effective) if instead of testing for this OOG we would sum all the callGasLimits and verificationGasLimits (needed for postOp) into `totalGasLimit` during the validation loop, and then just require(gasleft() > totalGasLimit+CONSTANT) before starting the execution loop.

It doesn't seem as secure as actually checking for the OOG condition in every op, but if the CONSTANT is sufficient, wouldn't it be enough (and more gas efficient, since it's just one more "add" per op, and then a single "require" for the whole bundle)?

Or are there variable costs that would be hard to account for as a CONSTANT?

contracts/core/EntryPoint.sol
drortirosh: changing to FailedOp
(and it can only be with a staked paymaster, since it requires a "postOp" )

test/entrypoint.test.ts
drortirosh: yes, of course, all tests pass now (and dind't before the fix.

contracts/core/EntryPoint.sol
yoavw: Actually, I can't find where the EIP says postOp requires staking.  I only see the storage part.

yoavw (approved):

PR #165, opened 2023-01-07
Clarify that verifying the userOpHash is not a requirement

derekchiang:
The way it's worded right now, it sounds like the account needs to verify the `userOpHash`, when in fact it doesn't have to, since the entryPoint is trusted.  Just re-organizing the text to make it clearer.

### DIFF
--- a/eip/EIPS/eip-4337.md
+++ b/eip/EIPS/eip-4337.md
@@ -121,16 +121,15 @@ interface IAccount {
 }
 ```
 
-The account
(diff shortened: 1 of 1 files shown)


arc0035:
I have the same question here, does AA account really have to verify the opHash?@yoavw

yoavw:
Yes, userOpHash is calculated by mempool and doesn't need to be recalculated.  It does need to be verified against the signature.

derekchiang:
@yoavw all good to merge?

drortirosh (approved):

PR #166, opened 2023-01-08
Specify that the client should check for sigFailed

derekchiang:
`sigFailed` is returned as part of the `ValidationResult` from `simulateValidation`, but the spec does not ask the clients to consider it.

Related: https://github.com/eth-infinitism/bundler/pull/39

### DIFF
--- a/eip/EIPS/eip-4337.md
+++ b/eip/EIPS/eip-4337.md
@@ -305,7 +305,7 @@ Either `validateUserOp` or `validatePaymasterUserOp` may return a "validAfter" a
 The simulateValidation call returns this range.
 A node MAY drop a UserOperation if it expires too soon (e.g. wouldn't make it to the next block)
 
-If the `ValidationResult` includes `sigFail`, the client MUST drop the `UserOperation`.
+If the `ValidationResult` includes `sigFail`, the client SHOULD drop the `UserOperation`.
 
 The operations differ in their opcode banning policy.
 In order to distinguish between them, there is a call to the NUMBER opcode (`block.number`), used as a delimiter between the 3 functions.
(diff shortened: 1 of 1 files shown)


eip/EIPS/eip-4337.md
```diff
@@ -305,6 +305,8 @@ Either `validateUserOp` or `validatePaymasterUserOp` may return a "validAfter" a
 The simulateValidation call returns this range.
 A node MAY drop a UserOperation if it expires too soon (e.g. wouldn't make it to the next block)
 
+If the `ValidationResult` includes `sigFail`, the client MUST drop the `UserOperation`.
```
drortirosh: the node SHOULD drop this UserOperation

eip/EIPS/eip-4337.md
```diff
@@ -530,6 +532,7 @@ The result `SHOULD` be set to the **userOpHash** if and only if the request pass
     * The `data` field SHOULD contain a `minimumStake` and `minimumUnstakeDelay`
   * **code: -32506** - transaction rejected because wallet specified unsupported signature aggregator
     * The `data` field SHOULD contain an `aggregator` value
+  * **code: -32507** - transaction rejected because either the UserOperation signature or the paymaster signature is invalid
```
drortirosh: -32506 - transaction rejected because of wallet signature check failed (or paymaster signature, if the paymaster uses its data as signature)

derekchiang: Did you mean -32507?  or did you want this to be merged with -32506

drortirosh: no, my mistake in the number, only the text.
the failure is for signature failure. normal signature check is UseOp.signature (which is the user's signature)
there is no "paymaster signature" specific field - the paymaster may put some sig info into its paymasterAndData, as our verifyingPaymaster does.
That's why I suggested the text:

- -32507 - transaction rejected because of wallet signature check failed (or paymaster signature, if the paymaster uses its data as signature)

eip/EIPS/eip-4337.md
derekchiang: I updated it, but why do u think it should be SHOULD instead of MUST?  You would think that there's no valid scenarios where a client should keep a UserOp with invalid sig in the mempool.

eip/EIPS/eip-4337.md
derekchiang: Gotchu -- updated.

derekchiang:
@drortirosh good to merge?

drortirosh (approved):

derekchiang:
Hi @drortirosh should we merge this?

PR #171, opened 2023-01-10
AA-108 Leaked basefee test

shahafn:


contracts/test/MaliciousAccount.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.12;
+import "../interfaces/IAccount.sol";
+import "../interfaces/IEntryPoint.sol";
+
+contract MaliciousAccount is IAccount {
```
drortirosh, line 7: it is unclear from this test what is purpose, and how it is malicous.
specifically, I don't understand from the class (and from the test) how it "records" the original block basefee, in order to fail on the 2nd block.

shahafn: I'm extracting on-chain the block.basefee as you can see, and passing the expected basefee to compare with through maxPriorityFeePerGas parameter. It could be through any parameter, including nonce and signature.

test/entrypoint.test.ts
```diff
@@ -387,6 +387,46 @@ describe('EntryPoint', function () {
...
+      try {
+        await expect(entryPoint.simulateValidation(userOp, { gasLimit: 1e6 }))
+          .to.revertedWith('ValidationResult')
+        console.log('after first simulation')
+        await ethers.provider.send('evm_mine', [])
+        await expect(entryPoint.simulateValidation(userOp, { gasLimit: 1e6 }))
+          .to.revertedWith('FailedOp')
+        expect.fail(null, null, 'should fail on first simulation')
+      } catch (e: any) {
+        expect(e.message).to.include('FailedOp')
```
drortirosh: I don't think you this catch and "expect" - the first simulation "succeds" (revets with ValidationResult)
the 2nd simulation fails (wit FailedOn) - so this outer catch is not in use

shahafn: What?

contracts/test/MaliciousAccount.sol
```diff
@@ -0,0 +1,20 @@
...
+    constructor(IEntryPoint _ep) payable {
+        ep = _ep;
+    }
+    function validateUserOp(UserOperation calldata userOp, bytes32, address, uint256 missingAccountFunds)
+    external returns (uint256 sigTimeRange) {
+        ep.depositTo{value : missingAccountFunds}(address(this));
+        uint256 requiredGas = userOp.callGasLimit + userOp.verificationGasLimit + userOp.preVerificationGas;
+        uint256 gasPrice = missingAccountFunds / requiredGas;
+        uint256 basefee = gasPrice - userOp.maxPriorityFeePerGas;
+        require (basefee == userOp.maxPriorityFeePerGas, "Revert after first validation");
```
drortirosh: I don't understand these last 2  lines:
you're checking that gasPrice = 2*maxPriorityFeePerGas
looking at this contract, it is completely unclear why, and why it is "malicious" in doing so.

contracts/test/MaliciousAccount.sol
drortirosh: I cannot see: looking at the contract it doesn't make any sense, since you don't explain how it is used.

contracts/test/MaliciousAccount.sol
shahafn: Where do I check that gasPrice = 2*maxPriorityFeePerGas?

I explicitly check that  basefee == userOp.maxPriorityFeePerGas, which is just a controlled parameter that any wallet can set to signal something on-chain.

drortirosh (approved):

PR #172, opened 2023-01-10
changed to 8byte for match sigTimeRange spec of EIP

leekt:
the EIP 4337 official spec(https://eips.ethereum.org/EIPS/eip-4337) says the sigTimeRange value should use 8 byte to use validUntil, validAfter.
EntryPoint for parsing sigTimeRange is implemented properly but only the IPaymaster.sol says it's 4 byte


drortirosh (dismissed):

shahafn:
To fix the failed gas-checks, you can run `./scripts/gascalc` and check in the changes

leekt:
weird... all i've changed is few comments... I'll check what is the issue here

drortirosh (approved):

PR #177, opened 2023-01-19
C4 539 - Unsigned Integer Comparison

livingrockrises:
Use != 0 instead of > 0 for Unsigned Integer Comparison


drortirosh:
why is it better?
if you look at the gas checking, it is slighly more expensive (if it had any other benefit, I'd ignore ~20 gas difference, but I don't see why it is considered better)

livingrockrises:
I am not sure why it's slightly more expensive.
it logically makes sense for unsigned integer like either it's zero or not zero 

+ some reference here https://gist.github.com/IllIllI000/bf2c3120f24a69e489f12b3213c06c94

drortirosh:
I don't know either... Solidity gas calculation are something magical... I don't try to optimize +/- 20, since sometimes even renaming a method get that effect... (it changes the methodSig, reorder them. it also changes the deployed code a little).
I would be glad to add changes that improve 100s of gas, if the don't break much the clarity of code
(assembly at strategic points can achieve that, but try to avoid it.)

PR #178, opened 2023-01-19
AA-114 Prevent revert reason bombing of innerHandleOp()

shahafn:


contracts/core/EntryPoint.sol
```diff
@@ -219,9 +222,9 @@ contract EntryPoint is IEntryPoint, StakeManager {
 
         IPaymaster.PostOpMode mode = IPaymaster.PostOpMode.opSucceeded;
         if (callData.length > 0) {
-
-            (bool success,bytes memory result) = address(mUserOp.sender).call{gas : callGasLimit}(callData);
+            bool success = Exec.call(mUserOp.sender, 0, callData, callGasLimit);
```
drortirosh, line 225: please change the param of `innerHandleOp` from `calldata callData` to `memory callData`...
it saves 100 gas (currently, this feature cost 230 gas. with the above fix its only 130)

test/entrypoint.test.ts
```diff
@@ -427,6 +429,25 @@ describe('EntryPoint', function () {
...
+        sender: testRevertAccount.address,
+        callGasLimit: 1e5,
+        maxFeePerGas: 1,
+        verificationGasLimit: 1e5,
+        callData: badData.data!
+      }
+      const beneficiaryAddress = createAddress()
+      await expect(entryPoint.simulateValidation(badOp, { gasLimit: 1e6 }))
+        .to.revertedWith('ValidationResult')
+      const tx = await entryPoint.handleOps([badOp], beneficiaryAddress, { gasLimit: 1e6 })
```
drortirosh: 1. you test it with huge gaslimit: the whole idea is that it doesn't take too much gas, but you check with 10 times gasLimit. should test with (say) callGas + verification + 10k
2. the test should follow its title, and check the returned reason is up to 2k bytes.

drortirosh (approved):

contracts/core/EntryPoint.sol
livingrockrises: should one make use of memory instead of calldata everywhere for external functions?

Isn't this true? : To save a significant amount it is better to use calldata in the arguments of an external function

drortirosh: YMMV... Solidity optimization is funny. It has to copy into memory in order to be used. "memory" in parameter requires early copy for into memory, but in some cases Solidity gets confused and copy it again.. so it is better to try the options: make a change, and run "yarn gascalc" to see the effect on some basic scenarios. 
In this specific case, making the "context" parameter memory instead of calldata actually takes more gas... go figure.

filmakarov: @livingrockrises I think I get why it happens here. It depends on how the input parameter is being used then.
In this case, `callData` is used for a `.call` then, and call operates with memory , so it has to be copied to memory one more time, thus making it spend extra gas 

So, when you process a parameter in the method, it can make sense to have it as a calldata (i.e. you somehow iterate over array etc)
However, if you just use the whole input parameter it can make no difference or even increase gas consumption if it is passed as calldata.

PR #179, opened 2023-01-19
AA-115 Preventing detection of simulation vs execution by cold/warm storage access

shahafn:


contracts/core/EntryPoint.sol
```diff
@@ -345,27 +374,23 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
         if (aggregator == SIMULATE_FIND_AGGREGATOR) {
             numberMarker();
 
-            if (sender.code.length == 0) {
-                // it would revert anyway. but give a meaningful message
-                revert FailedOp(0, address(0), "AA20 account not deployed");
+            // solhint-disable-next-line no-empty-blocks
+            try this._simulateFindAggregator(sender, paymaster) {}
+            catch Error(string memory revertReason) {
+                revert FailedOp(opIndex, address(0), revertReason);
```
drortirosh: you lost the paymaster value in the 'paymaster not deployed'.
However, it is a bit redundant. I was actually thinking to remove it, since:
- we only report paymaster this way, not factory (or aggregator)
- the "AA1x/AA2x/AA3x" code reflect if it is a factory/account/paymaster, respectively
- we always pass "index" so it can be derived from the original UserOp.
So I think that's ok here.

contracts/core/EntryPoint.sol
```diff
@@ -345,27 +374,23 @@ contract EntryPoint is IEntryPoint, StakeManager {
...
-                revert FailedOp(0, address(0), "AA30 paymaster not deployed");
-            }
-            // during simulation, we don't use given aggregator,
-            // but query the account for its aggregator
-            try IAggregatedAccount(sender).getAggregator() returns (address userOpAggregator) {
-                aggregator = actualAggregator = userOpAggregator;
-            } catch {
-                aggregator = actualAggregator = address(0);
+            catch (bytes memory revertReason) {
+                if (revertReason.length != 32) {
```
drortirosh, line 387: how that can be? revert is never just ABI-encoded string (that is, you expect it to start with (zero word, length word and then data). If this is a catch-all for unknown error, then don't `abi.decode` it, just cast it to string. This way, it is possible to parse the FailedOp to find the original error. with this decode, it will juust revert with no data)

contracts/test/TestWarmColdAccount.sol
```diff
@@ -0,0 +1,40 @@
...
+
+import "../interfaces/IAggregatedAccount.sol";
+import "../interfaces/IEntryPoint.sol";
+
+
+contract TestWarmColdAccount is IAggregatedAccount {
+    IEntryPoint private ep;
+    uint public state = 1;
+    constructor(IEntryPoint _ep) payable {
+        ep = _ep;
```
drortirosh, line 13: what the ep is for (I also wanted to say it sholud be immutable, so it won't be in cold/warm storage)

test/entrypoint.test.ts
```diff
@@ -453,6 +455,54 @@ describe('EntryPoint', function () {
       const revertReason = Buffer.from(arrayify(userOperationRevertReasonEvent?.args?.revertReason))
       expect(revertReason.length).to.equal(REVERT_REASON_MAX_LEN)
     })
+    describe('warm/cold storage detection in simulation vs execution', () => {
+      const TOUCH_GET_AGGREGATOR = 1
+      const TOUCH_PAYMASTER = 2
+      it('should prevent detection through getAggregator()', async () => {
+        const testWarmColdAccount = await new TestWarmColdAccount__factory(ethersSigner).deploy(entryPoint.address,
+          { value: '0x' + 1e18.toString(16) })
```
drortirosh: I prefer `parseEther('1')` for value

contracts/test/TestWarmColdAccount.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    function validateUserOp(UserOperation calldata userOp, bytes32, address, uint256 missingAccountFunds)
+    external override returns (uint256 sigTimeRange) {
+        ep.depositTo{value : missingAccountFunds}(address(this));
+        if (userOp.nonce == 1) {
+            this.touchStorage{gas: 1000}();
```
drortirosh, line 22: missing comments what this "1000" means. These tests should be educational... also, I wonder how to refer to the failed EntryPoint. maybe add to the comment "test fails on EP from commit 0x123456". 
To verify test test one should check out that commit, and separately get the tests (which exist in the next commit). not that trivial, but at least possible.

contracts/test/TestWarmColdAccount.sol
```diff
@@ -0,0 +1,40 @@
...
+        } else if (userOp.nonce == 2) {
+            address paymaster = address(bytes20(userOp.paymasterAndData[: 20]));
+            this.touchPaymaster{gas: 1000}(paymaster);
```
drortirosh, line 26: add a comment such as: "this method call can only succeed of paymaster's code is already warm (e.g. in simulation, which reference it before validation, while in execution it is only referenced afterwards."

contracts/core/EntryPoint.sol
shahafn: How did I lose it? The previously emitted was address(0)

shahafn: Anyway fixed that.

contracts/core/EntryPoint.sol
shahafn: Fixed.

test/entrypoint.test.ts
shahafn: Fixed

contracts/test/TestWarmColdAccount.sol
shahafn: Added comments

contracts/test/TestWarmColdAccount.sol
shahafn: Added comments

drortirosh (approved):

PR #180, opened 2023-01-19
for loops can be gas optimised

livingrockrises:


drortirosh:
Why do you think this is better?
It is not a better syntax, and it also doesn't save any gas, so I see no reason to add it.
see the gas-checker report: it marginally moved the total gas by +/- 20 gas

PR #183, opened 2023-01-20
Update GnosisAccountFactory.sol

mehranhydary:
Updating the name of contract to match file name


drortirosh:
If you want to apply such a change, then please make it complete (use latest "safe" npm packages, update all contracts and tests)

drortirosh:
I'm closing this pr as it is partial: we currently use GnosisSafe code before the rename to "Safe". We can't rename just a single contract (factory), without using the entire suite of renamed contracts

PR #184, opened 2023-01-22
Fix/verifying paymaster

leekt:
Fixing sample verifying paymaster's signature replay attack
Changes
- EIP712 support for signature validation
- new storage 'SenderNonce' for nonce per sender 
- userOp.signature now has to be packed with validUntil, validAfter to meet the updated standard
  ```
     * paymasterAndData[:20] : address(this)
     * paymasterAndData[20:84] : abi.encode(validUntil,validAfter)
     * paymasterAndData[84:] : signature
    ```


contracts/samples/VerifyingPaymaster.sol
```diff
@@ -52,26 +59,48 @@ contract VerifyingPaymaster is BasePaymaster {
...
+
         //ECDSA library supports both 64 and 65-byte long signatures.
         // we only "require" it here so that the revert reason on invalid signature will be of "VerifyingPaymaster", and not "ECDSA"
         require(sigLength == 64 || sigLength == 65, "VerifyingPaymaster: invalid signature length in paymasterAndData");
 
         //don't revert on signature failure: return SIG_VALIDATION_FAILED
-        if (verifyingSigner != hash.toEthSignedMessageHash().recover(paymasterAndData[20 :])) {
-            return ("",1);
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(
+            keccak256("ValidatePaymasterUserOp(bytes32 hash,uint64 validUntil,uint64 validAfter,uint256 senderNonce)"),
```
drortirosh: its a nice touch, to add support for the new validUntil/ValidFrom fields in the verification signature.
but why do you think we need EIP712  here? it is an internal signature, created by a backend service specifically for this contract.
I actually think its better to pass validFrom/validTo to getHash() method, to simplify the off-chain creation of the hash.

contracts/samples/VerifyingPaymaster.sol
```diff
@@ -52,26 +59,48 @@ contract VerifyingPaymaster is BasePaymaster {
...
         //no need for other on-chain validation: entire UserOp should have been checked
         // by the external service prior to signing it.
-        return ("", 0);
+        return ("",packSigTimeRange(false,validUntil,validAfter));
     }
 
+    function parsePaymasterAndData(bytes calldata paymasterAndData) public pure returns(uint64 validUntil, uint64 validAfter, bytes calldata signature) {
+        assembly {
+            validUntil := calldataload(sub(paymasterAndData.offset, 4))
+            validAfter := calldataload(add(paymasterAndData.offset, 4))
```
drortirosh: Nice play with offsets... but I still prefer pure solidity, where it doesn't add much overhead..

test/verifying_paymaster.test.ts
```diff
@@ -61,7 +71,7 @@ describe('EntryPoint with VerifyingPaymaster', function () {
         const sig = await offchainSigner.signMessage(arrayify('0xdead'))
         wrongSigUserOp = await fillAndSign({
           sender: account.address,
-          paymasterAndData: hexConcat([paymaster.address, sig])
+          paymasterAndData: hexConcat([paymaster.address, '0x000000007B9F2E9D', '0x0000000000001234', sig])
```
drortirosh: please make it a "recognizable" value, or a constant. I read understand "0x1234" as "just a dummy value we can put in and later verify", but "7B9F2E9D" is unclear

test/verifying_paymaster.test.ts
```diff
@@ -77,15 +87,50 @@ describe('EntryPoint with VerifyingPaymaster', function () {
 
     it('succeed with valid signature', async () => {
       const userOp1 = await fillAndSign({
-        sender: account.address
+        sender: account.address,
+        paymasterAndData: hexConcat([paymaster.address, '0x000000007B9F2E9D', '0x0000000000000000'])
       }, accountOwner, entryPoint)
       const hash = await paymaster.getHash(userOp1)
```
drortirosh: here we could have  `getHash(userOp, since, until)` (no need to pass the nonce - this view method can read it from the contract)

contracts/samples/VerifyingPaymaster.sol
leekt: I used EIP 712 for domain separator support. seems getHash() does not consider chainId and paymaster address which is weak point for signature replay.
But if we add chainid, paymaster address along with validUntil and validAfter on getHash() function, it will be enough.
I agree with simplification for off-chain so I can change this part to not use EIP712

contracts/samples/VerifyingPaymaster.sol
leekt: agree i'll make this part simpler with abi.decode

test/verifying_paymaster.test.ts
leekt: Good point i'll use more dummy-like values and use constant variables to make things clear

leekt:
@yoavw i have issue with getHash(), this part generates "stack too deep", and if i turn on viaIR option on hardhat config, it passes but solidity-coverage fails on "stack too deep" seems solidity-coverage does not allow viaIR option per files
tried turning on `configureYulOptimizer: true` for `.solcover.js` but it results in failed test cases on other parts
will it be ok if i implement the getHash() part using assembly?

it will be something like this
```
    function pack(UserOperation calldata userOp) internal pure returns (bytes memory ret) {
        //lighter signature scheme. must match UserOp.ts#packUserOp
        bytes calldata pnd = userOp.paymasterAndData;
        // copy directly the userOp from calldata up to (but not including) the paymasterAndData.
        // this encoding depends on the ABI encoding of calldata, but is much lighter to copy
        // than referencing each field separately.
        assembly {
            let ofs := userOp
            let len := sub(sub(pnd.offset, ofs), 32)
            ret := mload(0x40)
            mstore(0x40, add(ret, add(len, 32)))
            mstore(ret, len)
            calldatacopy(add(ret, 32), ofs, len)
        }
    }

    /**
     * return the hash we're going to sign off-chain (and validate on-chain)
     * this method is called by the off-chain service, to sign the request.
     * it is called on-chain from the validatePaymasterUserOp, to validate the signature.
     * note that this signature covers all fields of the UserOperation, except the "paymasterAndData",
     * which will carry the signature itself.
     */
    function getHash(UserOperation calldata userOp, uint64 validUntil, uint64 validAfter)
    public view returns (bytes32) {
        //can't use userOp.hash(), since it contains also the paymasterAndData itself.

        return keccak256(abi.encode(
                pack(userOp),
                block.chainid,
                address(this),
                senderNonce[userOp.getSender()],
                validUntil,
                validAfter
            ));
    }
```

update) seems it still needs update in solidity-coverage is it ok if i update the package.json and yarn.lock?

yoavw:
> @yoavw i have issue with getHash(), this part generates "stack too deep", and if i turn on viaIR option on hardhat config, it passes but solidity-coverage fails on "stack too deep" seems solidity-coverage does not allow viaIR option per files tried turning on `configureYulOptimizer: true` for `.solcover.js` but it results in failed test cases on other parts will it be ok if i implement the getHash() part using assembly?
> 
> it will be something like this
> 
> ```
>     function pack(UserOperation calldata userOp) internal pure returns (bytes memory ret) {
>         //lighter signature scheme. must match UserOp.ts#packUserOp
>         bytes calldata pnd = userOp.paymasterAndData;
>         // copy directly the userOp from calldata up to (but not including) the paymasterAndData.
>         // this encoding depends on the ABI encoding of calldata, but is much lighter to copy
>         // than referencing each field separately.
>         assembly {
>             let ofs := userOp
>             let len := sub(sub(pnd.offset, ofs), 32)
>             ret := mload(0x40)
>             mstore(0x40, add(ret, add(len, 32)))
>             mstore(ret, len)
>             calldatacopy(add(ret, 32), ofs, len)
>         }
>     }
> 
>     /**
>      * return the hash we're going to sign off-chain (and validate on-chain)
>      * this method is called by the off-chain service, to sign the request.
>      * it is called on-chain from the validatePaymasterUserOp, to validate the signature.
>      * note that this signature covers all fields of the UserOperation, except the "paymasterAndData",
>      * which will carry the signature itself.
>      */
>     function getHash(UserOperation calldata userOp, uint64 validUntil, uint64 validAfter)
>     public view returns (bytes32) {
>         //can't use userOp.hash(), since it contains also the paymasterAndData itself.
> 
>         return keccak256(abi.encode(
>                 pack(userOp),
>                 block.chainid,
>                 address(this),
>                 senderNonce[userOp.getSender()],
>                 validUntil,
>                 validAfter
>             ));
>     }
> ```
> 
> update) seems it still needs update in solidity-coverage is it ok if i update the package.json and yarn.lock?

@drortirosh

drortirosh:
yes, I think you can use this pack, as we use it also in EntryPoint.
I don't want to enable yul on more than needed, since it is still very slow in development, so we enable it only for the core contracts (entrypoint and SimpleAccount, which is used in all gas checks).

drortirosh:
please fix the PR (merge master, and "yarn gas-calc").

test/verifying_paymaster.test.ts
```diff
@@ -37,19 +41,30 @@ describe('EntryPoint with VerifyingPaymaster', function () {
...
+      expect(res.signature).equal(MOCK_SIG)
+    })
+  })
+
   describe('#validatePaymasterUserOp', () => {
     it('should reject on no signature', async () => {
       const userOp = await fillAndSign({
         sender: account.address,
-        paymasterAndData: hexConcat([paymaster.address, '0x1234'])
+        paymasterAndData: hexConcat([paymaster.address, defaultAbiCoder.encode(['uint64', 'uint64'], [MOCK_VALID_UNTIL, MOCK_VALID_AFTER]), '0x1234'])
```
drortirosh: you keep the old uint64, which is confusing.
it is uint48 now.

leekt: good catch

drortirosh (approved):

drortirosh (approved):

PR #186, opened 2023-01-24
Create an execution method for Gnosis account that reverts

derekchiang:
Right now, when the EntryPoint executes a UserOp on a Gnosis account, it presumably uses the [`execTransactionFromModule`](https://github.com/safe-global/safe-contracts/blob/9d188d3ef514fb7391466a6b5f010db4cc0f3c8b/contracts/base/ModuleManager.sol#L63) function.  The problem, however, is that `execTransactionFromModule` doesn't revert when the underlying call reverts, which means that [the EntryPoint will not emit the `UserOperationRevertReason` event](https://github.com/eth-infinitism/account-abstraction/blob/0b5186b85b2352400eb906d8f1a19059951d285c/contracts/core/EntryPoint.sol#L225-L232).

This is an issue because [the client/frontend relies on the `UserOperationRevertReason` event to determine if a UserOperation failed](https://github.com/eth-infinitism/bundler/blob/a27f1b57a6d22b24c8d524c6569579f5e78522a7/packages/sdk/src/UserOperationEventListener.ts#L85).

So this PR adds a function `executeAndRevert` which will revert if the underlying call fails, so that the EntryPoint correctly emits the `UserOperationRevertReason` event.


contracts/gnosis/EIP4337Manager.sol
```diff
@@ -57,6 +57,35 @@ contract EIP4337Manager is GnosisSafe, IAccount {
...
+    ) external {
+        address _msgSender = address(bytes20(msg.data[msg.data.length - 20 :]));
+        require(_msgSender == entryPoint, "account: not from entrypoint");
+
+        (bool success, bytes memory returnData) = execTransactionFromModuleReturnData(to, value, data, operation);
+
+        // Revert with the actual reason string
+        // Adopted from: https://github.com/Uniswap/v3-periphery/blob/464a8a49611272f7349c970e0fadb7ec1d3c1086/contracts/base/Multicall.sol#L16-L23
+        if (!success) {
+            if (returnData.length < 68) revert();
```
drortirosh, line 86: why do you re-encode the revert reason?
why not simply revert with original revert data: `revert(add(ret, 32), mload(ret))`
(also note that the suggested code will parse Error(string), but will fail for custom errors)

drortirosh:
nice.
Can you please add a test to use it (the tests succeed for this PR since they use the old `execFromModule` mode.

contracts/gnosis/EIP4337Manager.sol
drortirosh: ^^^

derekchiang:
> nice. Can you please add a test to use it (the tests succeed for this PR since they use the old `execFromModule` mode.

@drortirosh Done!

drortirosh (approved):

PR #190, opened 2023-02-01
AA-118 simulateHandleOp should support extra call

drortirosh:
notes:
- added target, targetCallData to simulate
- added targetSuccess, targetResult to the resulting result (error object)
- the test for simulateHandleOp now can be complete (previously, it was very difficult/impossible to test for side-effects...
- added a "numberMarker" just before the call, so that tracing it can easily distinguish execution from post-execution call


yoavw (commented):
Looks good.  Would be nice to have a less synthetic test, e.g. a counterfactual EIP-1271 signature check.  But definitely not a must, it just shows a real example of how it's used for counterfactual checks.

yoavw (approved):

PR #191, opened 2023-02-02
Make Gnosis account compatible with ERC-1271

derekchiang:
Gnosis does have a [`CompatibilityHandler`](https://github.com/safe-global/safe-contracts/blob/main/contracts/handler/CompatibilityFallbackHandler.sol) that implements ERC-1271 but it doesn't compile with Solidity 0.8.15 which we are using.  In any case, their ERC-1271 implementation is optimized for multisigs whereas our Gnosis account is explicitly single-sig and therefore can be made more efficient (as in this PR).


contracts/gnosis/EIP4337Fallback.sol
```diff
@@ -30,4 +34,20 @@ contract EIP4337Fallback is DefaultCallbackHandler, IAccount {
...
+        bytes memory _signature
+    ) external override view returns (bytes4) {
+        bytes32 hash = _hash.toEthSignedMessageHash();
+        address recovered = hash.recover(_signature);
+
+        GnosisSafe safe = GnosisSafe(payable(address(msg.sender)));
+
+        // Validate signatures
+        if (safe.isOwner(recovered)) {
+            return 0x1626ba7e;
```
drortirosh: can you please extract this constant (also in the test code) ?
I prefer not to see "anonymous" magic values in code...
(0xffff is good enough, since its value is as good as a constant name...)

drortirosh: Also, it would be nice to add such implementation to BaseAccount, for the benefit of other wallet types (though it should be a separate PR)

derekchiang:
@drortirosh comment addressed and conflicts resolved 👍

drortirosh (approved):

PR #195, opened 2023-02-02
AA-119 H-01 Invalid aggregate signature

drortirosh:


contracts/bls/BLSSignatureAggregator.sol
```diff
@@ -124,7 +122,7 @@ contract BLSSignatureAggregator is IAggregator {
     external view returns (bytes memory sigForUserOp) {
         uint256[2] memory signature = abi.decode(userOp.signature, (uint256[2]));
         uint256[4] memory pubkey = getUserOpPublicKey(userOp);
-        uint256[2] memory message = userOpToMessage(userOp);
+        uint256[2] memory message = _userOpToMessage(userOp, _getPublicKeyHash(pubkey));
```
forshtat, line 125: Did you replace `userOpToMessage` here with the code that does the same to avoid repeated call to the  `getUserOpPublicKey`? I think this breaks the idea of having the same code be reusable.
If this is not too late to change, maybe it could make sense to return `pubkey`, `hashPublicKey` and `message` from `userOpToMessage()` helper?

contracts/test/BrokenBlsAccount.sol
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+
+/**
+ * Based n SimpleAccountFactory
+ * can't be a subclass, since both constructor and createAccount depend on the
+ * actual wallet contract constructor and initializer
+ */
+contract BrokenBLSAccountFactory {
```
forshtat, line 52: Maybe it is better to have a Base Factory so that we can change the implementation and `initialize` encoding without having to copy the entire contract? Now this is just for the tests, but in the future we may want different proxy account types?

I can make a PR if you think this is okay.

test/y.bls.test.ts
```diff
@@ -46,10 +54,8 @@ describe('bls account', function () {
     accountDeployer = await new BLSAccountFactory__factory(etherSigner).deploy(entrypoint.address, blsAgg.address)
 
     // TODO: these two are not created via the 'accountDeployer' for some reason - I am not touching it for now
```
forshtat: This comment is no longer correct.

test/y.bls.test.ts
```diff
@@ -94,6 +100,39 @@ describe('bls account', function () {
...
+    const beneficiary = createAddress()
+    const handleRet = await entrypoint.callStatic.handleAggregatedOps([
+      {
+        userOps: [userOp],
+        aggregator: blsAgg.address,
+        signature
+      }
+    ], beneficiary).then(() => 'ok', e => e.errorName) as string
+
+    expect(`${singleOpSigCheck},${handleRet}`)
```
forshtat, line 131: I am having a problem understanding what is being asserted here. Can we have a more explicit `expect`?

drortirosh: I want to assert that both are "ok", and still in the log see the failures of each.
this is because the first commit adds the tests and fails (the simulation succeeds but the handleOps fails), and the 2nd commit fixes the issues.

contracts/bls/BLSSignatureAggregator.sol
drortirosh: Yes, its possible, but I dislike returning unused data. returning 3 items means that each call-site receives return values it never uses

contracts/test/BrokenBlsAccount.sol
drortirosh: I can hardly think you can extract a base factory. each method depends on actual wallet implementation and also constructor (and factory method) signatures differ - whether it is a single or multiple keys, or signature types, or recovery mechanism or whatever.

shahafn (approved):

PR #197, opened 2023-02-02
AA-125 L-06 Mismatched StakeLocked event parameter

drortirosh:


contracts/interfaces/IStakeManager.sol
```diff
@@ -19,11 +19,11 @@ interface IStakeManager {
         uint256 amount
     );
 
-    /// Emitted once a stake is scheduled for withdrawal
+    /// Emitted when stake or unstake delay are modified
     event StakeLocked(
         address indexed account,
         uint256 totalStaked,
-        uint256 withdrawTime
+        uint256 unstakeDelaySec
```
forshtat, line 26: Isn't it better to know the `withdrawTime` from the event? Maybe emit both values, actually?

shahafn (approved):

contracts/interfaces/IStakeManager.sol
shahafn: No, since when you lock, you don't really have withdrawTime, since it's only set upon unlocking the stake.

PR #198, opened 2023-02-02
AA-127 L-08: Missing error messages in require statements

drortirosh:
add missing require message.
Note that account's addDeposit() now calls the paymaster's depositTo() explicitly, instead of relying on the EntryPoint's receive() fallback method.


contracts/core/BasePaymaster.sol
```diff
@@ -104,7 +104,7 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
 
     /// validate the call is made from a valid entrypoint
     function _requireFromEntryPoint() internal virtual {
-        require(msg.sender == address(entryPoint));
+        require(msg.sender == address(entryPoint), "Sender not EntryPoint");
```
forshtat, line 109: Do we want to have `AAxx` prefix on revert messages as well?

drortirosh: no AAxxx prefixes are only added by entrypoint, not other contracts
(in fact, the EP  will add "AA23" prefix to mark it was thrown by the wallet)

shahafn (approved):

PR #199, opened 2023-02-02
AA-128 L-09: document getBlsPublicKey

drortirosh:
remove confusing reference to non-existent (and non-standard) method getAggregationInfo()
give "getBlsPublicKey" as an example.


eip/EIPS/eip-4337.md
```diff
@@ -145,9 +145,10 @@ interface IAggregatedAccount is IAccount {
 
 * **getAggregator()** returns the aggregator this account supports.
 * **validateUserOp()** (inherited from IAccount interface) MUST verify the `aggregator` parameter is valid and the same as `getAggregator`
-* The account should also support aggregator-specific getter (e.g. `getAggregationInfo()`).
+* The account should also support aggregator-specific getter.
```
forshtat, line 148: Maybe this is better left out of the ERC, like `the aggregator and the account interaction is not restricted by this ERC. The account may need to support aggregator-specific view functions.` or something along this line?

shahafn (approved):

PR #201, opened 2023-02-02
AA-129 L-10 calling _disableInitializers() from account constructor

shahafn:
Currnetly failing bls account test, since it uses initalizer(), pending #195 merge


contracts/samples/SimpleAccount.sol
```diff
@@ -41,6 +41,7 @@ contract SimpleAccount is BaseAccount, UUPSUpgradeable, Initializable {
 
     constructor(IEntryPoint anEntryPoint) {
         _entryPoint = anEntryPoint;
+        _disableInitializers();
```
forshtat, line 44: The same should be done in the BLSAccount, I assume.

forshtat (changes requested):

forshtat (dismissed):

drortirosh (approved):

PR #203, opened 2023-02-02
AA-135 N-04 Changing validFrom, validUntil to uint64

shahafn:


contracts/core/BasePaymaster.sol
```diff
@@ -114,7 +114,7 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
      * @param validUntil last timestamp this UserOperation is valid (or zero for infinite)
      * @param validAfter first timestamp this UserOperation is valid
      */
-    function packSigTimeRange(bool sigFailed, uint256 validUntil, uint256 validAfter) internal pure returns (uint256) {
-        return uint256(sigFailed ? 1 : 0) | uint256(validUntil << 8) | uint256(validAfter << (64 + 8));
+    function packSigTimeRange(bool sigFailed, uint64 validUntil, uint64 validAfter) internal pure returns (uint256) {
```
forshtat, line 119: Just asking, isn't it a duplicate function? Do we not want to have it somewhere in utils?

drortirosh (approved):

drortirosh (dismissed):

contracts/core/BasePaymaster.sol
shahafn: It is, we should do it in a different PR probably.

drortirosh (approved):

PR #204, opened 2023-02-02
AA-136 N-05 Calling setPublicKey when initializing BLSAccount

shahafn:
instead of seting public key directly.

This PR depends on #195 to remove `initalize()` call from bls tests.


contracts/bls/BLSAccount.sol
```diff
@@ -23,7 +23,7 @@ contract BLSAccount is SimpleAccount, IBLSAccount {
 
     function initialize(uint256[4] memory aPublicKey) public virtual initializer {
         super._initialize(address(0));
-        publicKey = aPublicKey;
+        setBlsPublicKey(aPublicKey);
```
forshtat: I don't understand something. Line 25 seems to set the owner to `address(0)`? And `setBlsPublicKey` is `onlyOwner`? How does it work?

shahafn: Becausae onlyOwner was modified to allow `address(this)` as you can see. A name change for  onlyOwner should probably also be done. But not in this PR.

drortirosh (dismissed):

drortirosh (approved):

PR #214, opened 2023-02-07
Expose sender creator

drortirosh:


drortirosh:
This solution is incomplete. instead, #195  now has a fix that checks initCode in _validateSignature

PR #216, opened 2023-02-08
AA-147 AA-134 N-03: remove aggregator from IAccount

drortirosh:


contracts/core/BaseAccount.sol
```diff
@@ -17,19 +18,9 @@ abstract contract BaseAccount is IAccount {
     using UserOperationLib for UserOperation;
 
     //return value in case of signature failure, with no time-range.
-    // equivalent to packSigTimeRange(true,0,0);
+    // equivalent to _packSigTimeRange(true,0,0);
```
shahafn: Maybe it's time to rename it to something like packedValidationData struct and add doc: 
/**
 * Packed representation of the data used for validation.
 * It is packed with the following fields:
 * - signature check status or aggregator address (uint160)
 * - validFrom (uint48)
 * - validUntil (uint48)
 */

nikeshnazareth:
Looks good! I like the simplification. A few thoughts:

- I was originally going to say that a bundler can set the `aggregator` to `1` in order to execute user operations where the signature fails (`validateAccountAndPaymasterValidationData` just ensures that the value chosen by the bundler matches the one chosen by the user, not that it's not `1`). Although now I see it will fail when the EntryPoint tries to call `validateSignatures` on the `1` address. Nevertheless, I think I'd prefer an explicit `aggregator != 1` check so that it fails for the right reason (and is easier to reason about).
- Does `_validateAccountPrepayment` still need a `simulateOnly` branch? I can understand that if you have a simulation branch, you might as well check that sender and paymaster are deployed, but I'm not sure I understand the reason behind creating that branch (and passing around a `simulateOnly` parameter) just for those checks. Is it just for the gas savings?
- the EIP still talks about the 1-byte `sigFailure` value
- the `IEntryPoint` description of `ReturnInfo` has two `sigFailed` `@param` statements
- I'm not sure why the Stake Manager `withdrawTime` should be changed in this PR, but I guess making it consistent with other times is fine. The explanation of the `DepositInfo` struct (in `IStakeManager`) is now incorrect though
- I think the various descriptions of how `validationData` is packed is confusing
   - from the EIP I would guess it's `authorizer | validUntil | validAfter` (ie . the `authorizer` take the most significant bytes)
   - the actual packing is `validAfter | validUntil | authorizer`
   - the `ValidationData` struct lists them in the order `(aggregator, validAfter, validUntil)` so a storage record that contained a `ValidationData` object would be different than one containing the equivalent `validationData uint256` 
   - It's not an issue but I think it would be clearer to use the same order everywhere

shahafn (dismissed):

shahafn (approved):

PR #220, opened 2023-02-08
AA-131 L-12: Unsafe ABI encoding

drortirosh:


shahafn (approved):

drortirosh:
Original issue was opened for 2 cases of using encodeWithSignature. 
One was replaced with direct call (with try/catch).
The other was encoding of an error, which can't use abi.encodeCall (abi.encodeError is an open issue for Solidity)

PR #221, opened 2023-02-08
AA-138 N-07 rename vars and internal methods

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -82,7 +82,7 @@ contract EntryPoint is IEntryPoint, StakeManager {
     /**
      * Execute a batch of UserOperation.
      * no signature aggregator is used.
-     * if any account requires an aggregator (that is, it returned an "actualAggregator" when
+     * if any account requires an aggregator (that is, it returned an actual when
```
shahafn: actual -> aggregator

shahafn (dismissed):

shahafn (approved):

PR #225, opened 2023-02-16
Remove side effect codes

jayden-sudo:
The following end result is exactly the same:
bytes20(initCode[0 : 20]) and bytes20(initCode).
But using bytes20(initCode) will save 42% of the gas fee (370 vs. 211)


drortirosh:
please run "yarn gas-calc", to trigger gas calculations check. from what I see, it doesn't save at all, or neglible (<10gas) 
note that we compile with "--via-ir", which has better optimization. could it be that you tested it with normal compilation?

jayden-sudo:
I see, thank you

PR #228, opened 2023-02-24
add token callback handler for account

ququzone:


contracts/samples/callback/TokenCallbackHandler.sol
```diff
@@ -0,0 +1,61 @@
...
+    ) external pure override {
+    }
+
+    function onERC721Received(
+        address,
+        address,
+        uint256,
+        bytes calldata
+    ) external pure override returns (bytes4) {
+        return 0x150b7a02;
```
drortirosh: is it possible at least to use the selector, instead of hard-code value
e.g.
```js
return IERC721Receiver.onERC721Received.selector;
```

ququzone: yeah, fixed.

drortirosh (approved):

PR #233, opened 2023-03-02
modifying VerifyingPaymaster to not take into account paymasterAndData offset and signature offset

kristofgazso:


drortirosh:
I'm a bit reluctant to push such a change just a moment after the contract was audited. the current implementation is not broken - it just requires a paymasterAndData in the correct length.

fangting-alchemy:
While digging into the update with my team, we realized that it is potentially exploitable. Bytes manipulation could cause dynamic type like bytes (like `initCode` and `callData`) be excluded from the hash.

Both the current verifyingPaymaster and this update, using the `paymasterAndData`'s offset as the cut off point for what data to include in the hash, if the sender move the `paymasterAndData`'s location above the `initCode` and/or `callData`, both of those two dynamic type won't be included in the hash.  

The exploit will only happen if paymaster off-chain signing service also sign over the hash of the manipulated bytes. Which is totally possible:
* user send UO bytes to paymaster off-chain to sign
* paymaster uses UO bytes and calls `verifyingPaymaster`'s `getHash` and sign over the hash.
* UO submitted and executed with un-authorized (not included in hash) `callData` and/or `initCode`.

Is this a concern? 

Edit:
Add attack [PoC](https://www.notion.so/alchemotion/AA-Pack-Attack-PoC-a9f19d35871d431f88b1eb22d6a3a136).

drortirosh:
While this encoding vulnerability is true, the only one who can use it is the signer (the verifying service that signs off-chain this UserOp)
It is unlikely that this service will exploit itself - it should just the same blindly sign any request, without any test.

Any other entity (including the wallet owner itself) can't do anything: Any re-ordering of fields (e.g. putting initCode outside of the signed area) will modify the **offset** of the initCode in the calldata, and thus render the VerifyingPaymaster signature invalid.

So I don't think we should add this fix.
Moreover, the fix itself is very hard to comprehend, and deeply depend on the structure of UserOperation.
The current implementation depends on the fact that `paymasterAndData` is the last (before signature) field, but no more than that, not even on the number of other fields in the UserOp.

See also the note I put on #237 

If you think of an attack scenario (by another entity) on this paymaster, please let me know, and reopen this issue.

PR #240, opened 2023-03-13
Revert "Add deploy_SimpleAccountFactory (#189)"

Dsips187:
This reverts commit 0d0230bdb117ead1e20601ea7d4887b6c7c95e50.


shahafn:
Why?

drortirosh:
Indeed, this is a sample that is only needed on local tests, and there's no need to deploy it on production networks.
But instead of removing it (and breaking tests...), it is better to disable it based on chainId - allow on chainId 1337, 31337

PR #247, opened 2023-03-20
AA-161 validate Nonce in EntryPoint

drortirosh:
account doesn't check nonce. instead, entryPoint validate and increment nonce.
nonces are two-dimensional, with 192 bit of "key" and 64 bit of sequence.

For description about this change, please see [this doc](https://docs.google.com/document/d/1MywdH_TCkyEjD3QusLZ_kUZg4ZEI00qp97mBze9JI4k/edit#)


contracts/core/NonceManager.sol
```diff
@@ -0,0 +1,37 @@
...
+    function getNonce(address sender, uint192 key)
+    public view override returns (uint256 nonce) {
+        return nonces[sender][key] | (uint256(key) << 64);
+    }
+
+    // allow an account to manually increment its own nonce.
+    // (mainly so that during construction nonce can be made non-zero,
+    // to "absorb" the gas cost of first nonce increment to 1st transaction (construction),
+    // not to 2nd transaction)
+    function incrementNonce(uint192 key) public override {
```
fangting-alchemy, line 25: Should `incrementNonce` be restricted to `entrypoint` and `account` self?

fangting-alchemy: nvm, missed they each use their own methods to update and `msg.sender` is enough to auth for external calls.

contracts/core/NonceManager.sol
```diff
@@ -0,0 +1,37 @@
...
+        nonces[msg.sender][key]++;
```
leekt: we can use ++i style here for small gas improvement

contracts/core/EntryPoint.sol
```diff
@@ -511,6 +514,11 @@ contract EntryPoint is IEntryPoint, StakeManager {
         uint256 gasUsedByValidateAccountPrepayment;
         (uint256 requiredPreFund) = _getRequiredPrefund(mUserOp);
         (gasUsedByValidateAccountPrepayment, validationData) = _validateAccountPrepayment(opIndex, userOp, outOpInfo, requiredPreFund);
+
+        if (!_validateAndUpdateNonce(mUserOp.sender, mUserOp.nonce)) {
```
leekt, line 517: I noticed some pattern on old sample accounts that it skip nonce increment when initCode.length > 0.
I assume it's for gas savings, but I don't see it here. Is it intended?

contracts/samples/SimpleAccount.sol
```diff
@@ -87,6 +80,10 @@ contract SimpleAccount is BaseAccount, UUPSUpgradeable, Initializable {
      */
     function initialize(address anOwner) public virtual initializer {
         _initialize(anOwner);
+        //increment nonce during construction.
+        // This way, the cost of moving from "0" to "1" is absorbed
+        // by first (construction) transaction, instead of during the 2nd transaction
+        entryPoint().incrementNonce(0);
```
leekt: I know that this part is just a sample but just out of curiosity, since this will be called on construction and userOp.nonce will start with 1 as far as i know since validateUserOp will check for nonce after initialize.  Is this a best practice?

And on my understanding, this nonce will be increased to 1 after validateUserOp even without this. Why bother going 0 -> 1 -> 2 for first userOp? instead of going 0 -> 1?

contracts/core/NonceManager.sol
drortirosh: afaik, with `via-ir` it doesn't make a change (did solc --asm --via-ir, and saw only differences in comments..)

contracts/core/EntryPoint.sol
drortirosh: It was an internal optimization of SimpleWallet.

contracts/samples/SimpleAccount.sol
drortirosh: It was done so that even wallets that are not created via a dedicated UserOp, will have their next operation gas "clean".
A wallet can be created by execution its initCode either by the UserOp, or by other means. 
Calling "incrementNonce" the first time will cost 20000 gas, and any further call 5000 (I ignore for now explicitly setting to a new key, which is will cost 20000 again, but that an explicit operation a wallet owner does)
Not incrementing it during construction will mean first TX after construction will cost 20000 gas more than any further TX. We prefer to make gas-per-transaction as consistent as possible, and thus make only the construction transaction different.

Calling "incrementNonce" from the factory, indeed causes double-increment of the nonce, but the overhead is minimal (~400 gas)
The side-effect is indeed an extra increment of the nonce.

contracts/core/EntryPoint.sol
```diff
@@ -345,11 +346,13 @@ contract EntryPoint is IEntryPoint, StakeManager {
     /**
      * Get counterfactual sender address.
      *  Calculate the sender contract address that will be generated by the initCode and salt in the UserOperation.
-     * this method always revert, and returns the address in SenderAddressResult error
+     * this method always revert, and returns the address (and the nonce) in SenderAddressResult error
      * @param initCode the constructor code to be passed into the UserOperation.
      */
     function getSenderAddress(bytes calldata initCode) public {
```
shahafn, line 352: Maybe rename it to `getSenderAndNonce` then?

contracts/interfaces/INonceManager.sol
```diff
@@ -0,0 +1,16 @@
...
+    /**
+     * return the next nonce for this sender
+     * @param sender the account address
+     * @param key the high 192 bit of the nonce
+     * @return nonce a full nonce to pass for next UserOp with this sender.
+     */
+    function getNonce(address sender, uint192 key)
+    external view returns (uint256 nonce);
+
+    function incrementNonce(uint192 key) external;
```
yoavw, line 26: Maybe add documentation, explaining that this function doesn't need to be called by the account?

contracts/samples/gnosis/EIP4337Manager.sol
```diff
@@ -158,7 +161,8 @@ contract EIP4337Manager is IAccount, GnosisSafeStorage, Executor {
         sig[64] = bytes1(uint8(27));
         sig[2] = bytes1(uint8(1));
         sig[35] = bytes1(uint8(1));
-        UserOperation memory userOp = UserOperation(address(safe), uint256(nonce), "", "", 0, 1000000, 0, 0, 0, "", sig);
+        uint nonce = uint256(IEntryPoint(manager.entryPoint()).getNonce(address(safe), 0));
```
yoavw: uint256 nonce

contracts/samples/gnosis/EIP4337Manager.sol
```diff
@@ -55,11 +55,6 @@ contract EIP4337Manager is IAccount, GnosisSafeStorage, Executor {
             validationData = SIG_VALIDATION_FAILED;
         }
```
yoavw, line 57: Shouldn't `validateUserOp` require `key==0` to enforce 1D nonce and preserve the previous Safe behavior?

contracts/interfaces/INonceManager.sol
drortirosh: actually, it is not needed at all. I just wonder if we should keep it for completeness, or remove it altogether..

contracts/samples/gnosis/EIP4337Manager.sol
drortirosh: good idea.

contracts/core/EntryPoint.sol
drortirosh: not relevant anymore, since this nonce was removed.

contracts/interfaces/INonceManager.sol
yoavw: I thought it was needed for the "bypass" case where an account wants to support non-EntryPoint interactions and still use the same nonce.  Is that no longer the case?

contracts/core/EntryPoint.sol
```diff
@@ -345,11 +346,12 @@ contract EntryPoint is IEntryPoint, StakeManager {
     /**
      * Get counterfactual sender address.
      *  Calculate the sender contract address that will be generated by the initCode and salt in the UserOperation.
-     * this method always revert, and returns the address in SenderAddressResult error
+     * this method always revert, and returns the address (and the nonce) in SenderAddressResult error
```
shahafn: Not returning nonce anymore

contracts/interfaces/IEntryPoint.sol
```diff
@@ -174,7 +175,7 @@ interface IEntryPoint is IStakeManager {
     /**
      * Get counterfactual sender address.
      *  Calculate the sender contract address that will be generated by the initCode and salt in the UserOperation.
-     * this method always revert, and returns the address in SenderAddressResult error
+     * this method always revert, and returns the address (and nonce) in SenderAddressResult error
```
shahafn: Doesn't return nonce

eip/EIPS/eip-4337.md
```diff
@@ -907,7 +920,7 @@ See `https://github.com/eth-infinitism/account-abstraction/tree/main/contracts`
...
+The entry point contract will need to be very heavily audited and formally verified, because it will serve as a central trust point for _all_ [EIP-4337](./eip-4337.md). In total, this architecture reduces auditing and formal verification load for the ecosystem, because the amount of work that individual _accounts_ have to do becomes much smaller (they need only verify the `validateUserOp` function and its "check signature, and pay fees" logic) and check that other functions are `msg.sender == ENTRY_POINT` gated (perhaps also allowing `msg.sender == self`), but it is nevertheless the case that this is done precisely by concentrating security risk in the entry point contract that needs to be verified to be very robust.
```
shahafn, line 923: I thought we changed EIP to ERC

contracts/interfaces/INonceManager.sol
Gooong: I think one side-effect of this function is breaking continuity of nonce, that lose the 'one nonce corresponding to one OP' property.

For example, wallet can call `_validateSignature` in OP calldata to increase nonce by 2 in one single OP.

Also, wallet can call `incrementNonce ` in `_validateSignature`, just before checking in `_validateAndUpdateNonce`, which makes it feasible to execute OP with nonce 3 when the next nonce is 2. So I suggest to check nonce `_validateAndUpdateNonce` before `_validateSignature`.

contracts/core/NonceManager.sol
Gooong, line 25: Wallet can call `incrementNonce` in `_validateSignature`, just before checking in `_validateAndUpdateNonce`, which makes it feasible to execute OP with nonce 3 when the next nonce is 2. So I suggest to check nonce `_validateAndUpdateNonce` before `_validateSignature`.

contracts/interfaces/INonceManager.sol
drortirosh: The match of "one nonce to one OP" exists, but in any case, it can't be trusted as a counter: e.g. "current nonce is # of past transactions" is not true, because:
- you can `incrementNonce()`, but also:
- you can as many different `key` values, each with another sent of transactions. 
- accounts can support "batch", which means a single UserOp can map to multiple transactions.

In short, at the system level, nonces are nothing but replay protection..

contracts/core/NonceManager.sol
drortirosh: I agree that in make sense to modify the nonce before calling a user-provided method.
Still, our opcode rules prevent the account from calling the entryPoint (including incrementNonce) from within validation.

contracts/samples/gnosis/EIP4337Manager.sol
shahafn: Are you adding that?

shahafn (approved):

contracts/samples/gnosis/EIP4337Manager.sol
drortirosh: Look at the code. It is there

PR #250, opened 2023-03-23
Use account factory to generate the init code of wallet

kaiix:
using the account factory makes the code more clear and easier to understand

### DIFF
--- a/src/AASigner.ts
+++ b/src/AASigner.ts
@@ -1,23 +1,21 @@
-import { BigNumber, Bytes, ethers, Event, Signer } from 'ethers'
-import { zeroAddress } from 'ethereumjs-util'
-import { BaseProvider, Provider, TransactionRequest } from '@ethersproject/providers'
+import { TransactionResponse } from '@ethersproject/abstract-provider'
+import { TransactionReceipt } from '@ethersproject/abstract-provider/src.ts/index'
+import { BytesLike, hexValue } from '@ethersproject/bytes'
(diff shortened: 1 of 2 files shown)


kaiix:
And, if use origin code and replace the `implementationAddress` with a deployed `SimpleAccount` https://github.com/eth-infinitism/account-abstraction/blob/ad7ce4990b3c2fe85b1ab593448e8ff50bd14a15/src/AASigner.ts#L242 like this:

```
async _deploymentTransaction (): Promise<BytesLike> {
    const implementationAddress = '0xF9aECb3577626DC1aD7ef6b8538ad52B6E409546' // a wallet implementation created from the deploy script
    const ownerAddress = await this.signer.getAddress()
    const initializeCall = new Interface(SimpleAccount__factory.abi).encodeFunctionData('initialize', [ownerAddress])
    return new ERC1967Proxy__factory(this.signer).getDeployTransaction(implementationAddress, initializeCall).data!
  }
```

`yarn runop` still encounter an error, the account created from the entrypoint has different address with it computed from the AASigner

```
Error: account validation failed: AA14 initCode must return sender
```

 I don't known what I misunderstood or there are other modifications needed?

drortirosh (dismissed):

kaiix:
fixed lint error

drortirosh:
sorry for late response... 
it looks ok, but lately we update to latest contracts.
please merge develop to re-run the tests.

drortirosh (approved):

PR #257, opened 2023-04-02
AA-159 prevent recursive call into handleOps

drortirosh:
### The Problem:
Currently, it is possible to make recursive calls to handleOps() during UserOp execution.

In order to associate events with UserOps, we parse all events since the previous UserOperationEvent.  This means that recursive handleOp will break parsing, potentially causing events to be associated with the wrong UserOp.

Note that there is no execution risk: everything is executed in order, as it should. It is only the EXTERNAL view on the events that gets split the wrong way.

#### Example:

Consider a “flashloan” contract that emits “BeforeLoan” and “AfterLoan” events.
A user submits a UserOp that calls the flashloan contract, with a transaction that recursively calls `handleOps` with another UserOp which emits a “Transfer” event.
The events sent are:
1. BeforeLoan
2. Transfer
3. UserOperationEvent (inner)
4. AfterLoan
5. UserOperationEvent (outer)

When parsing the events, the transactions look like:
UserOp1: BeforeLoan, Transfer
UserOp2: AfterLoan
The above split is the way the bundlers (using getUserOperationReceipt) would currently return the events.
Anyone parsing the flashloan events, will get confused - a single atomic flashloan is broken over 2 UserOps.

Additional issues with recursion:

* With EOA, it is impossible for an account to make a "recursive" call. Making it possible with AA could break user assumptions.
* Even with contracts, there is a strong implicit assumption that there are no recursions, and some famous hacks happened when this assumption didn't hold.
* Allowing recursive calls on accounts means that an "execute" of an account might be called from within the execution of the same account (though this arguably has to be accepted and signed (twice) by the wallet).
* Paymasters might assume they are not called recursively when handling the "postop" operation.
* Some chains enshrine account-abstraction into the protocol. In such cases, the non-recursion is also prevented at the protocol level, and we would like to keep "feature parity" across different networks.

### Solution:

One possible solution is to "enshrine" recursions, and provide mechanisms for bundlers and applications to detect them.  However, this would put the burden of handling recursions, however rare, by each account and each paymaster implementation.  In practice, accounts and paymasters might end up spending gas on a reentrancy guard in each UserOp.

A more efficient way to solve the issue, along with the additional issues above, is to prevent recursion entirely, by adding non-reentrancy guard in EntryPoint itself.  The cost of the reentrancy guard is split among all UserOps rather than having each account spend gas on its own protection.


contracts/utils/ReentrancyGuard.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.8.0) (security/ReentrancyGuard.sol)
+// Modified to use StorageSlot
+pragma solidity ^0.8.12;
+
+import "@openzeppelin/contracts/utils/StorageSlot.sol";
```
forshtat: @drortirosh What is the point of copying the source code of 'ReentrancyGuard.sol' into the 'account-abstraction' repo if it still imports a file ('StorageSlot.sol') presumably from 'node_modules?

drortirosh: 1. There are no "openzeppelin" imports from the core, only from the samples
2. there implementation doesn't use their own "StorageSlots", but use global members

shahafn (approved):

PR #259, opened 2023-04-05
add `virtual` to EntryPoint functions

taxio:
Thanks for the wonderful implementation and library.

EntryPoint is very complex, so when creating an extended EntryPoint with a specific context, I want to base it as much as possible on your implementation.

Therefore, I added `virtual` to the EntryPoint functions so that I can override each method.

### DIFF
--- a/contracts/core/EntryPoint.sol
+++ b/contracts/core/EntryPoint.sol
@@ -42,7 +42,7 @@ contract EntryPoint is IEntryPoint, StakeManager {
      * @param beneficiary the address to receive the fees
      * @param amount amount to transfer.
      */
-    function _compensate(address payable beneficiary, uint256 amount) internal {
+    function _compensate(address payable beneficiary, uint256 amount) internal virtual {
         require(beneficiary != address(0), "AA90 invalid beneficiary");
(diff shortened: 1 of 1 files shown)


shahafn:
EntryPoint is supposed to be a known singleton contract. Can you please elaborate on why you'd need to extend and deploy your own EntryPoint?

drortirosh:
yes. the entrypoint is deployed as a singleton. if you always modify the code locally to have your own, but it is discouraged.

yoavw:
@taxio since EntryPoint is a singleton and there should be only one per blockchain, it would be good if you share what you think is missing from it.  If you deploy a separate one, any account that uses it will be incompatible with the mempool so bundlers will not propagate its transactions.  If something is missing from EntryPoint, it would be better to discuss it and see if it should be added to the singleton.

taxio:
Thank you everyone for your comments. I didn't realize that EntryPoint is a singleton.

I casually submitted a PR thinking that various extended EntryPoints would appear for each business idea, but considering it's a singleton, I should have discussed whether or not to include the functionality in EntryPoint itself.

PR #261, opened 2023-04-11
remove senderNonce from sample verifying paymaster

livingrockrises:
The reason for adding a nonce to the paymaster was to handle cases where the nonce was not unique: a wallet could support static nonce, and it could re-use the VerifyingPaymaster validation multiple times.

But now nonces uniqueness is validated by entrypoint. So a paymaster no longer need to check for uniqueness by itself


drortirosh:
This implementation is broken: if the VerifyingPaymaster doesn't sign all fields, then an account can grief it, by specifying larger values for gas price and limits.
A VerifyingPaymaster has to approve entire data structure.

livingrockrises:
> This implementation is broken: if the VerifyingPaymaster doesn't sign all fields, then an account can grief it, by specifying larger values for gas price and limits. A VerifyingPaymaster has to approve entire data structure.

fixed.

drortirosh (approved):

Issue #264, opened 2023-04-17
`callAndRevert` name is misleading


0xPhaze:
One would expect `callAndRevert` to make a call and to always revert.

https://github.com/eth-infinitism/account-abstraction/blob/9b5f2e4bb30a81aa30761749d9e2e43fee64c768/contracts/utils/Exec.sol#L64-L69

It only reverts if the call was not a success, just like a normal call. Maybe naming it `callSafe` or similar would be better fitting, since its main purpose is to protect agains a return data bomb.

shahafn:
While changing the name may be beneficial, since it's changing the EntryPoint address, and it's a low priority change, we'll update it in the next release

PR #265, opened 2023-04-17
Changed function name form callAndRevert to callSafe

Jayrajrodage:
https://github.com/eth-infinitism/account-abstraction/issues/264#issue-1670662181


shahafn:
See my comment on #264

PR #267, opened 2023-04-18
add nonReentrant in simulateHandleOp

Gooong:
Add nonReentrant in simulateHandleOp to keep simulation and execution consistent.


drortirosh:
Yes, you're right that it should be added to simulation. But we don't want to change the EntryPoint just for that now.
There is a workaround, however, there is a way to add a "stateDiff" to a contract when doing eth_call (or traceCall).
This way, we can set the flag and prevent inconsistent execution.

Issue #274, opened 2023-05-11
Potential attack from staked account for token paymaster


huaweigu:
Terminology
`account` - a smart contract wallet usually hold by the user

Scenario
Two userOps from the same sending account are in the same bundler. 1st userOp is trying to withdraw all the token from the sender.

During a discussion with 4337 team (@drortirosh @yoavw @shahafn ), we found that if the sending account is staked with entry point contract, there's a chance that paymaster could get attacked by that staked account. For example, the bundler could accept two userOps from the same account (if they're staked), both userOps would pass the 1st simulation before they're added into mempool. However, the 2nd validation (before bundling) [would potentially revert](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/EntryPoint.sol#L455-L459) when 2nd userOp is being validated.

Even if the paymaster exit via `sigFailed` via a hacky way of handling attack, the code [would still revert.](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/EntryPoint.sol#L477-L480)
(note: the real `sigFailed` would fail in 1st simulation)

Both reverts would attribute the fault to paymaster, so paymaster could potentially [get banned by the bundler's reputation system](https://github.com/eth-infinitism/bundler/blob/main/packages/bundler/src/modules/BundleManager.ts#L127).

Theoretically, paymaster could build its own reputation system just like bundler service to prevent the account from using it again, but we're also wondering if the entry point/bundler could do something to prevent the account from using paymaster again, e.g. by allowing graceful exit from paymaster for certain reason code.

CC @kerikailiwang

yoavw:
Staked accounts are a seldom used use case.  And since the account is staked and has a reputation, it can be blamed for the failure instead of the paymaster.  It'll be the staked account's responsibility to defend its reputation against malicious paymasters by either blocking paymasters altogether (`require(op.paymaster==address(0))`) or by maintaining a whitelist of trusted paymasters.

TL;DR: we should add a reputation rule for staked accounts, assigning the blame of any failure to the account itself.

wildmolasses:
> Staked accounts are a seldom used use case.

Shouting out that as an implementer i'm pursuing an idea that relies on staked accounts. was talking with @shahafn about it recently. i'd like to know whether multiple userOps from same sender in same bundle will be possible or get deprecated.

yoavw:
[...]

PR #276, opened 2023-05-12
Disallow paymaster for staked account

huaweigu:
Related to https://github.com/eth-infinitism/account-abstraction/issues/274, staked account could have more than one userOp in the same bundler, so it could potentially grieve the paymaster if two userOps (from the same sender account) are trying to empty its balance.

Tests will be added if 4337 team is okay with this approach. Another potential approach recommended by Dror is whitelisting a list of paymasters for staked account. However, it's very rare for normal accounts to stake with entry point so I went with the current approach. 
```
A sender is exempt from this rule and may have multiple UserOperations in the pool and in a batch if it is staked (see reputation, throttling and banning section below), but this exception is of limited use to normal accounts. 
```

Please feel free to point out the missing important use cases that a staked account would need paymaster in comments.

### DIFF
--- a/contracts/core/EntryPoint.sol
+++ b/contracts/core/EntryPoint.sol
@@ -437,6 +437,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard 
     function _validatePaymasterPrepayment(uint256 opIndex, UserOperation calldata op, UserOpInfo memory opInfo, uint256 requiredPreFund, uint256 gasUsedByValidateAccountPrepayment)
     internal returns (bytes memory context, uint256 validationData) {
     unchecked {
+        // if the sender account has staked (rare to see), they could potentially grieve the paymaster with an insufficient account balance
+        // because the reverts would attribute the fault to paymaster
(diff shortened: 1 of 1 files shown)


yoavw:
Thanks, but EntryPoint is not the right place for this.  None of the reputation rules are on-chain.  Reputation (and the stake which is used for its sybil resistance) is maintained by each bundler separately, and is implemented locally by the bundler.

We'll add a reputation rule that doesn't block staked accounts from using paymasters, but assigns failures to the staked account: https://github.com/eth-infinitism/account-abstraction/issues/274#issuecomment-1546006743

The place to add this check is in the reference bundler code and in the bundler test suite to verify that bundlers implement it correctly.

As a general rule in the 4337 implementation, on-chain attacks are handled by on-chain code (e.g. protecting the account or the paymaster from theft), and off-chain attacks are handled by off-chain code (e.g. DoS against bundlers is mitigated by off-chain validation rules and staked reputation).  No reason to spend gas on protecting an off-chain entity that can defend itself by cheaper means.

PR #286, opened 2023-05-18
New TokenPaymaster

forshtat:


contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,207 @@
...
+            uint256 cachedPriceWithMarkup = cachedPrice * PRICE_DENOMINATOR / priceMarkup;
+            if (paymasterAndDataLength == 32) {
+                uint256 clientSuppliedPrice = uint256(bytes32(userOp.paymasterAndData[20 : 52]));
+                if (clientSuppliedPrice < cachedPriceWithMarkup) {
+                    // note: smaller number means 'more ether per token'
+                    cachedPriceWithMarkup = clientSuppliedPrice;
+                }
+            }
+            uint256 tokenAmount = weiToToken(preChargeNative, cachedPriceWithMarkup, false);
+            SafeERC20.safeTransferFrom(token, userOp.sender, address(this), tokenAmount);
```
huaweigu, line 141: minor, IIRC from our past debug experience, SafeERC20 uses banned [selfbalance opcode](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dfef6a68ee18dbd2e1f5a099061a3b8a0e404485/contracts/utils/Address.sol#L134)

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,207 @@
...
+            if (preCharge > actualTokenNeeded) {
+                // If the initially provided token amount is greater than the actual amount needed, refund the difference
+                SafeERC20.safeTransfer(
+                    token,
+                    userOpSender,
+                    preCharge - actualTokenNeeded
+                );
+            } else {
+                // If the token amount is not greater than the actual amount needed, revert to remove incentive to cheat
+                revert("TPM: preCharge was too low");
```
huaweigu: I wonder if it's worth of attempting to transfer the owed token (preCharge - actualTokenNeeded) from sender here

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,207 @@
...
+        setTokenPaymasterConfig(_tokenPaymasterConfig);
+        transferOwnership(_owner);
     }
 
+    /// @notice Updates the configuration for the Token Paymaster.
+    /// @param _tokenPaymasterConfig The new price markup percentage (1e6 = 100%).
+    function setTokenPaymasterConfig(
+        TokenPaymasterConfig memory _tokenPaymasterConfig
+    ) public onlyOwner {
+        require(_tokenPaymasterConfig.priceMarkup <= 1900000, "TPM: price markup too high");
```
huaweigu: Out of curiosity, how do we pick the upper bound (1900000)?

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,207 @@
...
+            uint256 actualChargeNative = actualGasCost + REFUND_POSTOP_COST * maxFeePerGas;
+            uint256 actualTokenNeeded = weiToToken(actualChargeNative, cachedPriceWithMarkup, false);
+            if (preCharge > actualTokenNeeded) {
+                // If the initially provided token amount is greater than the actual amount needed, refund the difference
+                SafeERC20.safeTransfer(
+                    token,
+                    userOpSender,
+                    preCharge - actualTokenNeeded
+                );
+            } else {
```
huaweigu: nitpick: maybe we shouldn't revert if preCharge == actualTokenNeeded (edge case)

contracts/samples/TokenPaymaster.sol
forshtat: Thanks, I did not know that! Created a small PR in the OpenZeppelin repo, hope this is all there is to this issue:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4271

I hope to avoid having to copy `SafeERC20.sol` library into the `account-abstraction` repo, but maybe we will have no choice.

contracts/samples/TokenPaymaster.sol
forshtat: None of these hard-coded parameters are selected yet, but in my opinion the maximum possible value for the markup should just be there to prevent human error or malicious configuration.

For example, Pimlico Paymaster used `120e4` which sounds to me like it may be a bit too limiting:
https://github.com/pimlicolabs/erc20-paymaster-contracts/blob/c9911402a49928398441b07fb9caa9f4b4d0352f/src/PimlicoERC20Paymaster.sol#L68

contracts/samples/TokenPaymaster.sol
forshtat: Right, thanks, will fix. To be clear, there is almost zero chance of that happening.

contracts/samples/TokenPaymaster.sol
forshtat: I think this may actually work, will bring this up with the team. So far I did add an attempt to withdraw the owed amount (`actualTokenNeeded - preCharge`) from the UserOp sender to the Paymaster.

contracts/samples/TokenPaymaster.sol
livingrockrises: maybe 2e6 is good upper limit

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,213 @@
...
+                SafeERC20.safeTransferFrom(
+                    token,
+                    userOpSender,
+                    address(this),
+                    actualTokenNeeded - preCharge
+                );
+            }
+
+            emit UserOperationSponsored(userOpSender, actualTokenNeeded, actualGasCost, cachedPrice);
+            refillEntryPointDeposit(_cachedPrice);
```
livingrockrises, line 196: Does this mean router swapping cost is dumped on to the users? 
since it is conditional postOp cost would change 
if not user who is paying for this computation - paymaster's gas deposit?

contracts/samples/TokenPaymaster.sol
forshtat: @livingrockrises Agreed. Will set for 200% for now. Will also make it explicitly related to the `PRICE_DENOMINATOR` value.

contracts/samples/TokenPaymaster.sol
forshtat: A single user should not get stuck with the bill for the Paymaster's swap operation on its own.
I am not sure but I think what we should do is to set the `POSTOP_COST` parameter such that it kind of spreads the cost of these periodic swaps over all other transactions.
Of course this way it may end up accumulating some loss or profit over time if left completely unattended.

contracts/samples/TokenPaymaster.sol
leekt: maybe we can use solady in that case https://github.com/Vectorized/solady/blob/f9a6cc652dc8b37387590588a78644abfa4be3ac/src/utils/SafeTransferLib.sol#L238

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,213 @@
...
+    function _postOp(PostOpMode mode, bytes calldata context, uint256 actualGasCost) internal override {
+        if (mode == PostOpMode.postOpReverted) {
+            return;
+        }
+        unchecked {
+            uint256 priceMarkup = tokenPaymasterConfig.priceMarkup;
+            uint256 preCharge = uint256(bytes32(context[0 : 32]));
+            uint256 maxFeePerGas = uint256(bytes32(context[32 : 64]));
+            address userOpSender = address(bytes20(context[64 : 84]));
+            if (mode == PostOpMode.postOpReverted) {
```
leekt, line 166: This branch cannot be reached since there is a immediate return

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,213 @@
...
     }
 
-    // when constructing an account, validate constructor code and parameters
-    // we trust our factory (and that it doesn't have any other public methods)
-    function _validateConstructor(UserOperation calldata userOp) internal virtual view {
-        address factory = address(bytes20(userOp.initCode[0 : 20]));
-        require(factory == theFactory, "TokenPaymaster: wrong account factory");
+    /// @notice If necessary this function uses this Paymaster's token balance to refill the deposit on EntryPoint
+    function refillEntryPointDeposit(uint256 _cachedPrice) private {
+        uint256 currentEntryPointBalance = entryPoint.balanceOf(address(this));
```
leekt, line 200: STATICCALL to entrypoint is not allowed
```
Limitation on “CALL” opcodes (CALL, DELEGATECALL, CALLCODE, STATICCALL):
must not use value (except from account to the entrypoint)
must not revert with out-of-gas
destination address must have code (EXTCODESIZE>0)
cannot call EntryPoint’s methods, except depositFor (to avoid recursion)
```

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,213 @@
...
+            if (mode == PostOpMode.postOpReverted) {
+                emit PostOpReverted(userOpSender, preCharge);
+                // Do nothing here to not revert the whole bundle and harm reputation
+                return;
+            }
+            uint256 _cachedPrice = updateCachedPrice(false);
+        // note: as price is in ether-per-token and we want more tokens increasing it means dividing it by markup
+            uint256 cachedPriceWithMarkup = _cachedPrice * PRICE_DENOMINATOR / priceMarkup;
+        // Refund tokens based on actual gas cost
+            uint256 actualChargeNative = actualGasCost + REFUND_POSTOP_COST * maxFeePerGas;
```
livingrockrises: instead of using maxFeePerGas, does it make sense to use tx.gasprice ?

contracts/samples/TokenPaymaster.sol
livingrockrises: If paymaster is always charging premium, POSTOP_COST can just be everything unaccounted on-chain on entry point and postOp - swap executionGas.

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,213 @@
...
+    event ConfigUpdated(TokenPaymasterConfig tokenPaymasterConfig);
+
+    event UserOperationSponsored(address indexed user, uint256 actualTokenCharge, uint256 actualGasCost, uint256 actualTokenPrice);
+
+    event PostOpReverted(address indexed user, uint256 preCharge);
+
+    event Received(address indexed sender, uint256 value);
+
+    /// @notice All 'price' variables are multiplied by this value to avoid rounding up
+    uint256 private constant PRICE_DENOMINATOR = 1e26;
```
livingrockrises, line 53: what is the rationale behind using 1e26?

contracts/samples/TokenPaymaster.sol
forshtat: You are right, thanks! However, we cannot just use `tx.gasprice` as the bundler can manipulate this value. I think we should use `userOp.gasPrice()` here. Updated.

contracts/samples/TokenPaymaster.sol
forshtat: Thanks. Fixed.

contracts/samples/TokenPaymaster.sol
forshtat: At this point constants are actually pretty random and will be finalised after we make some gas usage optimisations. Do you have suggestions on what value should be used and why?

contracts/samples/TokenPaymaster.sol
livingrockrises: userop.gasPrice() touches BASEFEE which forbidden opcode. this would be fine in postOp(), but you shouldn't calculate it in validatePaymasterUserOp and pass it in context.
Instead pass maxPriorityFee and maxFee in context and calculate like below in the postOp

```
uint256 opGasPrice;
        unchecked {
            if (maxFeePerGas == maxPriorityFeePerGas) {
                opGasPrice = maxFeePerGas;
            } else {
                opGasPrice = Math.min(maxFeePerGas, maxPriorityFeePerGas + block.basefee);
            }
        }
```

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,212 @@
...
+    function _validatePaymasterUserOp(UserOperation calldata userOp, bytes32, uint256 requiredPreFund)
+    internal
+    override
+    returns (bytes memory context, uint256 validationResult) {unchecked {
+            uint256 priceMarkup = tokenPaymasterConfig.priceMarkup;
+            uint256 paymasterAndDataLength = userOp.paymasterAndData.length - 20;
+            require(paymasterAndDataLength == 0 || paymasterAndDataLength == 32,
+                "TPM: invalid data length"
+            );
+            uint256 gasPrice = userOp.gasPrice();
```
livingrockrises: this will revert on the bundler
cc @leekt

contracts/samples/TokenPaymaster.sol
forshtat: Right 😀

contracts/samples/TokenPaymaster.sol
leekt: yup this will revert for sure

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,224 @@
...
+            uint256 actualTokenNeeded = weiToToken(actualChargeNative, cachedPriceWithMarkup);
+            if (preCharge > actualTokenNeeded) {
+                // If the initially provided token amount is greater than the actual amount needed, refund the difference
+                SafeERC20.safeTransfer(
+                    token,
+                    userOpSender,
+                    preCharge - actualTokenNeeded
+                );
+            } else if (preCharge < actualTokenNeeded) {
+                // Attempt to cover Paymaster's gas expenses by withdrawing the 'overdraft' from the client
```
leekt, line 185: Not sure about this part. Won't this make `clientSuppliedPrice` invalid? `clientSuppliedPrice` should be the guard for how much token user is willing to pay but this part will end up forcing user to pay the full slippage

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,224 @@
...
+
+    event PostOpReverted(address indexed user, uint256 preCharge);
+
+    event Received(address indexed sender, uint256 value);
+
+    /// @notice All 'price' variables are multiplied by this value to avoid rounding up
+    uint256 private constant PRICE_DENOMINATOR = 1e26;
+
+    /// @notice Estimated gas cost for refunding tokens after the transaction is completed
+    uint256 public constant REFUND_POSTOP_COST = 40000;
```
drortirosh: postop_cost depends on price of token transfer, which varies with different token implementation.
to be accurate, this is better be a constructor parameter

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,224 @@
...
+    function setTokenPaymasterConfig(
+        TokenPaymasterConfig memory _tokenPaymasterConfig
+    ) public onlyOwner {
+        require(_tokenPaymasterConfig.priceMarkup <= 2 * PRICE_DENOMINATOR, "TPM: price markup too high");
+        require(_tokenPaymasterConfig.priceMarkup >= PRICE_DENOMINATOR, "TPM: price markup too low");
+        tokenPaymasterConfig = _tokenPaymasterConfig;
+        emit ConfigUpdated(_tokenPaymasterConfig);
     }
 
+    function setOracleConfiguration(
```
drortirosh: do we need setters? if the data is immutable, it the logic much cheaper.
There is no harm at re-deploy the paymaster with different configuration.

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,224 @@
...
+                uint256 clientSuppliedPrice = uint256(bytes32(userOp.paymasterAndData[20 : 52]));
+                if (clientSuppliedPrice < cachedPriceWithMarkup) {
+                    // note: smaller number means 'more ether per token'
+                    cachedPriceWithMarkup = clientSuppliedPrice;
+                }
+            }
+            uint256 tokenAmount = weiToToken(preChargeNative, cachedPriceWithMarkup);
+            SafeERC20.safeTransferFrom(token, userOp.sender, address(this), tokenAmount);
+            context = abi.encodePacked(tokenAmount, userOp.maxFeePerGas, userOp.maxPriorityFeePerGas, userOp.sender);
+            validationResult = 0;
```
drortirosh: Better return validUntil as `cachedPriceTimestamp + maxCacheAge`

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,224 @@
...
-        return valueEth / 100;
+    /// @notice Performs post-operation tasks, such as updating the token price and refunding excess tokens.
+    /// @dev This function is called after a user operation has been executed or reverted.
+    /// @param mode The post-operation mode (either successful or reverted).
+    /// @param context The context containing the token amount and user sender address.
+    /// @param actualGasCost The actual gas cost of the transaction.
+    function _postOp(PostOpMode mode, bytes calldata context, uint256 actualGasCost) internal override {
+        unchecked {
+            uint256 priceMarkup = tokenPaymasterConfig.priceMarkup;
+            uint256 preCharge = uint256(bytes32(context[0 : 32]));
```
drortirosh: `abi.decode()` is that bad?
it surely much clearer to understand.

contracts/samples/utils/OracleHelper.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    /// @notice Updates the token price by fetching the latest price from the Oracle.
+    function updateCachedPrice(bool force) public returns (uint256 newPrice) {
+        uint256 cacheTimeToLive = oracleHelperConfig.cacheTimeToLive;
+        uint256 priceUpdateThreshold = oracleHelperConfig.priceUpdateThreshold;
+        IOracle tokenOracle = oracleHelperConfig.tokenOracle;
+        IOracle nativeOracle = oracleHelperConfig.nativeOracle;
+
+        uint256 cacheAge = block.timestamp - cachedPriceTimestamp;
+        if (!force && cacheAge <= cacheTimeToLive) {
```
drortirosh, line 81: better be moved up, and avoid loading storage fields before the quick bail-out.

contracts/samples/TokenPaymaster.sol
forshtat: This paymaster will accumulate client base by getting 'approval' in ERC-20 token used. Deploying a new paymaster just to switch to a different price oracle will lose all existing clients until they migrate their approval.
Do you want me to make the fields immutable and make the paymaster an upgradeable proxy instead?

contracts/samples/utils/OracleHelper.sol
forshtat: Fixed.

contracts/samples/TokenPaymaster.sol
forshtat: Fixed. I was trying to same some bytes by doing `abi.encodePacked`.

contracts/samples/TokenPaymaster.sol
forshtat: Done.

contracts/samples/TokenPaymaster.sol
forshtat: @drortirosh
Bumping this one. Should we?

contracts/samples/TokenPaymaster.sol
forshtat: Done.

contracts/samples/TokenPaymaster.sol
forshtat: Well, yes, with this implementation the user-supplied price can only increase the pre-charge for the user, but does not set an upper limit of the token price, so once the users approve this paymaster they basically agree to pay for gas at any current market price. 

@drortirosh what do you think about this feature?

CC: @huaweigu

contracts/samples/TokenPaymaster.sol
```diff
@@ -142,8 +145,12 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
             }
             uint256 tokenAmount = weiToToken(preChargeNative, cachedPriceWithMarkup);
             SafeERC20.safeTransferFrom(token, userOp.sender, address(this), tokenAmount);
-            context = abi.encodePacked(tokenAmount, userOp.maxFeePerGas, userOp.maxPriorityFeePerGas, userOp.sender);
-            validationResult = 0;
+            context = abi.encode(tokenAmount, userOp.maxFeePerGas, userOp.maxPriorityFeePerGas, userOp.sender);
+            validationResult = _packValidationData(
+                false,
+                uint48(cachedPriceTimestamp + tokenPaymasterConfig.priceMaxAge),
```
drortirosh, line 145: storage optimization: instead referencing 2 storage slots on each call, use 
`cachedPriceExpirationTimestamp = timestamp + priceMaxAge`

test/samples/OracleHelper.test.ts
```diff
@@ -102,6 +102,8 @@ describe.only('OracleHelper', function () {
     const owner = await ethersSigner.getAddress()
 
     const tokenPaymasterConfig: TokenPaymaster.TokenPaymasterConfigStruct = {
+      priceMaxAge: 86400,
```
drortirosh, line 110: ^^ remove "only" from `describe.only('OracleHelper')`

contracts/index.ts
```diff
@@ -1,2 +0,0 @@
-export { UserOperationStruct } from './types/EntryPoint'
```
drortirosh, line 1: true, should be removed (along with the code to use it during "yarn publish"), but unrelated to TokenPaymaster.

contracts/test/TestUniswap.sol
```diff
@@ -0,0 +1,60 @@
...
+            params.tokenOut
+        );
+        IERC20(params.tokenIn).transferFrom(msg.sender, address(this), params.amountIn);
+        IERC20(params.tokenOut).transfer(params.recipient, amountOut);
+        return amountOut;
+    }
+
+    /// @notice Simplified code copied from here:
+    /// https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/PeripheryPayments.sol#L19
+    function unwrapWETH9(uint256 amountMinimum, address recipient) public payable {
```
drortirosh, line 48: unused?

contracts/samples/TokenPaymaster.sol
```diff
@@ -1,113 +1,233 @@
...
+                // If the transfer reverts also revert the 'postOp' to remove the incentive to cheat
+                SafeERC20.safeTransferFrom(
+                    token,
+                    userOpSender,
+                    address(this),
+                    actualTokenNeeded - preCharge
+                );
+            }
+
+            emit UserOperationSponsored(userOpSender, actualTokenNeeded, actualGasCost, cachedPrice);
```
drortirosh: I think this is `_cachedPrice`, not `cachedPrice`

contracts/samples/TokenPaymaster.sol
drortirosh: @livingrockrises 
A paymaster is expected to set `refundPostopCost` to actual cost of postOp, with faction of the cost of the uniswap, since the swap is expected to happen only every "N" transaction.

contracts/samples/TokenPaymaster.sol
drortirosh: As a user, I trust the config values the paymaster have - but I really can't fully trust it, since it can change its configuration at will  (e.g. increase the postOp cost, or switch to a malicious oracle)

I want to trust an address given by some authority, and know this address can't change.
Better yet - I want to know this address define all parameters values.

Also other parameters (like minEPbalance) are not expected to be modified much.

drortirosh (commented):
Should also add gas-check test in `gas-calc/`

contracts/index.ts
forshtat: This was messing up the environment for me though.

contracts/test/TestUniswap.sol
forshtat: Why not? This is used by the `UniswapHelper.sol` file:

```
    function unwrapWeth(uint256 amount) internal {
        IPeripheryPayments(address(uniswap)).unwrapWETH9(amount, address(this));
    }
```

contracts/samples/TokenPaymaster.sol
forshtat: Please elaborate where does the `timestamp` variable come from?

contracts/samples/utils/OracleHelper.sol
```diff
@@ -67,7 +67,7 @@ abstract contract OracleHelper {
 
     function _setOracleConfiguration(
         OracleHelperConfig memory _oracleHelperConfig
-    ) internal {
+    ) private {
```
drortirosh, line 70: leaving this method (called only from constructor) means that we still pay the SLOAD when using these parameters.
(and you can't make a member "immutable" if it is set from a private method)

Currently, the paymaster uses 15 slots (14, if we exclude "owner")
(checked with `solc --storage-layout`)
So it uses 14 storage slots on each operation, which is at minimum 14*2100 = 29400 gas .
But I suggest adding a "gascalc" test before optimizing it.

drortirosh (approved):

contracts/samples/TokenPaymaster.sol
livingrockrises: at Biconomy we are returning a fee quote to the sender and allow them to update userop.calldata with the approval transaction with necessary amount each time. then partialUserOp is filled with paymasterAndData and signature. SDK account APIs can facilitate this for decoding calldata and append approval transaction in a batch and have new callData, callGasLimit

Blackjack789 (approved):

PR #292, opened 2023-05-29
refactor: check for token support before `transferFrom`

therealharpaljadeja:
It would be better to check for token support before you even transfer the tokens, this will save gas.


drortirosh (approved):
yes, I guess you're right.

PR #295, opened 2023-06-06
fix: deterministicDeployment for replay protection chain

hangleang:
**Current Behavior**: fail to deploy on replay protection chains, cause by legacy pre-EIP155 not supported

**PR**: use [Singleton factory](https://github.com/safe-global/safe-singleton-factory) from Safe to enable such deployment on replay protection chains, by include `chainId` into signed transactions

### DIFF
--- a/deploy/1_deploy_entrypoint.ts
+++ b/deploy/1_deploy_entrypoint.ts
@@ -1,16 +1,13 @@
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 import { DeployFunction } from 'hardhat-deploy/types'
-import { Create2Factory } from '../src/Create2Factory'
-import { ethers } from 'hardhat'
 
 const deployEntryPoint: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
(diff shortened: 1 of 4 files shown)


drortirosh:
We prefer to keep the entrypoint deployment permissionless, and not rely on 3rd-party deployers.
There is an effort to push networks to support the de-factory standard deployers by deploying it using a "nonstandard state change".
I know that some networks (e.g. coinbase) already did it - that is, even though they support EIP155 from day one, they still deployed arachnid's deployer.

hangleang:
Thank you for your response.

So how can we deploy on such of chains? currently, I use the Singleton factory to sign the transaction to get the deterministic deployment, then add it into `customSingletonFactoryInfo` in `hardhat.config.ts`. Could you please give some suggestions on the better and safer approach?

hangleang:
I can use the custom deterministic deployment without relying on 3rd-party deployer like you said. 

Here is the [Reference](https://github.com/wighawag/hardhat-deploy#4-deterministicdeployment-ability-to-specify-a-deployment-factory)

PR #296, opened 2023-06-07
Add GPL LICENSE file

drortirosh:


forshtat (approved):

pad01g:
@drortirosh 
Is it possible to loosen the license to MIT in future or do you have strong reasons to stick to GPLv3?
Edit: I found out [some libraries](https://www.npmjs.com/package/@gnosis.pm/safe-contracts) are GPLv3. I guess that's why.

PR #305, opened 2023-06-23
Add yarn.lock consistency check to GitHub Actions

MikeMoldawsky:
# Summary
This PR adds a new job to GitHub Actions, which verifies the consistency of `yarn.lock` with `package.json`, helping to avoid unintentional yarn errors in local environments.

# Description
The new job runs `yarn install --frozen-lockfile` and checks for modifications in `yarn.lock`. If there are changes, it indicates an inconsistency between `yarn.lock` and `package.json`.

This job runs as a first step in the workflow, prompting contributors to update their `yarn.lock` when required, thus preventing potential build failures.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,33 @@ env:
 
 # todo: extract shared seto/checkout/install/compile, instead of repeat in each job.
 jobs:
+  check-yarn-lock:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - name: Cache Node.js modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Check for changes
+        run: |
+          if git diff --exit-code; then
+              echo "No changes in yarn.lock"
+          else
+              echo "yarn.lock has changed"
+              exit 1
+          fi
 
   test:
     runs-on: ubuntu-latest


MikeMoldawsky:
I see dependency errors in my local environment when installing the project.

I'd suggest the following after this PR is (hopefully 😅)  merged:
1. Someone from the core team should run `yarn install` and ensure the `yarn.lock` is up to date. If not, we should open a PR to make sure it's up to date to make it easier for people to contribute to the repo.
2. I'll open a following PR updating the _README.md_ with instructions on how to run the project - it'll help new devs for a quicker onboarding process.

cc @yoavw @drortirosh 

P.S.
I can help with updating the yarn.lock and resolving the issues if I'll get a confirmation from a core dev that the problems occurs in their local environment as well.

drortirosh:
@MikeMoldawsky 
I'm trying to understand this PR..
Yes, it makes sense to validate the lock file is consistent, but maybe I don't understand yarn well enough.
Why do we need to run it as a separate "action"? why not just add the "--frozen-lockfile" to the current build?

MikeMoldawsky:
> @MikeMoldawsky I'm trying to understand this PR.. Yes, it makes sense to validate the lock file is consistent, but maybe I don't understand yarn well enough. Why do we need to run it as a separate "action"? why not just add the "--frozen-lockfile" to the current build?

It's a common practice to check the lock file in a separate job. Specifically, you want the job that runs the tests to mock the production behavior, and `--frozen-lockfile` flag isn't used in production.

You could add it to an existing job, but it's an anti-pattern. Similarly, you could transform `check-gas`, `lint` jobs to additional steps in the `test` job.

Happy to change the PR if you prefer it as steps inside another job.

.github/workflows/build.yml
```diff
@@ -12,6 +12,33 @@ env:
...
+        uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - name: Cache Node.js modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
```
drortirosh, line 29: any special reason you're using npm cache, and not Yarn's ?

drortirosh: we're working with yarn. why does this caching uses npm's lock file?

PR #306, opened 2023-06-24
[Fix] Add "etherscan" to HardhatUserConfig to resolve Type Error

MikeMoldawsky:
In this PR, we address an error encountered while running `yarn tsc` due to the incorrect ordering of import statements.

### The Issue:
Running `yarn tsc` yielded a Type error related to the `etherscan` property in `HardhatUserConfig`, as it wasn't recognized.

```typescript
hardhat.config.ts:70:3 - error TS2322: Type '{
   ...
  etherscan: {
    apiKey: process.env.ETHERSCAN_API_KEY
  };
}' is not assignable to type 'HardhatUserConfig'.
  Object literal may only specify known properties,
   and 'etherscan' does not exist in type 'HardhatUserConfig'.
  ```

### The Fix:

The root cause was traced back to the order of import statements, which affected the application of type extensions. 

By reordering ['@nomiclabs/hardhat-etherscan'](https://www.npmjs.com/package/@nomiclabs/hardhat-etherscan) import to the top of the file we can be sure that the `etherscan` property will exist on the `HardhatUserConfig`.

This enables the successful execution of `yarn tsc` without the previously encountered error. 

### Error Deep Dive:

Below we can see that `'hardhat-deploy'` package (which was imported before `@nomiclabs/hardhat-etherscan`) is copying the `userConfig.etherscan` to `userConfig.verify.etherscan`. 

This means, at the bare minimum, the `@nomiclabs/hardhat-etherscan` should be imported before it.

[Github Full Code:](https://github.com/wighawag/hardhat-deploy/blob/v0.11.23/src/index.ts#L162-L164)

```typescript
extendConfig(
  (config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) => {
    ...
    // backward compatibility for runtime (js)
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    if ((userConfig as any).etherscan) {
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      config.verify.etherscan = (userConfig as any).etherscan;
    }
  }
);

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -61,9 +61,11 @@ const config: HardhatUserConfig = {
   mocha: {
     timeout: 10000
   },
-    etherscan: {
+
+  etherscan: {
       apiKey: process.env.ETHERSCAN_API_KEY
   }
+  
 }
 
 // coverage chokes on the "compilers" settings
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -5,7 +5,6 @@ import '@typechain/hardhat'
 import { HardhatUserConfig } from 'hardhat/config'
 import 'hardhat-deploy'
 
-
 import 'solidity-coverage'
 
 import * as fs from 'fs'
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -61,11 +61,8 @@ const config: HardhatUserConfig = {
   mocha: {
     timeout: 10000
   },
-
-  verify: {
     etherscan: {
       apiKey: process.env.ETHERSCAN_API_KEY
-    }
   }
 }
 
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -1,9 +1,10 @@
 
-import 'hardhat-deploy'
+import '@nomiclabs/hardhat-etherscan'
 import '@nomiclabs/hardhat-waffle'
 import '@typechain/hardhat'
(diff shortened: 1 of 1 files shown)


drortirosh:
thank you very much.
Still, I don't really understand where did you come with this error.
The build passes successfully both on local builds, and on server build 
e.g. here is the latest build output on the "develop" branch: 
https://github.com/eth-infinitism/account-abstraction/actions/runs/5378303380/jobs/9757923671

MikeMoldawsky:
It happens because my local dependencies (`yarn install`) differ. In these dependencies, the import order has changed, which caused the `yarn tsc` to break locally, making it harder to onboard ([Similar Ethereum Stackexchange Issue](https://ethereum.stackexchange.com/questions/135618/namedaccounts-does-not-exist-in-type-hardhatuserconfig))

That's why I opened #305, so we can be sure the repo's `yarn.lock` is always up-to-date. 

If you get other issues on the matter, you'll know that that's the reason :) - thanks for the review @drortirosh ! 


> thank you very much. Still, I don't really understand where did you come with this error. The build passes successfully both on local builds, and on server build e.g. here is the latest build output on the "develop" branch: https://github.com/eth-infinitism/account-abstraction/actions/runs/5378303380/jobs/9757923671

PR #308, opened 2023-06-25
Fix EntryPoint revert if postOp reverts with short revert reason

forshtat:


livingrockrises:
changes in linting makes diff view little less readable.

drortirosh (approved):

PR #313, opened 2023-07-07
Add `TestExpiryAccount.sol` implementation with test code

Sangyeup:
### Implementation
1. Register temporary owner(session key)
   - Only owner can register temporary owners
2. Validate UserOp from session key 
   - Only signatures of registered temporary owners will pass
   - Only registered contracts and methods in UserOp will pass
   - Each methods has its own valid time range

---
### **`TestExpiryAccount.sol`**
- **Data Structures**: Two new data structures, **`TargetMethods`** and **`TargetInfo`**, have been introduced:
    - **TargetMethods**: A struct that stores the delegated contract address and an array of delegated function selectors. This gives us a schema for the contracts and functions a session key can interact with.
    - **TargetInfo**: A struct creating mappings to track the contracts, functions, and their valid delegation periods for each temporary owner. It consists of three mappings:
        - **`delegatedContractMap`**: Maps a contract address to a boolean value indicating if the temporary owner has delegation rights.
        - **`delegatedFunctionPeriods`**: Maps a contract address to another mapping that then maps a function selector to a byte string encoding of the valid delegation period (**`_after`** and **`_until`** times).
        - **`delegatedFunctionMap`**: Maps a contract address to another mapping that maps a function selector to a boolean value indicating if the temporary owner has delegation rights to that function.
- **_validateSignature Function**:
    - Overrides the **`_validateSignature`** function from **`BaseAccount`**.
    - Checks whether the signer of a **`UserOperation`** is the original contract owner.
    - If not, it calls **`_validateSessionKey`** to check if the signer (a session key) has appropriate permissions for the requested operation.
- **_validateSessionKey Function**:
    - Checks whether a session key (temporary owner) is authorized to execute a **`UserOperation`**
    - Checks if the target contract and function selector are among those authorized for the session key.
    - Checks if the current time is within the allowed range for the session key.
- **Helper Functions**:
    - **`getSelector`**: Extracts the function selector from the call data.
    - **`_decodeSingle`**: Decodes the call data for a single function call.
    - **`_decodeBatch`**: Decodes the call data for a batch function call.
    - **`_decode`**: Extracts the valid delegation period (**`_after`** and **`_until`** times) from the byte string.

---
### For Security
1. Only the account owner can add tempory owners
2. Account owner is registered as a temporary owner having maximum time range at the time account is deployed.
3. Not registered contracts and methods will make the time range to [0,0]
4. Temporary owners can call execute() function only through the entrypoint.

--- 
### Test Case

At `entrypoint.test.ts`,
1. validateUserOp time-range with calldata validation
    1. should accept non-expired owner
    2. should not reject expired owner
    3. should accept executeBatch
2. validatePaymasterUserOp with deadline
    1. should accept non-expired paymaster request
    2. should not reject expired paymaster request
3. time-range overlap of paymaster and account should intersect
    1. should use lower "after" value of paymaster
    2. should use lower "after" value of account
    3. should use higher "until" value of paymaster
    4. should use higher "until" value of account
4. handleOps abort
    1. should revert on expired account
    2. should revert on date owner
    3. should revert on unregistered contract
    4. should revert on unregistered method
    5. should revert on deleted account


drortirosh:
It looks that you try to implement a full fledged session keys mechanism. At a general review, it looks ok, but it misses one thing: we wanted the test to be minimal, only to highlight how time-ranges can be used for session keys.
A better place for such session keys mechanism is not a specific account , but as an extension module to modular accounts (erc-6900), where it can be added to various wallets.

Sangyeup:
Thanks for the feedback!
I understand that our full implementation could be more suitable as an extension module to ERC-6900 modular accounts. Is there a way for us to contribute our code to the development of ERC-6900?

Additionally, I wanted to ask if there is any need or interest in a contribution to the existing TestExpiryAccount.sol. We're open to simplifying our code if it fits better with the current project scope.

drortirosh:
As I said, this is a test, and thus it should be minimal to what the test is required.
Your code is a sample, but we don't have a repo for such usage samples.
I suggest to follow the ERC-6900 progress, and when they start work on some registry for modules it might be a place for such contribution.

PR #315, opened 2023-07-12
Refactor _payPrefund with Yul

magnetto90:
This modification save some gas and is equivalent to the function made with Solidity.

Test made with forge

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import "forge-std/Test.sol";

contract PayPrefund {
    function payPrefundOriginal(uint256 missingAccountFunds) external {
        if (missingAccountFunds != 0) {
            (bool success,) = payable(msg.sender).call{value: missingAccountFunds, gas: type(uint256).max}("");
            (success);
            //ignore failure (its EntryPoint's job to verify, not account.)
        }
    }

    function payPrefundYul(uint256 missingAccountFunds) external {
        if (missingAccountFunds != 0) {
            assembly {
                pop(call(sub(0x00, 0x01), caller(), missingAccountFunds, 0, 0, 0, 0))
            }
            //ignore failure (its EntryPoint's job to verify, not account.)
        }
    }
}

contract PayPrefundTest is Test {
    PayPrefund public payPrefund;
    address public caller = makeAddr("caller");

    function setUp() public {
        payPrefund = new PayPrefund();
        vm.deal(address(payPrefund), 500);
        vm.startPrank(caller);
    }

    function testOriginal() public {
        payPrefund.payPrefundOriginal(100);
    }

    function testYul() public {
        payPrefund.payPrefundYul(100);
    }
}
```

image

### DIFF
--- a/contracts/core/BaseAccount.sol
+++ b/contracts/core/BaseAccount.sol
@@ -18,7 +18,7 @@ abstract contract BaseAccount is IAccount {
 
     //return value in case of signature failure, with no time-range.
     // equivalent to _packValidationData(true,0,0);
-    uint256 internal constant SIG_VALIDATION_FAILED = 1;
+    uint256 constant internal SIG_VALIDATION_FAILED = 1;
 
     /**
      * Return the account nonce.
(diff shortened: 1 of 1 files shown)


drortirosh:
It is great to optimize the SimpleAccount
Still, simpleAccount comes to be simple, to read and understand, and less being gas-optimized of feature-rich.

I think a must more important gas optimization is paying more than the requiredPrefund: if the account pays extra, then these funds are available in the deposit for future transactions: that is, completely skip the call to deposit in a future transactions.
(an account can always call `entryPoint.withdrawTo()` to withdraw this deposit. it is not locked in any way)

PR #321, opened 2023-08-01
AA-173: Remove simulation and view-only methods from the EntryPoint contract

forshtat:


.github/workflows/build.yml
```diff
@@ -31,6 +31,25 @@ jobs:
 
       - run: yarn run ci
 
+  test-anvil:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: foundry-rs/foundry-toolchain@v1
```
drortirosh: personally, I don't like to rely on github actions: I prefer to be able to run locally the same build as much as possible.
seems like you just need to run
`docker run --rm -ti -p 8545:8545 ghcr.io/foundry-rs/foundry anvil`

test/anvil/entrypointsimulations.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    entryPointSimulations = EntryPointSimulations__factory.createInterface()
+
+    accountOwner = createAccountOwner();
+    ({
+      proxy: account,
+      accountFactory: simpleAccountFactory
+    } = await createAccount(ethersSigner, await accountOwner.getAddress(), entryPoint.address))
+  })
+
+  it('should use state diff when running the simulation', async function () {
```
drortirosh: this is actually a "sanity check" that the tests are valid. so maybe call it from the "before", not as a separate test.

test/anvil/entrypointsimulations.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      const initCode = getAccountInitCode(accountOwner1.address, simpleAccountFactory)
+      const sender = await entryPoint.callStatic.getSenderAddress(initCode).catch(e => e.errorArgs.sender)
+      const op0 = await fillAndSign({
+        initCode,
+        sender,
+        verificationGasLimit: 5e5,
+        maxFeePerGas: 0
+      }, accountOwner1, entryPoint)
+      // must succeed with enough verification gas.
+      await simulateValidation(op0, entryPoint.address, { gas: '0xF4240' })
```
drortirosh: why hex?

test/anvil/entrypointsimulations.test.ts
```diff
@@ -0,0 +1,252 @@
...
+          account.address,
+          account.interface.encodeFunctionData('execute', [sender, 0, '0x'])
+        ]),
+        sender
+      }, accountOwner, entryPoint)
+      const error = await simulateValidation(op1, entryPoint.address).catch(e => e)
+      expect(error.message).to.match(/initCode failed or OOG/, error)
+    })
+
+    // TODO: this test is impossible to do with the "state override" approach
```
drortirosh: since anvil does support debug_traceCall, it is possible to write it better.
but this test is very partial.
should instead say:
for actual opcode and storage rule restrictions, see the reference bundler ValidationManager

test/anvil/entrypointsimulations.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      }, accountOwner1, entryPoint)
+      await fund(op1.sender)
+      await simulateValidation(op1, entryPoint.address, { gas: 10e6 }).catch(e => e)
+      const block = await ethers.provider.getBlock('latest')
+      const hash = block.transactions[0]
+      await checkForBannedOps(hash, false)
+    })
+  })
+
+  // describe('#simulateHandleOp', () => {
```
drortirosh: the one thing missing from simulateHandleOp (which is why we need this "Simulation" contract) is to return the actualGasUsed and success/revert reason of actual call.
they are dumped in an event deep inside innerHandleOp, and the only simple way to propagate them out to the calling "simulateHandleOp" is using storage - a big no-no for a real production code, but good enough for simulate function.

contracts/interfaces/IEntryPoint.sol
```diff
@@ -87,57 +87,17 @@ interface IEntryPoint is IStakeManager, INonceManager {
      *                  so a failure can be attributed to the correct entity.
      */
     error FailedOp(uint256 opIndex, string reason);
+    error FailedOp2(address opIndex, string reason);
```
drortirosh: what is this for?

test/entrypoint.test.ts
```diff
@@ -425,8 +264,10 @@ describe('EntryPoint', function () {
         paymasterAndData: '0x'
       }
       try {
-        await expect(entryPoint.simulateValidation(userOp, { gasLimit: 1e6 }))
-          .to.revertedWith('ValidationResult')
+        const snapshot = await ethers.provider.send('evm_snapshot', [])
+        await entryPoint.simulateValidation(userOp, { gasLimit: 1e6 })
```
drortirosh: use callStatic, and no need for snapshot/revert

test/testutils.ts
```diff
@@ -241,33 +245,20 @@ export async function checkForBannedOps (txHash: string, checkPaymaster: boolean
...
  */
-export function simulationResultWithAggregationCatch (e: any): any {
-  if (e.errorName !== 'ValidationResultWithAggregation') {
-    throw e
-  }
-  return e.errorArgs
-}
-
-export async function deployEntryPoint (provider = ethers.provider): Promise<EntryPoint> {
+export async function deployEntryPointSimulations (provider = ethers.provider): Promise<EntryPointSimulations> {
```
drortirosh: why rename the method? leave the name "deployEntryPoint", and it would avoid a lot of source modifications...

test/entrypoint.test.ts
```diff
@@ -453,8 +294,10 @@ describe('EntryPoint', function () {
         callData: badData.data!
       }
       const beneficiaryAddress = createAddress()
-      await expect(entryPoint.simulateValidation(badOp, { gasLimit: 3e5 }))
-        .to.revertedWith('ValidationResult')
+      const snapshot = await ethers.provider.send('evm_snapshot', [])
+      await entryPoint.simulateValidation(badOp, { gasLimit: 3e5 })
```
drortirosh: use callStatic, and no need for snapshot/revert

test/anvil/entrypointsimulations.test.ts
forshtat: The third parameter of `simulateValidation` is a transaction details object and is fed directly to the `eth_call` RPC method as `{...tx}`. It does not pass through the Ethers.js where `gasLimit?: string | number | BigNumber;` is translated into a hex `gas` field.

.github/workflows/build.yml
forshtat: This will probably not work that well as docker is big and slow. You are free to try though.

contracts/interfaces/IEntryPoint.sol
forshtat: Removed.

test/anvil/entrypointsimulations.test.ts
forshtat: Added comment.

test/anvil/entrypointsimulations.test.ts
forshtat: Moved the sanity check to the "before".

drortirosh:
also, to fix the code-coverage, check the COVERAGE env. variable to skip the failed test

forshtat:
> also, to fix the code-coverage, check the COVERAGE env. variable to skip the failed test

I wanted to see why does it fail in the first place. There seems to be something wrong with the test, but I don't understand what.

.github/workflows/build.yml
forshtat: Won't run Anvil for now.

contracts/core/EntryPointSimulations.sol
```diff
@@ -0,0 +1,165 @@
...
+            data.validUntil,
+            getMemoryBytesFromOffset(outOpInfo.contextOffset)
+        );
+
+        if (aggregator != address(0) && aggregator != address(1)) {
+            AggregatorStakeInfo memory aggregatorInfo = AggregatorStakeInfo(
+                aggregator,
+                _getStakeInfo(aggregator)
+            );
+            return ValidationResult(
```
drortirosh: better "merge" the 2 separate struct returns.

contracts/core/EntryPointSimulations.sol
```diff
@@ -0,0 +1,165 @@
...
+            data.validAfter,
+            data.validUntil,
+            targetSuccess,
+            targetResult
+        );
+    }
+
+    function _simulationOnlyValidations(
+        UserOperation calldata userOp
+    ) private view {
```
drortirosh: internal

test/entrypoint.test.ts
forshtat: Fixed.

test/testutils.ts
forshtat: Fixed.

test/entrypoint.test.ts
forshtat: Fixed.

contracts/core/EntryPointSimulations.sol
forshtat: Fixed.

contracts/core/EntryPointSimulations.sol
forshtat: Changed.

drortirosh:
Comments by @yoavw :

A couple of questions:
1. I don't see any gas measurements in the EntryPointSimulations so I'm not sure I understand how it works.  I understand it's now an eth_call rather than traceCall, so how are the different gas parameters passed to the caller?
2. Did everyone review it and look for redpill attacks?  Let's all make sure there's no way to detect simulation due to this change.
3. In simulateHandleOp there are numberMarker() calls.  I thought we removed tracing for this, and just use eth_call.  Are these markers still used for something?
4. Part of our censorship resistance is that if bundlers are censoring a user, the user can always submit a single-op bundle directly.  Does this method make it possible for wallets to estimate gas without a running a full bundler?  Do public nodes allow the required state injection for this simulation method?
5. We to doc the new method in the EIP, and maybe also a separate doc about simulation to communicate the change to all bundler devs.

drortirosh:
1. gas measurements: this PR is a first step: it moves the view functions into a separate contract, and remove the need for "Revert" structure, but otherwise, doesn't change their calculation. a separate PR should change simulateHandleOp so that it could return out the actualGasUsed and success (which are currently only emitted in a log)
2. @shahafn - please also review the code.
3. numberMarkers will go away, in a next PR (as explained above)
4. stateOverride is a parameter of eth_call, and do not require tracing (so in a way, this change actually makes it easier for wallets to estimate gas, since for actual callGasLimit, calculation they currently require traceCall - at least for estimating first UserOp, since you can "hack" if the account is already deployed
5. right. need to update the EIP to accomodate with this change. @forshtat  ?

shahafn:
There is one way that is a potential exploit:

The new approach means that there are two different EntryPoint contracts, one for simulation and one for execution. 
This means that calls to the very same function (e.g. depositTo())  may have slightly different gas costs (by a diff of ~30 gas).
This depends on the code diff between EntryPoint and EntryPointSimulations, and it may also change between each modification we make to either of them.

In the case where EntryPointSimulations.depositTo() costs 30 gas less than EntryPoint.depositTo(), for instance, an account can call depositTo() during validation 20 times, amplifying the diff to 600 gas, then emit an event. This will pass in simulation since it's cheaper, but fail on actual EntryPoint.

A possible mitigation is to allow only one call to depositTo() during validation, but I think that with gas calculation precise enough, even 30 gas diff can be abused (e.g. through emitting events). 

Wdyt?

yoavw:
@shahafn , @drortirosh and I discussed this a couple of days ago and I believe the conclusion was that every function in EntryPointSimulations must take at least the same amount of gas as its EntryPoint implementation.  If it's marginal, add an extra SSTORE and deduct this fixed cost from the reported gas cost.  This way there are no cases where something results in OOG on chain after passing simulation successfully.  We're ok if OOG happens during simulation - we just increase the gas limit.  What we're trying to prevent is on-chain OOG.

contracts/core/EntryPointSimulations.sol
```diff
@@ -0,0 +1,168 @@
...
+/* solhint-disable no-inline-assembly */
+
+import "./EntryPoint.sol";
+import "../interfaces/IEntryPointSimulations.sol";
+
+/*
+ * This contract inherits the EntryPoint and extends it with the view-only methods that are executed by
+ * the bundler in order to check UserOperation validity and estimate its gas consumption.
+ */
+contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
```
drortirosh, line 14: as per @yoavw 's suggestion, 
please add reverting constructor, with explanation that this code should only be used with "stateOverride"

shahafn (approved):

PR #323, opened 2023-08-04
Rename getAddress -> getAccountAddress

voltrevo:
In ethers v6, you can no longer get the address of a contract synchronously, so they've added an async `.getAddress()` method.

This means using `getAddress` as a contract method breaks the ethers API. `.getAddress` continues to return the address of the factory, and the contract method for getting the account address is inaccessible. The types generated by typechain also incorrectly show the contract method, but the JS actually resolves to ethers' `.getAddress`.

Incorrect types show custom `getAddress` defined in solidity:

Screen Shot 2023-08-04 at 3 07 19 pm

Actually resolves to the ethers' builtin to get the address of the contract, regardless of the parameters passed:

Screen Shot 2023-08-04 at 3 09 08 pm

This PR renames `getAddress` to `getAccountAddress` to fix this problem.

Passes `yarn hardhat test`.

### DIFF
--- a/contracts/samples/SimpleAccountFactory.sol
+++ b/contracts/samples/SimpleAccountFactory.sol
@@ -26,7 +26,7 @@ contract SimpleAccountFactory {
      * This method returns an existing account address so that entryPoint.getSenderAddress() would work even after account creation
      */
     function createAccount(address owner,uint256 salt) public returns (SimpleAccount ret) {
-        address addr = getAddress(owner, salt);
+        address addr = getAccountAddress(owner, salt);
         uint codeSize = addr.code.length;
         if (codeSize > 0) {
             return SimpleAccount(payable(addr));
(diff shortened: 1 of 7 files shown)


drortirosh:
yes, we could do that, since these are basically samples, not "core" contracts.
Still, I think you could do `contract.functions.getAccount(...)` in the application. can't you?

voltrevo:
> yes, we could do that, since these are basically samples, not "core" contracts.

Cool. Any blockers here?

> Still, I think you could do `contract.functions.getAccount(...)` in the application. can't you?

Not in v6 it turns out:

Screen Shot 2023-08-08 at 10 07 43 am

(I checked that `.functions` is a thing in v5. Must have been removed.)

I was able to use `simpleAccountFactory.createAccount.staticCall(...)` as a workaround though.

Doesn't solve the intellisense/typescript issue, but that also has a workaround. No hard blockers here, just clunky.

shahafn:
If ethers v6 has the potential to shadow a contract method, without exposing any way to access it, then it sounds like a bug in ethers that should be fixed there, no?
Did you open an issue for that over there as well?

voltrevo:
@shahafn Good point. I didn't open an issue over there but I've now found this one:
https://github.com/ethers-io/ethers.js/issues/4165

Turns out there is another workaround:
```ts
simpleAccountFactory.getFunction('getAccount')(...)
```

Commenters over there still see it as an issue for ethers though. I believe it's also worth fixing here to improve DX.

voltrevo:
Hi @drortirosh. Did you feel this was not worth solving? Why close this?

PR #324, opened 2023-08-14
AA-183: Reformatting and comments in a separate commit

forshtat:


forshtat:
@drortirosh can't get rid of this newline in prettier. They are not there in actual code.

```
$: diff output.txt output.develop 
473,474c473
<         {
<         } catch Error(string memory revertReason) {
---
>         {} catch Error(string memory revertReason) {
708,709c707
<                         {
<                         } catch Error(string memory reason) {
---
>                         {} catch Error(string memory reason) {

```

drortirosh (approved):

PR #328, opened 2023-08-17
AA-194  verify that EntryPoint calls never cost less during simulation

drortirosh:


contracts/core/EntryPointSimulations.sol
```diff
@@ -159,4 +159,10 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
         // always revert
         revert("");
     }
+
+    //make sure depositTo cost is more than normal EntryPoint's cost.
+    // empiric test showed that without this wrapper, simulation depositTo costs less..
```
yoavw, line 164: Any idea why it costs less? It literally calls the same function that gets called on-chain in EntryPoint.  If the way EntryPointSimulations calls it is cheaper, we should do the same in EntryPoint :)

yoavw (commented):
LGTM, just had one question.
And we need a similar test for the entire validation simulation, to ensure that it never costs less gas than real validation.

contracts/core/EntryPointSimulations.sol
drortirosh: probably some optimization of the "VMT".. because we added more functions, it did some different "grouping" which caused this method to be little more efficient...

forshtat (approved):

PR #330, opened 2023-08-17
AA-190: restrict access to EntryPoint address

forshtat:


eip/EIPS/eip-4337.md
```diff
@@ -308,7 +308,7 @@ While simulating `userOp` validation, the client should make sure that:
    3. destination address must have code (EXTCODESIZE>0)
    4. cannot call EntryPoint's methods, except `depositFor` (to avoid recursion)
```
drortirosh: 1. the method is `depositTo`, not depositFor
2. allowed to access the fallback (empty calldata), which calls depositTo(msg.sender)

eip/EIPS/eip-4337.md
```diff
@@ -308,7 +308,7 @@ While simulating `userOp` validation, the client should make sure that:
...
 5. `EXTCODEHASH` of every address accessed (by any opcode) does not change between first and second simulations of the op.
-6. `EXTCODEHASH`, `EXTCODELENGTH`, `EXTCODECOPY` may not access address with no code.
+6. `EXTCODEHASH`, `EXTCODELENGTH`, `EXTCODECOPY` may not access the EntryPoint contract address or an address with no deployed code.
```
drortirosh: an exception is that checking the existence of an existing contract is allowed.
That is, the sequence `EXTCODESIZE ISZERO` (which is equivalent to `if (address.code.length==0)`) is always allowed, provided that the result is FALSE (this rule is required to allow a call to `entryPoint.depositTo`)

shahafn: And it's technically `EXTCODESIZE`, not `EXTCODELENGTH`, should be fixed in the eip.

drortirosh (approved):

PR #331, opened 2023-08-20
AA-176: Add ERC-165 "supportsInterface" to the EntryPoint

forshtat:
Note: due to 'StakeManager', 'NonceManager' being 'abstract' and only parts of the EntryPoint, and due to IEntryPoint inheriting from all interfaces, the implementation looks a little weird.

I think this this is fine, though.


contracts/interfaces/IEntryPoint.sol
```diff
@@ -9,12 +9,15 @@ pragma solidity ^0.8.12;
...
+// we also require '@gnosis.pm/safe-contracts' and both libraries have 'IERC165.sol', leading to conflicts
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol" as OpenZeppelin;
+
 import "./UserOperation.sol";
 import "./IStakeManager.sol";
 import "./IAggregator.sol";
 import "./INonceManager.sol";
 
-interface IEntryPoint is IStakeManager, INonceManager {
+interface IEntryPoint is IStakeManager, INonceManager, OpenZeppelin.IERC165 {
```
drortirosh: why should the IEntryPoint inherit from ERC165 ? (you later manually remove it, so maybe it should not be in the interface itself...

reports/gas-checker.txt
```diff
@@ -4,36 +4,36 @@
...
+║ gas estimate "big tx 5k" │ 125248 ║
 ╚══════════════════════════╧════════╝
 
 ╔════════════════════════════════╤═══════╤═══════════════╤════════════════╤═════════════════════╗
 ║ handleOps description          │ count │ total gasUsed │ per UserOp gas │ per UserOp overhead ║
 ║                                │       │               │     (delta for │        (compared to ║
 ║                                │       │               │    one UserOp) │     account.exec()) ║
 ╟────────────────────────────────┼───────┼───────────────┼────────────────┼─────────────────────╢
-║ simple                         │     1 │         81899 │                │                     ║
+║ simple                         │     1 │         81943 │                │                     ║
```
drortirosh, line 15: cost 40 per op? 250 per 10 ?

test/entrypoint.test.ts
```diff
@@ -1362,4 +1367,42 @@ describe('EntryPoint', function () {
       })
     })
   })
+
+  describe('ERC-165', function () {
+    it('should return true for EntryPoint interface IDs', async function () {
+      const epInterface = IEntryPoint__factory.createInterface()
```
drortirosh: I miss a sample for: how a bundler can check the current entryPoint adddress it has is OK ?

test/entrypoint.test.ts
```diff
@@ -1362,4 +1367,42 @@ describe('EntryPoint', function () {
...
+          'simulateHandleOp'
+        ].includes(it.name))
+      ])
+      const epInterfaceID = getERC165InterfaceID([...epInterface.fragments])
+      const smInterfaceID = getERC165InterfaceID([...smInterface.fragments])
+      const nmInterfaceID = getERC165InterfaceID([...nmInterface.fragments])
+      const res1 = await entryPoint.supportsInterface(epInterfaceID)
+      const res2 = await entryPoint.supportsInterface(smInterfaceID)
+      const res3 = await entryPoint.supportsInterface(nmInterfaceID)
+      const res4 = await entryPoint.supportsInterface(epPureInterfaceID)
```
drortirosh: better:

expect(await entryPoint.supoprtsInterface). to.be.true

contracts/interfaces/IEntryPoint.sol
drortirosh: actually, maybe its good, so a bundler can work only with IEntryPoint, not with EntryPoint contract itself.

contracts/core/EntryPoint.sol
```diff
@@ -39,6 +39,15 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard
      */
     uint256 public constant SIG_VALIDATION_FAILED = 1;
 
+    /// @inheritdoc OpenZeppelin.IERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override(OpenZeppelin.ERC165, OpenZeppelin.IERC165) returns (bool) {
+        return interfaceId == (type(IEntryPoint).interfaceId ^ type(IStakeManager).interfaceId ^ type(INonceManager).interfaceId) ||
```
drortirosh, line 48: 1. add a comment here that solidity ignores the "inheritance" of interfaces.
2. need to add the IERC165 Id itself (since it IS part of the IEntryPoint as we define it

src/Utils.ts
```diff
@@ -0,0 +1,17 @@
...
+export function getERC165InterfaceID (abi: JsonFragment[]): string {
+  let interfaceId =
+    abi
+      .filter(it => it.type === 'function' && it.name != null)
+      .map(it => {
+        const iface = new Interface([it])
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return iface.getSighash(it.name!)
+      })
+      .filter(it => it !== '0x01ffc9a7') // remove the IERC165 method itself
```
drortirosh: why do you exclude it? see comment in the contract.

drortirosh (approved):

PR #333, opened 2023-08-21
AA-164 better estimate gas

drortirosh:
new implementation for simulateHandleOp, to callGasLimit


contracts/core/EntryPoint.sol
```diff
@@ -680,30 +675,62 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
             );
         } // unchecked
     }
 
-    /**
-     * The gas price this UserOp agrees to pay.
-     * Relayer/block builder might submit the TX with higher priorityFee, but the user should not.
-     * @param mUserOp - The userOp to get the gas price from.
+    //solhint-disable-next-line no-empty-blocks
+    function _postOpGas(uint256 postOpGas) internal virtual {
```
forshtat, line 698: I am wondering, is it really the only time we may want to expose an internal variable to the simulation code? How come?

contracts/core/EntryPointSimulations.sol
```diff
@@ -112,16 +116,51 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
         if (target != address(0)) {
             (targetSuccess, targetResult) = target.call(targetCallData);
         }
-        return ExecutionResult(
-            opInfo.preOpGas,
-            paid,
+        uint userOpTotalValidationGasUsed = opInfo.preOpGas - op.preVerificationGas;
+        //userOpActualGasUsed, userOpSuccess are saved by _emitUserOperationEvent, below..
```
forshtat, line 120: Maybe `userOpActualGasUsed` and `userOpSuccess` should not be different from `userOpPostOpGas`?
Make a call to an empty callback in the deployed code. Instead of creating two distinct `_emitUserOperationEvent` functions, for simulation and execution?

contracts/interfaces/IEntryPointSimulations.sol
```diff
@@ -5,12 +5,31 @@ import "./UserOperation.sol";
 import "./IEntryPoint.sol";
 
 interface IEntryPointSimulations is IEntryPoint {
+
     // Return value of simulateHandleOp.
```
forshtat: Please join the two comments into a single one.

test/entrypoint.test.ts
```diff
@@ -74,7 +74,7 @@ describe('EntryPoint', function () {
 
     const chainId = await ethers.provider.getNetwork().then(net => net.chainId)
 
-    entryPoint = await deployEntryPoint()
+    entryPoint = await deployEntryPoint() as any as EntryPoint
```
forshtat: Why? The `deployEntryPoint` seems to return `EntryPoint` anyways.

forshtat (changes requested):

contracts/core/EntryPoint.sol
drortirosh: We want to expose anything that is gas-related

contracts/core/EntryPointSimulations.sol
drortirosh: the _emit method will stay, since it so happens that emitting from a method is slightly more gas-efficient than inlining it...
Do you think we should pass those parameters twice, once to the _emit method and another to a debug-only method?

contracts/core/EntryPoint.sol
shahafn: Maybe call it `simulationHookPostOpGas()` or something similar, as this is what it is? Plus, maybe we want more hooks like this one in other places for simulation?

drortirosh:
stale. should be re-implemented.

PR #342, opened 2023-09-13
AA-170: validation rules as a separate ERC.

drortirosh:


eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+considered an attack. The node is expected to apply web2 security measures and throttle requests based on API key,
+source IP address, etc.
+RPC nodes already do that to prevent being spammed with invalid transactions which also have a validation cost.
+
+### Constants:
+
+
+| Title                                | Value                       | Comment                                                                         |
+|--------------------------------------|-----------------------------|---------------------------------------------------------------------------------|
+| `MIN_UNSTAKE_DELAY`                  | 84600                       | 1 day                                                                           |
```
kristofgazso: 84600 -> 86400

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+   that is performed during the bundle creation.
+2. Submit `UserOperations` that are valid in isolation during validation, but when bundled together become invalid.
+3. Submit valid `UserOperations` but "front-run" them by executing a state change on the
+   network that causes them to become invalid. The "front-run" in question must be economically viable.
+
+To prevent such attacks, we attempt to "sandbox" the validation code.
+We isolate the validation code from other `UserOperations`, from external changes to the storage, and
+from information about environment such as a current block timestamp.
+
+### What is not considered a `mass invalidation attacl`
```
kristofgazso: attacl -> attack

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Definition of the `mass invalidation attack`
+
+A possible set of actions is considered to be a `mass invalidation attack` on the network if a large number of
+`UserOperations` that did pass the initial validation and were accepted by nodes and propagated further into the
+mempool to all bundlers in the network, becomes invalid and not eligible for inclusion in a block.
+
+There are 3 ways to perform such an attack:
+
+1. Create a code that passes the initial validation, but later fails the re-validation
```
kristofgazso: code -> UserOperation

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+   that is performed during the bundle creation.
+2. Submit `UserOperations` that are valid in isolation during validation, but when bundled together become invalid.
+3. Submit valid `UserOperations` but "front-run" them by executing a state change on the
+   network that causes them to become invalid. The "front-run" in question must be economically viable.
+
+To prevent such attacks, we attempt to "sandbox" the validation code.
+We isolate the validation code from other `UserOperations`, from external changes to the storage, and
+from information about environment such as a current block timestamp.
```
kristofgazso: about _the_ environment

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Abstract
+
+This document describes the validation rules we impose on the validation context of Account Abstraction transactions,
+such as [ERC-4337](./eip-4337) `UserOperation` or [EIP-9999](./eip-9999), which are enforced off-chain by a
+block builder or a standalone bundler, and the rationale behind each one of them.
+
+## Motivation
+
+All transactions initiated by EOAs initiated have an implicit validation phase where balance, nonce, and signature are
```
ch4r10t33r: The second "initiated" is redundant.

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+checked to be valid for the current state of the Ethereum blockchain.
+Once the transaction is checked to be valid by a node, only another transaction by the same EOA can modify the Ethereum
```
ch4r10t33r: only _then_ another

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+* Mainnet- (12-second blocks) need 12000 UserOps
+    * The initial cost (deploy+initial state change) is 12000*60000 = 720M gas
+    * Sustained cost (only storage modification) - 12000*5000 =  70Mgas
+* Total attack cost
+    * Polygon (@150 gwei) - $18 + $1.5 per block ($2700/hour)
+    * Mainnet: (@40 gwei) - $51840 + $5040 per block ($1.5M/hour)
+    * Note that the above costs ignore the gas-price increase because the underlying network is also overloaded.
+* Mitigations
+    * A **staked factory:** 
+        * Can attempt such an attack once, since after invalidating a single bundle creation its reputation drops and another staked bundler has to be used.
```
ch4r10t33r: Can we secure this further by allowing reputation information also to be shared via p2p interface of the bundlers? @drortirosh @yoavw

ch4r10t33r: Shouldn't this read "another staked _paymaster_ has to be used."?

shahafn: @kristofgazso This is problematic, reputation was always meant to be local to the bundler, otherwise bundlers could not only censor an account/paymaster/factory/aggregator, they could also direct other bundlers to censor them. Sharing reputation as part of the p2p core protocol opens a can of worms in that regard.

eip/EIPS/eip-4337.md
```diff
@@ -462,24 +430,10 @@ it is critical that the exact same opcode and precompile banning rules as well a
 for the `handleOps` validation in its entirety as for individual UserOperations.
 Otherwise, attackers may be able to use the banned opcodes to detect running on-chain and trigger a `FailedOp` revert.
 
-Banning an offending entity for a given bundler is achieved by increasing its `opsSeen` value by `1000000`
-and removing all UserOperations for this entity already present in the mempool.
-This change will allow the negative reputation value to deteriorate over time consistent with other banning reasons.
-
-If any of the three conditions is violated, the client should reject the `op`. If both calls succeed (or, if `op.paymaster == ZERO_ADDRESS` and the first call succeeds)without violating the three conditions, the client should accept the op. On a bundler node, the storage keys accessed by both calls must be saved as the `accessList` of the `UserOperation`
+The storage slots accessed during the above tracing may be used as "accessList" when submitting the transaction.
```
dancoombs: Couple things about this section:

The above is suggesting to use a `debug_traceCall` on the entire `handleOps` transaction. Shouldn't a bundler only need to care about the validation section of `handleOps`? Tracing the entire thing seems like a waste of resources.

> use the consumed gas for the actual transaction execution. 

This seems like an easy attack on a bundler: have something use less gas during simulation than during onchain execution. To me, the only safe way to do this is to use the gas limit of each UO (plus a buffer) as the total gas limit of the. 

I understand that we eventually want bundlers to be block builder such that they know the exact onchain state when they run the simulation. However I'm not sure we should be baking that into the procedures suggested here.

> The storage slots accessed during the above tracing may be used as "accessList" when submitting the transaction.

Similarly to above, the slots accessed during execution can change between bundle simulation and execution. It seems only safe to use slots accessed during validation in the access list. Again, that changes if you know the exact state and have guaranteed top of block inclusion, but I don't see why we should be baking this into the spec.

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+state in a way that makes the first transaction invalid.
+
+With Account Abstraction, however, the validation can also include an arbitrary EVM code and rely on storage as well,
+which means that unrelated `UserOperations` or transactions may invalidate each other.
+
+If not addressed, this would make the job of maintaining a mempool of valid `UserOperations` and producing valid
+bundles computationally infeasible and susceptible to DoS attacks.
+
+This document describes set of validation rules that, if applied by a block builder prior including a `UserOperation`,
+can prevent such attacks.
```
dancoombs: > if applied by a block builder prior including

Should this be: "if applied by a block builder prior to including a `UserOperation` in the mempool"

dancoombs: Are we committing to using the term "block builder" over "bundler" here?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Constants:
+
+
+| Title                                | Value                       | Comment                                                                         |
+|--------------------------------------|-----------------------------|---------------------------------------------------------------------------------|
+| `MIN_UNSTAKE_DELAY`                  | 84600                       | 1 day                                                                           |
+| `MIN_STAKE_VALUE`                    | Adjustable per chain value  | Equivalent to ~$1000 in native tokens                                           |
+| `SAME_SENDER_MEMPOOL_COUNT`          | 4                           |                                                                                 |
+| `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` | 10                          |                                                                                 |
```
dancoombs, line 71: What is the rational for `10` here?

That value seems low enough that any highly used paymaster or factory will need to stake. Is that the intention?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Validation Rules
+
+### **Definitions**:
+1. **Validation Phases**: there are up to 3 phases of validation
+   1. smart account deployment
+   2. smart account validation 
+   3. paymaster validation.
+2. **Entity**: a contract that is explicitly used by the `UserOperation`.
+   Includes the `deployer`, `paymaster`, `aggregator` and staked `account`, as discussed below. \
```
dancoombs: `deployer` -> `factory`?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+    * **[OP-062]** Precompiles:
+        * Only allowed the core 9 precompiles.\
+          Specifically the computation precompiles that do not access anything in the blockchain state or environment.
+
+### Code Rules
+
+* **[COD-010]** Between the first and the second validations, the `EXTCODEHASH` value of any visited address,
+entity or referenced library, may not be changed.\
+If the code is modified, that entity is throttled down. \
+If the code of a library was modified, then the entity that used that library is throttled.
```
dancoombs: Is this an explicit throttling? Or are the user ops just dropped? What does it mean to "throttle down"?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+* **[COD-020]**: Code Modification Throttling: \
+In order to prevent "mass invalidation" by a self-modifying code, if the contract code is modified,
+it is throttled down and gains reputation as long as it does not change.
+    * Maintain a list of `{address, codehash, last modified}`
```
dancoombs: What is `last modified` used for?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+      `*CALL` during validation -  not just the code that was actually visited during validation.
+
+### Storage rules.
+
+The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase are limited as follows:
+
+* **[STO-010]** Access to the "account" storage is always allowed.
+* Access to associated storage of the account in an external (non-entity contract) is allowed if:
+    * **[STO-021]**  The account already exists
+    * **[STO-022]**  There is an `initCode` and the `factory` contract is staked.
```
dancoombs, line 210: Might help to clarify that this is an OR condition

eip/EIPS/eip-aa-rules.md
drortirosh: obsolete... will be removed.

eip/EIPS/eip-aa-rules.md
drortirosh: (bundler is wrong here, but it is actually another staked factory)

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+    * Any entity code except the account itself  (`paymaster`, `factory`, and any referenced library) may include it only if that entity is staked.
+    * This rule applies to the entire executed code section - including external contract code called by
+      `*CALL` during validation -  not just the code that was actually visited during validation.
+
+### Storage rules.
+
+The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase are limited as follows:
+
+* **[STO-010]** Access to the "account" storage is always allowed.
+* Access to associated storage of the account in an external (non-entity contract) is allowed if:
```
dancoombs: Does this mean that accessing associated storage on an entity contract is never allowed? For example if a sender wanted to access associated storage on a paymaster contract. Or is this always allowed?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+    * **[STO-021]**  The account already exists
+    * **[STO-022]**  There is an `initCode` and the `factory` contract is staked.
+* If the entity (`paymaster`, `factory`) is staked, then it is also allowed:
+    * **[STO-031]** Access the entity's own storage.
+    * **[STO-032]** Access a storage slot in any non-entity contract that is associated with the entity.
+* **[STO-040]** `UserOperation` may not use an entity address (`factory`/`paymaster`/`aggregator`) that is used as an "account" in another `UserOperation` in the mempool. \
+This means that a `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
+* **[STO-041]** A contract whose Associated Storage slot is accessed during a `UserOperation` may not be an address of a "sender" in another `UserOperation` in the mempool.
```
dancoombs, line 216: `STO-040` and `STO-041` seem like they are going to be very difficult to enforce in the P2P network since each bundler will have a different view of the mempool.

Its also dependent on the ordering upon which UOs enter the mempool. Unless each time we add a UO to the pool we do a scan over every operation in the pool for storage accesses.

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+* **[STO-050]** `UserOperations` that are using the same unstaked entity are only allowed up to appear `SAME_SENDER_MEMPOOL_COUNT` times in the mempool.
```
dancoombs: Should the be `SAME_SENDER_MEMPOOL_COUNT` for senders and `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for other entities?

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Entity-specific rules:
+
+* **[EREP-010]** For each `paymaster`, the mempool must maintain the total gas `UserOperations` using this `paymaster` may consume.
+    * Do not add a `UserOperation` to the mempool if the maximum total gas usage, including the new `UserOperation`, is above the deposit of the `paymaster` at the current gas price.
+    * TODO: allow "fractional reserve" based on reputation?
+* **[EREP-020]** A staked factory is "accountable" for account breaking the rules. \
+That is, if the `validateUserOp()` is rejected for any reason, it is treated as if the factory caused this failure, and thus this affects its reputation.
+* **[EREP-030]** A Staked Account is accountable for failures in other entities (`paymaster`, `aggregator`) even if they are staked. \
+(see [https://github.com/eth-infinitism/account-abstraction/issues/274](https://github.com/eth-infinitism/account-abstraction/issues/274))
```
dancoombs, line 240: Does this mean a staked paymaster is not accountable for any other entity breaking the rules? Only the paymaster itself?

eip/EIPS/eip-4337.md
drortirosh: 1. yes, the bundler only cares only the validation part of the handleOps. it can either "stop tracing" (using a flag) after the validation phase, or it can use alternate handleOp impl to trace (current ref. bundler uses the former)

drortirosh: 2. I agree that the gas rules should be removed, and replaced with general suggestions: They are not really required by the "consensus", and should be defined by the bundler itself.

drortirosh: 3. re: accessList building - you're right. We indeed specified it as may (which doesn't imply preference), but it can (may?/should?) be removed....

dancoombs: Agreed with the above. I feel like (3) should be removed unless its really pressing.

(will split into individual comments next time, sorry about the numbering)

eip/EIPS/eip-aa-rules.md
drortirosh: reverted to bundler...

eip/EIPS/eip-aa-rules.md
drortirosh: it is small enough to prevent "mass-invalidation" attacks, but the rules below cause the actual value to rise rapidly (as long as UserOps gets included).

eip/EIPS/eip-aa-rules.md
drortirosh: yes (otherwise, a malicious deployer might cause a paymaster reputation drop.)

eip/EIPS/eip-aa-rules.md
drortirosh: yes, though see it was moved to UREP-10 (its "unstaked" rule, not "storage" rule)...

eip/EIPS/eip-aa-rules.md
drortirosh: We try to prevent storage access between entities at all.
"Associated storage" in another entity is allowed by both (one as "self" storage, another as "associated") - and thus it is blocked.

eip/EIPS/eip-aa-rules.md
```diff
@@ -0,0 +1,339 @@
...
+    * TODO: can be removed once we remove the "2nd postOp".
+
+### Unstaked Entities Reputation Rules
+
+* Definitions:
+    * **`opsSeen`, `opsIncluded`, and reputation calculation** are defined above.
+    * `UnstakedReputation` of an entity is a maximum number of entries using this entity allowed in the mempool.
+    * `opsAllowed` is a reputation based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
+* **[UREP-010]** `UserOperation` with unstaked sender are only allowed up to `SAME_SENDER_MEMPOOL_COUNT` times in the mempool.
+* **[UREP-020]** `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
```
shahafn: Specifically for paymaster/factory/aggregator, not senders.

eip/EIPS/eip-aa-rules.md
drortirosh: These rules come more to protect the bundler itself, and not the entire mempool.
That is, you may ignore those rules (making you easier to be DoS - attacked). 
Also, you're not expected to mark a sender of such userops as a "spammer"
We should add better definition of these type of rules (consensus rules and bundler-specific rules)

forshtat (approved):

PR #349, opened 2023-10-03
AA-214 Extracting UserOperationLib from UserOpertaion.sol and moving to core/

shahafn:


forshtat (approved):

contracts/core/UserOperationLib.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: GPL-3.0
```
drortirosh, line 1: we do have "util" folder. doesn't it fit there?

shahafn: I think that this library should be part of the core, no?

drortirosh: it is part of core (e.g. entrypoint imports "exec")
but you're right in that it is for "non-ep-specific" data, and UserOp (and its lib) are ep-specific.

PR #350, opened 2023-10-03
AA-215 Bls aggregator replay fix

shahafn:


forshtat (approved):

contracts/core/UserOperationLib.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: GPL-3.0
```
drortirosh, line 1: please don't refactor UseROperation as part of "bls aggregator fix", but in a separate PR.

contracts/samples/bls/BLSSignatureAggregator.sol
```diff
@@ -104,14 +111,13 @@ contract BLSSignatureAggregator is IAggregator {
...
 
-    // helper for test
     function getUserOpHash(UserOperation memory userOp) public view returns (bytes32) {
         bytes32 publicKeyHash = _getPublicKeyHash(getUserOpPublicKey(userOp));
         return _getUserOpHash(userOp, publicKeyHash);
     }
 
     function _getUserOpHash(UserOperation memory userOp, bytes32 publicKeyHash) internal view returns (bytes32) {
-        return keccak256(abi.encode(internalUserOpHash(userOp), publicKeyHash, address(this), block.chainid));
+        return keccak256(abi.encode(internalUserOpHash(userOp), publicKeyHash, address(this), block.chainid, entryPoint));
```
drortirosh, line 120: why hash aggregator at all? doesn't EP replaces it?

shahafn: Is there a reason not to hash it? I'm usually inclined to prevent malleability even if we can't think of a way to attack it right now.

PR #351, opened 2023-10-03
Initial discussion of unused gas onchain solution

shahafn:
Currently, a client can send a `UserOperation` with an arbitrarily high `callGasLlimit` but leaving most of it unused, in order to occupy most of the bundle gas limit. This can potenitally hurt the bundler  profitability, by making it skip other `UserOperations` due to block gas limit.
Instead of each bundler having to implement its own custom protection against such griefing attempt, we suggest the following fix in EntryPoint:

For a user operation exceeding a certain threshold (e.g. 1M gas limit): 
1.  Allow the user to have a cushion of min(50K, 0.1 * userOp.callGasLimit). If it passes that threshold, then:
2.  Charge the user extra 10% of the remaining unused gas above threshold.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,3 +9,4 @@ Implementation of contracts for [ERC-4337](https://eips.ethereum.org/EIPS/eip-43
 [Bundler reference implementation](https://github.com/eth-infinitism/bundler)
 
 [Bundler specification test suite](https://github.com/eth-infinitism/bundler-spec-tests)
+


dancoombs:
> For a user operation exceeding a certain threshold (e.g. 1M gas limit):

To clarify, does this mean that this only applies to UOs with a `callGasLimit > 1M`? What is the rational for that? Why wouldn't this be a policy for all UOs?

dancoombs:
Because this only deals with `callGasLimit`, we're leaving bundlers to deal with verification overestimates themselves.

Either: if during simulation `verificationGasLimit` >> `gas used during simulation` the bundler could reject the UO (this could be a validation rule). 

Or: the bundler just uses gas used during simulation (accounting for 63/64ths rule) as the `verificationGasLimit` that it applies to its own gas limit calculation.

Other options?

yoavw:
> > For a user operation exceeding a certain threshold (e.g. 1M gas limit):
> 
> To clarify, does this mean that this only applies to UOs with a `callGasLimit > 1M`? What is the rational for that? Why wouldn't this be a policy for all UOs?

The rationale is to avoid penalizing users for minor inaccuracies that can happen in small ops. For example, an op that does very little but accesses several slots in a frequently used contract, may benefit from these slots being pre-warmed by another op. The numbers would be fairly small, not cause a real issue for the bundler, and it seems unnecessary to penalize the user for it.  Whereas if an op declared that it needs 5M gas and ended up using 50k, it does cause a problem.  Keeping in mind the problem we're trying to solve, it seemed unjustified to apply this to small ops.

However, there's the issue of defining "large".  I'm not too thrilled about introducing a constant, and the numbers may differ between networks.  But do you see a better way to spare normal transactions while penalizing unreasonable ones (much larger than average, and leaving an unreasonable amount of unused gas)?

yoavw:
> Because this only deals with `callGasLimit`, we're leaving bundlers to deal with verification overestimates themselves.

That's intentional.  Bundlers will already limit validation gas to prevent DoS through long validations that fail.  Therefore validation gas can't be used to effectively attack the bundler and cause it not to build big bundles.  Only callGas is uncapped.

> 
> Either: if during simulation `verificationGasLimit` >> `gas used during simulation` the bundler could reject the UO (this could be a validation rule).

The bundler could do that anyway, but due to the reason above, we don't need to enforce a penalty in EntryPoint for this one.

> Or: the bundler just uses gas used during simulation (accounting for 63/64ths rule) as the `verificationGasLimit` that it applies to its own gas limit calculation.

That seems risky to assume.  An attacker could invalidate entire bundles on chain, and eth_sendRawTransactionConditional can't defend the bundler against it because the validation will fail on OOG rather than a changed slot.  Worse yet, the gas might run out in another op rather than the one from which it was deducted.  If I were implementing a bundler and had to mitigate unused validation gas (which, again, I'm not sure is needed), I'd use the first option and drop these ops.

> Other options?

Yes - ignore validation gas and only solve the problem for callGas.  Do you think it can be used to perform an effective attack?

drortirosh:
why should we separate validation and execution? I think the simplest solution is to compare the total gas (maxPossibleGas, which the account already paid as a refund) with the "actualGasUsed" - the value we calculate it should eventually pay.
The different is the "overpayment" - and for that we want the account to pay 10%
This will also deter (staked) paymasters, deployers from creating different gas usages between validation and execution

drortirosh:
Suggested payment mechanism
- preVG is always paid in full, so there is no penalty
- verificationGasLimit should not penalized (bundlers were supposed to validate it anyway)
- unused callGasLimit is paid  10%

drortirosh:
resolved in #356

PR #353, opened 2023-10-10
eip: remove "increment nonce" clause

drortirosh:
nonce is managed by entrypoint. account doesn't have to increment it.

### DIFF
--- a/eip/EIPS/eip-4337.md
+++ b/eip/EIPS/eip-4337.md
@@ -1018,7 +1018,7 @@ See `https://github.com/eth-infinitism/account-abstraction/tree/main/contracts`
 
 ## Security Considerations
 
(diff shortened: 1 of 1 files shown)


yoavw (commented):
It doesn't have to increment the nonce, but maybe "optionally enforce nonce restrictions" instead?

E.g. require(key==0) for incremental nonces, or require(seq==0) for unordered replay protection.

drortirosh:
> but maybe "optionally enforce nonce restrictions" instead?

there is a whole "Semi-abstracted Nonce Support" section. just removed obsolete mention to "increment" here.

yoavw (approved):
Approved, following our discussion in the previous review.

PR #355, opened 2023-10-11
Allow read only access to global storage.

drortirosh:


eip/EIPS/eip-aa-rules.md
```diff
@@ -210,7 +210,8 @@ The permanent storage access with `SLOAD` and `SSTORE` instructions within each
     * **[STO-022]**  There is an `initCode` and the `factory` contract is staked.
 * If the entity (`paymaster`, `factory`) is staked, then it is also allowed:
     * **[STO-031]** Access the entity's own storage.
-    * **[STO-032]** Access a storage slot in any non-entity contract that is associated with the entity.
+    * **[STO-032]** Write Access to storage slots that is associated with the entity, in any non-entity contract.
```
shahafn: That's true for unstaked entities as well

yoavw: I don't think so. An unstaked entity can only access storage associated with the **account**, not the entity.

leekt: does account has to be staked to read the global storage?

drortirosh: This is for entity, not specifically account (e.g. mostly used by paymasters, deployers)
A paymaster/deployer MAY use this global storage even if the account is not staked.
If it is not clear from the description, we'll have to elaborate.

yoavw: @leekt but if, for some reason, the account is staked - then it'll also be allowed to SLOAD unassociated storage. Normally, accounts are not staked but paymasters and deployers are.

If you intend to use this new rule, it's your code's responsibility to ensure that ops can't get invalidated by reading this global storage.

E.g. it's safe for your paymaster to use a token like USDC that requires global SLOAD (loading its proxy implementation slot and the `paused` flag) without requiring an alt mempool.  But if you also SLOAD the price from a price oracle, then your paymaster runs a risk of getting banned if someone sends many ops from accounts with small USDC balances and then the price goes up so they become invalid.

Global SLOAD gives a lot more power to staked entities, but also more ways to shoot themselves in the foot if they're not careful.

yoavw: "Full access" or "Read/Write access", since the next line says "Read-only" rather than "Read" so it's not exactly complementing.

yoavw (approved):

PR #358, opened 2023-10-26
use deposited token to pay gas

robinsdan:
It appears that the behavior of the code is inconsistent with the comments. And I found the entrypoint will not call paymaster with `postOpReverted` mode.


drortirosh:
DepositPaymaster is obsolete, and should be completely reverted.
The whole reason for having a deposit is to make sure the account can't withdraw its balance between validation and execution, so the deposit was meant to be kept until after the execution is finished.
This logic is no longer required (for quite some time), since a paymaster is allowed to withdraw funds from its validation function - hence this contract is obsolete.

PR #359, opened 2023-10-29
chore: remove erc777

### LINKED ISSUE #354: For newer version of @openzeppelin/contracts@^5.0.0 ERC777 has been removed.
In the most recent release of the @openzeppelin/contracts library, the ERC777-related contracts have been removed. As a consequence,  the IERC777Recipient interface used in the @account-abstraction/contracts/samples/callback/TokenCallbackHandler.sol file, triggers an error.

For additional context and reference, here are the relevant screenshots and links

Openzeppelin change log. 

Screenshot 2023-10-11 121522

Screenshot 2023-10-11 121451

@account-abstraction/contracts/samples/callback/TokenCallbackHandler.sol 
Screenshot 2023-10-11 121617

.

dawksh:
Recently OpenZeppelin upgraded contracts to v.5.0.0 and above which has removed the `ERC777` implementation. Since the AA kit still uses it, it results in conflicts.

ERC777 is an overengineered solution, more about it [here](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620)

This PR fixes #354 and makes the kit supported out of the box without hassle of changing dependencies.

### DIFF
--- a/contracts/samples/callback/TokenCallbackHandler.sol
+++ b/contracts/samples/callback/TokenCallbackHandler.sol
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0
 pragma solidity ^0.8.12;
 
-/* solhint-disable no-empty-blocks */
-
 import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
--- a/contracts/samples/callback/TokenCallbackHandler.sol
+++ b/contracts/samples/callback/TokenCallbackHandler.sol
@@ -4,7 +4,6 @@ pragma solidity ^0.8.12;
 /* solhint-disable no-empty-blocks */
 
 import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
-import "@openzeppelin/contracts/token/ERC777/IERC777Recipient.sol";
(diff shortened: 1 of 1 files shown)


contracts/samples/callback/TokenCallbackHandler.sol
```diff
@@ -4,25 +4,14 @@ pragma solidity ^0.8.12;
 /* solhint-disable no-empty-blocks */
```
ManuLpz4: I think that you can remove this line because is no longer needed. What do you think? :)

contracts/samples/callback/TokenCallbackHandler.sol
```diff
@@ -52,7 +41,9 @@ contract TokenCallbackHandler is IERC777Recipient, IERC721Receiver, IERC1155Rece
         return IERC1155Receiver.onERC1155BatchReceived.selector;
     }
 
-    function supportsInterface(bytes4 interfaceId) external view virtual override returns (bool) {
```
ManuLpz4, line 55: And this seems like a line wrapping commit, maybe because the author was using 120 characters and you 80 (not sure)

contracts/samples/callback/TokenCallbackHandler.sol
dawksh: Makes sense, have pushed a fix

leopoldjoy:
Hello @dawksh @ManuLpz4, would it be possible to get this fix merged? Thank you!

dawksh:
I'm not sure why the checks are failing @ManuLpz4 any ideas?

drortirosh:
done as part of #384

PR #360, opened 2023-10-30
AA-221: Prevent Paymaster 'postOp' function from reverting the entire bundle

forshtat:


test/entrypoint.test.ts
```diff
@@ -1075,6 +1075,26 @@ describe('EntryPoint', function () {
...
+          initCode: getAccountInitCode(account3Owner.address, simpleAccountFactory),
+
+          verificationGasLimit: 3e6,
+          callGasLimit: 1e6
+        }, account3Owner, entryPoint)
+        const beneficiaryAddress = createAddress()
+        await entryPoint.handleOps([op], beneficiaryAddress)
+      })
+
+      it('should not revert when paymaster reverts with known EntryPoint error in postOp', async function () {
```
drortirosh, line 1111: at minimum, should verify that the action was reverted (success=false, read the pm "reverted" data)

contracts/core/EntryPoint.sol
```diff
@@ -662,9 +662,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
                 if (context.length > 0) {
                     actualGasCost = actualGas * gasPrice;
                     if (mode != IPaymaster.PostOpMode.postOpReverted) {
-                        IPaymaster(paymaster).postOp{
+                        try IPaymaster(paymaster).postOp{
                             gas: mUserOp.verificationGasLimit
-                        }(mode, context, actualGasCost);
+                        }(mode, context, actualGasCost)
+                        {} catch (bytes memory reason) {
+                            revert(string(abi.encodePacked("AA96 postOp reverted", reason)));
```
drortirosh: this is problematic: you append binary encoding of custom error to a string.
(try to decode the error from the test..)
at a minimum: change string suffix to "...reverted: ", as we do with other reverts

test/entrypoint.test.ts
```diff
@@ -1090,6 +1091,27 @@ describe('EntryPoint', function () {
...
+          callData: accountExecFromEntryPoint.data,
+          initCode: getAccountInitCode(account3Owner.address, simpleAccountFactory),
+
+          verificationGasLimit: 3e6,
+          callGasLimit: 1e6
+        }, account3Owner, entryPoint)
+        const beneficiaryAddress = createAddress()
+        const rcpt1 = await entryPoint.handleOps([op], beneficiaryAddress).then(async t => await t.wait())
+        const logs1 = await entryPoint.queryFilter(entryPoint.filters.UserOperationEvent(), rcpt1.blockHash)
+        expect(logs1[0].args.success).to.be.false
```
drortirosh, line 1129: also the "postOp reverted" message

test/entrypoint.test.ts
```diff
@@ -1090,6 +1091,27 @@ describe('EntryPoint', function () {
         await entryPoint.handleOps([op], beneficiaryAddress)
```
drortirosh: handleOps doesn't revert, but you should check event side effects here too

test/entrypoint.test.ts
forshtat: > also the "postOp reverted" message

There seems to be no place where `"postOp reverted"` error could be observed. We do not currently emit it as a part of any event, do we?

test/entrypoint.test.ts
forshtat: Added a check for `success == false` here. What other side effects do you want to check here?

test/entrypoint.test.ts
forshtat: This is done.

contracts/core/EntryPoint.sol
forshtat: Changed to revert with custom error that we can parse off-chain if we add the "postOp reverted" event (see next PR https://github.com/eth-infinitism/account-abstraction/pull/362)

test/testutils.ts
```diff
@@ -172,6 +172,13 @@ export function decodeRevertReason (data: string, nullIfNoMatch = true): string
   } else if (methodSig === '0x4e487b71') {
     const [code] = ethers.utils.defaultAbiCoder.decode(['uint256'], dataParams)
     return `Panic(${panicCodes[code] ?? code} + ')`
+  } else if (methodSig === '0x8d6ea8be') {
+    const [reason] = ethers.utils.defaultAbiCoder.decode(['string'], dataParams)
+    return `CustomError("${reason as string}")`
```
drortirosh, line 177: This `testutils` package was not meant to be a place for test-specific methods, but for "system-wide" reverts (yes, due to a lack of such a decoder in ethers itself)

drortirosh (approved):

PR #362, opened 2023-11-01
Emit 'PostOpRevertReason' event

forshtat:


shahafn:
Looks ok, tests fail though

contracts/core/EntryPoint.sol
```diff
@@ -667,7 +674,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
                         }(mode, context, actualGasCost)
                         // solhint-disable-next-line no-empty-blocks
                         {} catch (bytes memory reason) {
```
drortirosh, line 676: `catch(bytes memory reason)` forces the compiler to allocate and copy the entire returned data, potentially causing OOG.
I wouldn't care if we can verify that it always attributed to the paymaster.
An alternative is use `catch {`, and use `getReturnData(1000)`, to cap the amount of revert data we copy .

test/testutils.ts
```diff
@@ -172,6 +172,13 @@ export function decodeRevertReason (data: string, nullIfNoMatch = true): string
   } else if (methodSig === '0x4e487b71') {
     const [code] = ethers.utils.defaultAbiCoder.decode(['uint256'], dataParams)
     return `Panic(${panicCodes[code] ?? code} + ')`
+  } else if (methodSig === '0x8d6ea8be') {
+    const [reason] = ethers.utils.defaultAbiCoder.decode(['string'], dataParams)
+    return `CustomError("${reason as string}")`
```
drortirosh, line 177: `CustomError` is an internal error of the test. this method is for decoding "published API" errors.

test/testutils.ts
```diff
@@ -172,6 +172,13 @@ export function decodeRevertReason (data: string, nullIfNoMatch = true): string
   } else if (methodSig === '0x4e487b71') {
```
drortirosh, line 172: I think this entire method is obsolete... I can see it decodes an obsolete signature of FailedOp (we removed the "paymaster" parameter a long time ago

forshtat: That's right, this method is not really maintained. What should we do about it?

test/testutils.ts
forshtat: This code is in the `testutils.ts` file, why not decode some test-specific errors here?

PR #363, opened 2023-11-02
AA-216 Proposal: Packing UserOperation and paymaster gas limits

shahafn:
This PR is the biggest change we propose for erc4337 v0.7
While the changes below are distinct, it was best to implement this as a single PR

### The changes:
1. The current `verificationGasLimit` field was used for three different purposes (account validation, paymaster validation, and paymaster postOp). We discovered it makes the paymaster vulnerable.
  As a result, we added two new gas fields: `paymasterValidationGasLimit` and `postOpGasLimit`
2. Instead of adding more fields to the UserOperation struct, we encode all four paymaster-related parameters into a single “paymasterData” field.
    This means that if no paymaster is in use, there is only a single “zeros” field, instead of four.
3. We also decided to pack the two account gas parameters (verificationGasLimit and callGasLimit) and pack them into a single field.
4. So in total, we added two more parameters but reduced the size of a simple UserOperation.
5. Note that all the above fields are usually not in use by on-chain contracts (account or paymaster)
6. When used off-chain, we introduced  a separation between the “UserOperation” structure (the JSON API used to contact the bundler) and the “PackedUserOperation” structure in Solidity, which is passed to the contracts (the EntryPoint, account’s validateUserOp, and paymaster’s validatePaymasterUserOp)

### Changes to Account
- The validateUserOp method now has a new method signature, with the new PackedUserOperation structure
- Account should be upgraded to the new version to support the new EntryPoint v0.7

### Supporting multiple entrypoints is still possible:
- `validateUserOperation(PackedUserOperation)` is the new method that supports the new entrypoint.
- validateUserOperation(UserOperationV060) can be implemented to support the old entryPoint.
- The execute (and executeBatch) should be modified to support calls from both entrypoint contracts

### Changes to Paymasters
Paymaster’s paymasterAndData field processing now needs to account for both the 20 bytes of paymaster address and the 32 bytes of gas fields.
We added PAYMASTER_DATA_OFFSET constant to indicate the offset where paymaster-specific data starts


contracts/core/EntryPoint.sol
```diff
@@ -261,7 +263,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
         unchecked {
             // handleOps was called with gas limit too low. abort entire bundle.
             if (
-                gasleft() < callGasLimit + mUserOp.verificationGasLimit + 5000
+                gasleft() <
+                callGasLimit +
+                mUserOp.verificationGasLimit +
+                mUserOp.paymasterVerificationGasLimit +
```
drortirosh: why do you need pmVGL inside inner ?
afaik, you only need postGL

contracts/core/UserOperationLib.sol
```diff
@@ -65,12 +64,24 @@ library UserOperationLib {
...
         );
     }
 
+    function unpackAccountGasLimits(
+        bytes32 accountGasLimits
+    ) internal pure returns(uint128 validationGasLimit, uint128 callGasLimit) {
+        return (uint128(bytes16(accountGasLimits)), uint128(uint256(accountGasLimits)));
+    }
+
+    function unpackPaymasterStaticFields(
```
drortirosh, line 83: is it ever used?

contracts/samples/DepositPaymaster.sol
```diff
@@ -131,7 +131,7 @@ contract DepositPaymaster is BasePaymaster {
 
         (userOpHash);
         // verificationGasLimit is dual-purposed, as gas limit for postOp. make sure it is high enough
-        require(userOp.verificationGasLimit > COST_OF_POST, "DepositPaymaster: gas too low for postOp");
+        require(uint128(bytes16(userOp.paymasterAndData[36:52])) > COST_OF_POST, "DepositPaymaster: gas too low for postOp");
```
drortirosh: merge master, as we removed DepositPaymaster.

contracts/samples/VerifyingPaymaster.sol
```diff
@@ -50,8 +50,8 @@ contract VerifyingPaymaster is BasePaymaster {
                     userOp.nonce,
                     keccak256(userOp.initCode),
                     keccak256(userOp.callData),
-                    userOp.callGasLimit,
-                    userOp.verificationGasLimit,
+                    userOp.accountGasLimits,
+                    uint128(bytes16(userOp.paymasterAndData[20:52])),
```
drortirosh: afaik, VerifyingPaymaster should verify both pmgl and postgl, use entire 256 bits.

contracts/samples/gnosis/EIP4337Manager.sol
```diff
@@ -163,7 +163,7 @@ contract EIP4337Manager is IAccount, GnosisSafeStorage, Executor {
         sig[2] = bytes1(uint8(1));
         sig[35] = bytes1(uint8(1));
         uint256 nonce = uint256(IEntryPoint(manager.entryPoint()).getNonce(address(safe), 0));
-        UserOperation memory userOp = UserOperation(address(safe), nonce, "", "", 0, 1000000, 0, 0, 0, "", sig);
+        UserOperation memory userOp = UserOperation(address(safe), nonce, "", "", bytes32(bytes3(0x0f4240)), 0, 0, 0, "", sig);
```
drortirosh: why is this needed?

contracts/test/MaliciousAccount.sol
```diff
@@ -13,7 +14,14 @@ contract MaliciousAccount is IAccount {
         ep.depositTo{value : missingAccountFunds}(address(this));
         // Now calculate basefee per EntryPoint.getUserOpGasPrice() and compare it to the basefe we pass off-chain in the signature
         uint256 externalBaseFee = abi.decode(userOp.signature, (uint256));
-        uint256 requiredGas = userOp.callGasLimit + userOp.verificationGasLimit + userOp.preVerificationGas;
+        (uint128 verificationGasLimit, uint128 callGasLimit) = UserOperationLib.unpackAccountGasLimits(userOp.accountGasLimits);
+        uint128 paymasterVerificationGasLimit = uint128(bytes16(userOp.paymasterAndData[20:36]));
```
drortirosh: we do need userOpLib functions to extract the gas values..

contracts/samples/gnosis/EIP4337Manager.sol
shahafn: What is "this" that you're asking about? `bytes32(bytes3(0x0f4240))` is there since it's the account gas limits that are now packed.

contracts/test/MaliciousAccount.sol
shahafn: We do or don't need? I added the functions

contracts/core/EntryPoint.sol
shahafn: Ok

contracts/samples/DepositPaymaster.sol
shahafn: Ok, once we're ready with this change, I'll merge master to it

contracts/samples/VerifyingPaymaster.sol
shahafn: Fixed

contracts/samples/gnosis/EIP4337Manager.sol
drortirosh: should use `packGasLimit(callGasLimit, verificationGasLimit)`

same func can be used for pmgas and for acc.gas

contracts/core/UserOperationLib.sol
drortirosh, line 83: I don't think we ever decode the paymaster itself, only paymasterGaslimits
(the reason is that this method is expected to be used from within validatePaymasterUserOp itself)

contracts/samples/DepositPaymaster.sol
drortirosh: ?? how is it still here?

contracts/samples/TokenPaymaster.sol
```diff
@@ -123,15 +123,15 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
     override
     returns (bytes memory context, uint256 validationResult) {unchecked {
             uint256 priceMarkup = tokenPaymasterConfig.priceMarkup;
-            uint256 paymasterAndDataLength = userOp.paymasterAndData.length - 20;
+            uint256 paymasterAndDataLength = userOp.paymasterAndData.length - 52;
```
drortirosh: we need to define these as constants, in UserOpLib. its not good to have so many explicit numerics spread in our code.
e.g. 
PAYMASTER_GASLIMIT_OFFSET (offset of (paymasterGasLimit,paymasterPostOpGasLimit))
PAYMASTER_DATA_OFFSET (offset of "rest of paymaster data")

contracts/core/UserOperationLib.sol
shahafn: Maybe we should have these helper functions (and use them in our EntryPoint/samples) since the logic to unpack is not trivial?

contracts/samples/VerifyingPaymaster.sol
```diff
@@ -50,8 +50,8 @@ contract VerifyingPaymaster is BasePaymaster {
                     userOp.nonce,
                     keccak256(userOp.initCode),
                     keccak256(userOp.callData),
-                    userOp.callGasLimit,
-                    userOp.verificationGasLimit,
+                    userOp.accountGasLimits,
+                    uint256(bytes32(userOp.paymasterAndData[20:52])),
```
drortirosh: afaik, bytes32 is enough (`bytes32(userOp.paymasterAndData[PAYMASTER_GASLIMT_OFFSET])`)

contracts/core/EntryPoint.sol
```diff
@@ -325,8 +329,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
                 "AA93 invalid paymasterAndData"
             );
             mUserOp.paymaster = address(bytes20(paymasterAndData[:20]));
+            mUserOp.paymasterVerificationGasLimit = uint128(bytes16(paymasterAndData[20:36]));
```
drortirosh: 1. use decode helper
2. need to check if need all those in mUserOp

contracts/core/EntryPoint.sol
```diff
@@ -325,8 +329,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
+            mUserOp.paymasterPostOpGasLimit = uint128(bytes16(paymasterAndData[36:52]));
         } else {
             mUserOp.paymaster = address(0);
+            mUserOp.paymasterVerificationGasLimit = 0;
```
drortirosh, line 363: Solidity structs are pre-initialized to all-zeros.

contracts/core/EntryPoint.sol
```diff
@@ -472,13 +478,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
     ) internal returns (bytes memory context, uint256 validationData) {
         unchecked {
             MemoryUserOp memory mUserOp = opInfo.mUserOp;
+        // TODO: Do we actually need that still?
             uint256 verificationGasLimit = mUserOp.verificationGasLimit;
```
drortirosh: indeed, unused (the only reason for verificationGasLimit was to calculate "gas", which was used as gaslimit.
the only reason this test doesn't say "unused local var" is because of the "require", and the only reason the "require" was ever needed was to give a proper error instead of Panic revert on the subtraction, which was removed...)

contracts/core/EntryPoint.sol
```diff
@@ -488,7 +493,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             }
             paymasterInfo.deposit = uint112(deposit - requiredPreFund);
             try
-                IPaymaster(paymaster).validatePaymasterUserOp{gas: gas}(
+                IPaymaster(paymaster).validatePaymasterUserOp{gas: mUserOp.paymasterVerificationGasLimit}(
```
drortirosh, line 505: do we really need a gas limit here?
we verify that innerHandleOp has enough gas for both exec+postOp.
we don't care if postOp uses "leftover" gas from callGasLimit. the limit's real use is for the prepay for it all.

regardless of this, we do need to have a test that if callData uses the entire "callGasLimit", there is postOpGasLimit left for postop

contracts/samples/gnosis/EIP4337Manager.sol
```diff
@@ -163,7 +163,7 @@ contract EIP4337Manager is IAccount, GnosisSafeStorage, Executor {
         sig[2] = bytes1(uint8(1));
         sig[35] = bytes1(uint8(1));
         uint256 nonce = uint256(IEntryPoint(manager.entryPoint()).getNonce(address(safe), 0));
-        UserOperation memory userOp = UserOperation(address(safe), nonce, "", "", 0, 1000000, 0, 0, 0, "", sig);
+        UserOperation memory userOp = UserOperation(address(safe), nonce, "", "", bytes32(bytes16(uint128(0x0f4240))), 0, 0, 0, "", sig);
```
drortirosh: great place to use packGasLimits() helper...

contracts/interfaces/UserOperation.sol
```diff
@@ -7,13 +7,12 @@ pragma solidity ^0.8.12;
...
+ * @param accountGasLimits      - Packed gas limits for validateUserOp and gas limit passed to the callData method call.
  * @param preVerificationGas    - Gas not calculated by the handleOps method, but added to the gas paid.
  *                                Covers batch overhead.
  * @param maxFeePerGas          - Same as EIP-1559 gas parameter.
  * @param maxPriorityFeePerGas  - Same as EIP-1559 gas parameter.
- * @param paymasterAndData      - If set, this field holds the paymaster address and paymaster-specific data.
+ * @param paymasterAndData      - If set, this field holds the paymaster address, verification gas limit, postOp gas limit and paymaster-specific extra data
  *                                The paymaster will pay for the transaction instead of the sender.
  * @param signature             - Sender-verified signature over the entire request, the EntryPoint address and the chain ID.
  */
```
drortirosh, line 18: we said the on-chain type should be called "PackedUserOp", didn't we?

test/paymaster.test.ts
```diff
@@ -95,9 +95,10 @@ describe('EntryPoint with paymaster', function () {
         calldata = await account.populateTransaction.execute(account.address, 0, updateEntryPoint).then(tx => tx.data!)
       })
       it('paymaster should reject if account doesn\'t have tokens', async () => {
-        const op = await fillAndSign({
+        const op = await fillSignAndPack({
           sender: account.address,
-          paymasterAndData: paymaster.address,
+          paymaster: paymaster.address,
+          paymasterPostOpGasLimit: 3e5,
```
drortirosh, line 110: just like other fields, fillAndSign should have default value for postOpGasLimit.

test/samples/TokenPaymaster.test.ts
```diff
@@ -313,7 +312,7 @@ describe('TokenPaymaster', function () {
     })
 
     const preChargeTokens = decodedLogs[0].args.value
-    const requiredGas = BigNumber.from(op.callGasLimit).add(BigNumber.from(op.verificationGasLimit).mul(2)).add(op.preVerificationGas).add(40000 /*  REFUND_POSTOP_COST */)
+    const requiredGas = BigNumber.from(op.callGasLimit).add(BigNumber.from(op.verificationGasLimit).add(BigNumber.from(op.paymasterVerificationGasLimit))).add(BigNumber.from(op.paymasterPostOpGasLimit)).add(op.preVerificationGas).add(40000 /*  REFUND_POSTOP_COST */)
```
drortirosh, line 329: we're missing 
`bnSum = function(){ return [...arguments].reduce((sum,x)=>bn.from(x).add(sum)) }`
and then
`bnSum(op.callGasLimit, op.paymasterVerificationGasLimit, op.paymasterPostOpGasLimit, 40000)`

test/UserOp.ts
```diff
@@ -83,7 +77,10 @@ export const DefaultsForUserOp: UserOperation = {
   preVerificationGas: 21000, // should also cover calldata cost.
   maxFeePerGas: 0,
   maxPriorityFeePerGas: 1e9,
-  paymasterAndData: '0x',
+  paymaster: '0x',
+  paymasterData: '0x',
+  paymasterVerificationGasLimit: 3e5,
+  paymasterPostOpGasLimit: 0,
```
drortirosh, line 83: why default is zero? yes, it does raise the paymaster required balance a little, but simplify using it.
`fillUserOp` can ignore (set to zero) the those pm gas limits if paymaster address is zero

test/UserOp.ts
```diff
@@ -83,7 +77,10 @@ export const DefaultsForUserOp: UserOperation = {
   preVerificationGas: 21000, // should also cover calldata cost.
   maxFeePerGas: 0,
   maxPriorityFeePerGas: 1e9,
-  paymasterAndData: '0x',
+  paymaster: '0x',
```
drortirosh: AddressZero (or left unk, if logic of "fill" handles it)

test/entrypoint.test.ts
```diff
@@ -275,14 +276,18 @@ describe('EntryPoint', function () {
         // we need maxFeeperGas > block.basefee + maxPriorityFeePerGas so requiredPrefund onchain is basefee + maxPriorityFeePerGas
         maxFeePerGas: block.baseFeePerGas.mul(3),
         maxPriorityFeePerGas: block.baseFeePerGas,
-        paymasterAndData: '0x'
+        paymaster: '0x',
```
drortirosh: AddressZero. I wonder how it works with such invalid address.

test/entrypoint.test.ts
```diff
@@ -275,14 +276,18 @@ describe('EntryPoint', function () {
         // we need maxFeeperGas > block.basefee + maxPriorityFeePerGas so requiredPrefund onchain is basefee + maxPriorityFeePerGas
         maxFeePerGas: block.baseFeePerGas.mul(3),
```
drortirosh, line 276: actually, this field looks anachronism: why doesn't it use DefaultUserOp or fillUserOp ?

test/simple-wallet.test.ts
```diff
@@ -53,8 +53,9 @@ describe('SimpleAccount', function () {
 
   it('should pack in js the same as solidity', async () => {
     const op = await fillUserOpDefaults({ sender: accounts[0] })
+    const encoded = encodeUserOp(op)
     const packed = packUserOp(op)
-    expect(await testUtil.packUserOp(op)).to.equal(packed)
+    expect(await testUtil.packUserOp(packed)).to.equal(encoded)
```
drortirosh: took me a white to read this...
clearly, need to rename the solidity helper too, its `testUtils.encodeUserOp`

test/testutils.ts
```diff
@@ -286,3 +275,20 @@ export async function createAccount (
     proxy
   }
 }
+
+export function packAccountGasLimits (validationGasLimit: BytesLike | Hexable | number | bigint, callGasLimit: BytesLike | Hexable | number | bigint): string {
```
drortirosh: the type is "BigNumberish", to cover all numeric and "lookalike"

test/testutils.ts
```diff
@@ -286,3 +275,20 @@ export async function createAccount (
...
+
+export function packPaymasterData (paymaster: string, paymasterVerificationGasLimit: BytesLike | Hexable | number | bigint, postOpGasLimit: BytesLike | Hexable | number | bigint, paymasterData: string): string {
+  return ethers.utils.hexConcat([
+    paymaster, hexZeroPad(hexlify(paymasterVerificationGasLimit, { hexPad: 'left' }), 16),
+    hexZeroPad(hexlify(postOpGasLimit, { hexPad: 'left' }), 16), paymasterData
+  ])
+}
+
+export function unpackAccountGasLimits (accountGasLimits: string): { validationGasLimit: number, callGasLimit: number } {
+  return { validationGasLimit: parseInt(accountGasLimits.slice(2, 34), 16), callGasLimit: parseInt(accountGasLimits.slice(34), 16) }
```
drortirosh, line 321: I wonder, if string slice is clearer than 
```
const B128 = BigNumber.from(1).shl(128);
[top,bottom] = [ BigNumber.from(x).div(B128).toNumber(), BigNumber.from(x).and(B128.sub(1)).toNumber() ]

encode = hexilify(BigNumber.from(top).mul(B128).add(bottom))
```

test/verifying_paymaster.test.ts
```diff
@@ -43,9 +43,17 @@ describe('EntryPoint with VerifyingPaymaster', function () {
 
   describe('#parsePaymasterAndData', () => {
     it('should parse data properly', async () => {
-      const paymasterAndData = hexConcat([paymaster.address, defaultAbiCoder.encode(['uint48', 'uint48'], [MOCK_VALID_UNTIL, MOCK_VALID_AFTER]), MOCK_SIG])
+      const paymasterAndData = hexConcat([
```
drortirosh: missing helper `packPaymasterAndData(paymasterr, pmGas, postOpGas, data?)`
we don't want to mix "protocol" encoding (pm, gas), with "app" encoding (time-range)

contracts/core/EntryPoint.sol
```diff
@@ -261,7 +263,10 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
         unchecked {
             // handleOps was called with gas limit too low. abort entire bundle.
             if (
-                gasleft() < callGasLimit + mUserOp.verificationGasLimit + 5000
+                gasleft() <
+                callGasLimit +
+                mUserOp.paymasterPostOpGasLimit +
```
forshtat, line 297: At this point we already have a `context`, so if it is empty the `postOp` will not be called. I think it may make sense to allow the transaction to go on in this case, instead of always adding `paymasterPostOpGasLimit` to the required gas left?
Say, the UserOp have requested 1'000'000 gas for a `postOp` but it does not need it any more, why should this check fail?

contracts/core/EntryPoint.sol
```diff
@@ -313,8 +318,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
     ) internal pure {
         mUserOp.sender = userOp.sender;
```
forshtat, line 348: Why not have the entire `UserOperation` => `MemoryUserOp` conversion in the `UserOperationLib` library? Construction of a `MemoryUserOp` takes a lot of space and will call to the `UserOperationLib` at least twice.

forshtat (changes requested):

contracts/core/EntryPoint.sol
drortirosh: why should we (entrypoint) waste gas to check it? it is the paymaster's job to put a valid postOpGasLimit, even if context is empty.
There is another edge case: `validatePaymasterUserOp() returns a `context`, and the `postOpGasLimit` is zero: in this case, we still execute the callData and revert the postOp (on OOG) - again, because it is not the EntryPoint's job to validate the userop fields.

contracts/core/EntryPoint.sol
drortirosh: This copy is an internal optimization of EntryPoint. we might even remove copying some fields if we find they are not needed (e.g. used exactly once)
UserOperationLib is for general-use functions, that are expected to be used even outside the EntryPoint itself.

contracts/core/UserOperationLib.sol
```diff
@@ -65,12 +64,24 @@ library UserOperationLib {
...
+        bytes calldata paymasterAndData
+    ) internal pure returns(address paymaster, uint128 validationGasLimit, uint128 postOp) {
+        return (address(bytes20(paymasterAndData[:20])), uint128(bytes16(paymasterAndData[20:36])), uint128(bytes16(paymasterAndData[36:52])));
```
drortirosh: don't use offsets, extract to constants.
also, I think we should use this from entrypoint.

contracts/core/EntryPoint.sol
shahafn: So should I remove this require?
It seems that `gasUsedByValidateAccountPrepayment` is no longer needed then, no?

contracts/core/UserOperationLib.sol
shahafn: Ok

contracts/core/UserOperationLib.sol
shahafn: We do decode it..

contracts/samples/DepositPaymaster.sol
shahafn: Ignore this file

contracts/samples/TokenPaymaster.sol
shahafn: Done

contracts/samples/gnosis/EIP4337Manager.sol
shahafn: There is no helper for it on-chain, since it's not needed in the core contracts atm.

contracts/core/EntryPoint.sol
drortirosh: exactly. it should be removed.
that's what I meant in my previous comment: it is completely unused code, and the only reason solidity didn't complain, is because we use it in "require"...

shahafn: Already removed and pushed

test/UserOp.ts
shahafn: Why not zero?

test/UserOp.ts
shahafn: Does it matter?

contracts/samples/LegacyTokenPaymaster.sol
```diff
@@ -69,13 +70,13 @@ contract LegacyTokenPaymaster is BasePaymaster, ERC20 {
...
       */
-    function _validatePaymasterUserOp(UserOperation calldata userOp, bytes32 /*userOpHash*/, uint256 requiredPreFund)
+    function _validatePaymasterUserOp(PackedUserOperation calldata userOp, bytes32 /*userOpHash*/, uint256 requiredPreFund)
     internal view override returns (bytes memory context, uint256 validationData) {
         uint256 tokenPrefund = getTokenValueOfEth(requiredPreFund);
 
         // verificationGasLimit is dual-purposed, as gas limit for postOp. make sure it is high enough
         // make sure that verificationGasLimit is high enough to handle postOp
-        require(userOp.verificationGasLimit > COST_OF_POST, "TokenPaymaster: gas too low for postOp");
+        require(uint128(bytes16(userOp.paymasterAndData[UserOperationLib.PAYMASTER_POSTOP_GAS_OFFSET:UserOperationLib.PAYMASTER_DATA_OFFSET])) > COST_OF_POST, "TokenPaymaster: gas too low for postOp");
```
drortirosh: simplify a bit: remove the 2nd offset (e.g. `bytes16(paymasterAndData[offset:])`)

contracts/samples/TokenPaymaster.sol
```diff
@@ -118,20 +119,20 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
...
+            uint256 paymasterAndDataLength = userOp.paymasterAndData.length - UserOperationLib.PAYMASTER_DATA_OFFSET;
             require(paymasterAndDataLength == 0 || paymasterAndDataLength == 32,
                 "TPM: invalid data length"
             );
             uint256 preChargeNative = requiredPreFund + (tokenPaymasterConfig.refundPostopCost * userOp.maxFeePerGas);
         // note: as price is in ether-per-token and we want more tokens increasing it means dividing it by markup
             uint256 cachedPriceWithMarkup = cachedPrice * PRICE_DENOMINATOR / priceMarkup;
             if (paymasterAndDataLength == 32) {
-                uint256 clientSuppliedPrice = uint256(bytes32(userOp.paymasterAndData[20 : 52]));
+                uint256 clientSuppliedPrice = uint256(bytes32(userOp.paymasterAndData[UserOperationLib.PAYMASTER_DATA_OFFSET : 84]));
```
drortirosh: remove 2nd constant (84) - unneeded.

contracts/samples/VerifyingPaymaster.sol
```diff
@@ -19,13 +19,13 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 contract VerifyingPaymaster is BasePaymaster {
 
     using ECDSA for bytes32;
-    using UserOperationLib for UserOperation;
+    using UserOperationLib for PackedUserOperation;
 
     address public immutable verifyingSigner;
 
-    uint256 private constant VALID_TIMESTAMP_OFFSET = 20;
+    uint256 private constant VALID_TIMESTAMP_OFFSET = 52;
```
drortirosh: can't reference PAYMASTER_ constants?

contracts/samples/gnosis/EIP4337Manager.sol
```diff
@@ -163,8 +163,8 @@ contract EIP4337Manager is IAccount, GnosisSafeStorage, Executor {
         sig[2] = bytes1(uint8(1));
         sig[35] = bytes1(uint8(1));
         uint256 nonce = uint256(IEntryPoint(manager.entryPoint()).getNonce(address(safe), 0));
-        UserOperation memory userOp = UserOperation(address(safe), nonce, "", "", 0, 1000000, 0, 0, 0, "", sig);
-        UserOperation[] memory userOps = new UserOperation[](1);
+        PackedUserOperation memory userOp = PackedUserOperation(address(safe), nonce, "", "", bytes32(bytes16(uint128(0x0f4240))), 0, 0, 0, "", sig);
```
drortirosh: afaik this packing is not needed: the userop is without paymaster, so `bytes32("")` is enough (afaik, its the shortest way in solidity to write it, better than bytes32(uint(0)), since byte32(0) is not allowed..

test/UserOp.ts
```diff
@@ -5,68 +5,62 @@ import {
...
+    signature: userOp.signature
+  }
+}
+export function encodeUserOp (userOp: UserOperation, forSignature = true): string {
+  const packedUserOp = packUserOp(userOp)
   if (forSignature) {
     return defaultAbiCoder.encode(
       ['address', 'uint256', 'bytes32', 'bytes32',
-        'uint256', 'uint256', 'uint256', 'uint256', 'uint256',
+        'uint256', 'uint256', 'uint256', 'uint256',
```
drortirosh: you changed one of these to bytes32.. should be reflected here.

test/UserOp.ts
drortirosh: you assume the encoder accepts "0x" as a valid address... we try to reduce implicit assumptions and stick to the actual type.
(it passes compliation only because typescript currently doesn't differentiate "address" type from "bytes" type)

test/UserOp.ts
drortirosh: it is a helper for tests, and now it clutters all tests with the need to add it.

test/entrypoint.test.ts
```diff
@@ -348,16 +355,18 @@ describe('EntryPoint', function () {
         const badOp: UserOperation = {
           ...DefaultsForUserOp,
           nonce: TOUCH_PAYMASTER,
-          paymasterAndData: paymaster.address,
+          paymaster: paymaster.address,
+          paymasterVerificationGasLimit: 150000,
           sender: testWarmColdAccount.address
         }
         const beneficiaryAddress = createAddress()
+        const badOpPacked = packUserOp(badOp)
```
drortirosh, line 362: not only here:
you use two different ways in tests.
one (like here) the "badOp" is unpacked, and you have a separate "badOpPacked"
in others (like many places below), you use "op = fillSignAndPack", which means the op is "packedUserOp", and simpler to use.
I think the latter is better: leaves less clutter (and also reduces the diff: the main change becomes  just "fillSign" replaced with "fillSignAndPack"

the only exception (if any) is if we ever create an op, use it and then modify it. in that case, we maybe keeping the "unpacked" version makes sense.

test/entrypoint.test.ts
```diff
@@ -1046,8 +1061,10 @@ describe('EntryPoint', function () {
 
       it('paymaster should pay for tx', async function () {
         await paymaster.deposit({ value: ONE_ETH })
-        const op = await fillAndSign({
-          paymasterAndData: paymaster.address,
+        const op = await fillSignAndPack({
+          paymaster: paymaster.address,
+          paymasterData: '0x',
```
drortirosh: are you required to add "paymasterData" here ? I think it can use the default.

test/paymaster.test.ts
drortirosh: this is why I think we should add postOpGasLimit to the defaults: to avoid adding it in every test below.

forshtat (dismissed):

contracts/core/EntryPoint.sol
```diff
@@ -676,7 +664,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
                     actualGasCost = actualGas * gasPrice;
                     if (mode != IPaymaster.PostOpMode.postOpReverted) {
                         try IPaymaster(paymaster).postOp{
-                            gas: mUserOp.verificationGasLimit
+                            gas: mUserOp.paymasterPostOpGasLimit
                         }(mode, context, actualGasCost)
                         // solhint-disable-next-line no-empty-blocks
                         {} catch {
```
drortirosh, line 670: see `executionGasLimit += mUserOp.verificationGasLimit;`, below
should it be paymasterPostOpGasLimit, or completely removed?

shahafn: Fixed

test/UserOp.ts
shahafn: Fixed, forgot to remove that

contracts/samples/LegacyTokenPaymaster.sol
drortirosh: (NOT removing 2nd offset: solidity creates slightly better code for `bytes32(data[offset1:offset2])` than `bytes32(data[offset1:])`

contracts/interfaces/UserOperation.sol
shahafn: Fixed

test/paymaster.test.ts
shahafn: I don't think that postOp should have a default non-zero value

contracts/core/EntryPoint.sol
yoavw: Except when one does crazy stuff like recycling memory by using the free-pointer back - which we do :)  So we can't assume zeros anywhere.

drortirosh: Yes, we can assume zeros
all storage variables in solidity are initialized to zero - either "stack" or memory variables (structs and arrays).
(there are Yul injected functions "allocate_and_zero_memory..." for that)

shahafn: The paymaster initialization to address(0) was there, I only added the other fields. Do you want to remove all of them?

test/entrypoint.test.ts
shahafn: I didn't touch this code at all

drortirosh (approved):

contracts/core/EntryPoint.sol
yoavw: > Yes, we can assume zeros all storage variables in solidity are initialized to zero - either "stack" or memory variables (structs and arrays). (there are Yul injected functions "allocate_and_zero_memory..." for that)

Interesting, I wonder why they do that. Solidity always allocates new memory (containing zeros) so why do they need to explicitly write zeros?  (We do, due to our asm code, but normal solidity doesn't)

PR #369, opened 2023-11-08
fix: move all free functions to helpers library

vladikopl01:
Solves: #368


drortirosh:
can you please rebase (or merge) develop, so we can run tests on it?

PR #370, opened 2023-11-08
AA-237: paymaster: check EntryPoint's interface

drortirosh:
verify the paymaster is configured with EntryPoint with the expected IEntryPoint interface.


test/paymaster.test.ts
```diff
@@ -68,6 +68,11 @@ describe('EntryPoint with paymaster', function () {
       ownerAddr = await ethersSigner.getAddress()
     })
 
+    it('paymaster should revert on wrong entryPoint type', async () => {
+      await expect(new LegacyTokenPaymaster__factory(ethersSigner).deploy(factory.address, 'ttt', AddressZero))
+        .to.be.revertedWith('')
```
forshtat, line 77: @drortirosh don't you get a full revert message here?

forshtat (changes requested):

test/paymaster.test.ts
drortirosh: zero address reverts with empty string. 
added a test with some contract with different supportsInterface.

shahafn (dismissed):

forshtat (dismissed):

forshtat (approved):

PR #383, opened 2023-12-07
remove unused event PostOpReverted from TokenPaymaster

livingrockrises:
1. Removing the event PostOpReverted which is not being used in current paymaster. I would also like to understand reason behind [adding this](https://github.com/eth-infinitism/account-abstraction/pull/362/files) in the entry point and how to use it in paymasters. (in context of current entry point and upcoming EP)

2. Adding this back in _postOp()

```
 if (mode == PostOpMode.postOpReverted) {
            return; // Do nothing here to not revert the whole bundle and harm reputation
        }
```

since current entry point still uses postOp twice and current code will just revert twice if postOp can not transfer tokens etc. 

pre charging in validatePaymasterUserOp works but it's not a great UX to give pre approval to the paymaster as part of other userOp sponsored by some other paymaster. instead of that dapps would prefer limited approvals given as part of each calldata (based on fee quotes) 

in new entry point with [this](https://github.com/eth-infinitism/account-abstraction/pull/311) it would change couple of things again regarding UX of token paymaster.

### DIFF
--- a/contracts/samples/TokenPaymaster.sol
+++ b/contracts/samples/TokenPaymaster.sol
@@ -42,8 +42,6 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
 
     event UserOperationSponsored(address indexed user, uint256 actualTokenCharge, uint256 actualGasCost, uint256 actualTokenPrice);
 
-    event PostOpReverted(address indexed user, uint256 preCharge);
-
     event Received(address indexed sender, uint256 value);
 
     /// @notice All 'price' variables are multiplied by this value to avoid rounding up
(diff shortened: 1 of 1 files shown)


livingrockrises:
@forshtat @leekt

drortirosh:
postOp is no longer called twice, only once:
if PostOp reverts, then:
- PostOpReverted is emitted
- the account's exec is reverted (as well as paymaster's postOp)
- the UserOpEvent is emitted, with "success=false"
- the account's nonce is incremented (so you can't re-play this UserOp)
- the handleOps continues with the rest of the UserOps in this bundle
- the payment of this userop (validation and reverts of account's exec and postOp) are done by the paymaster.
For naive account, this is a normal revert on-chain, (except that it was not caused by faulty account exec code, but by a faulty paymaster)

drortirosh:
btw - the PostOpReverted is indeed not in use, and should be removed. you can update your pr, for that.

drortirosh (approved):

PR #384, opened 2023-12-07
AA-244 Upgrade open zeppelin 5.0

drortirosh:


contracts/samples/SimpleAccount.sol
```diff
@@ -99,14 +96,14 @@ contract SimpleAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, In
     /// implement template method of BaseAccount
     function _validateSignature(UserOperation calldata userOp, bytes32 userOpHash)
     internal override virtual returns (uint256 validationData) {
-        bytes32 hash = userOpHash.toEthSignedMessageHash();
-        if (owner != hash.recover(userOp.signature))
+        bytes32 hash = MessageHashUtils.toEthSignedMessageHash(userOpHash);
```
forshtat, line 99: Can't we do `using MessageHashUtils for bytes32;` and keep these two lines as-is? I don't know if I prefer the shorter syntax, but we kind of used it everywhere else.

forshtat (dismissed):

contracts/samples/SimpleAccount.sol
drortirosh: Yes, we did use it in the past.
But I find it confusing doing `userOpHash.toEthSgiendMessageHash` followed by `hash.recover`, where both items are generic `bytes32, but using functions from different external libraries.

forshtat (dismissed):

shahafn (approved):

PR #388, opened 2023-12-18
AA-246 update ERC-4337, ERC-7562 (validation rules)

drortirosh:


erc/ERCS/erc-4337.md
```diff
@@ -434,6 +434,16 @@ Otherwise, attackers may be able to use the banned opcodes to detect running on-
...
+
+The entry point-based approach allows for a clean separation between verification and execution, and keeps accounts' logic simple. The alternative would be to require accounts to follow a template where they first self-call to verify and then self-call to execute (so that the execution is sandboxed and cannot cause the fee payment to revert); template-based approaches were rejected due to being harder to implement, as existing code compilation and verification tooling is not designed around template verification.
```
forshtat: In general, I think it is better to avoid talking about "the alternative", as this is usually an infinite list of ways to do something differently.

erc/ERCS/erc-4337.md
```diff
@@ -434,6 +434,16 @@ Otherwise, attackers may be able to use the banned opcodes to detect running on-
...
+
+In this proposal, we expect accounts to have a `validateUserOp` method that takes as input a `UserOperation`, and verify the signature and pay the fee. This method is required to be almost-pure: it is only allowed to access the storage of the account itself, cannot use environment opcodes (eg. `TIMESTAMP`), and can only edit the storage of the account, and can also send out ETH (needed to pay the entry point). The method is gas-limited by the `verificationGasLimit` of the `UserOperation`; nodes can choose to reject operations whose `verificationGasLimit` is too high. These restrictions allow block builders and network nodes to simulate the verification step locally, and be confident that the result will match the result when the operation actually gets included into a block.
```
forshtat: `almost pure` - shouldn't we say something like "restricted in compliance with ERC-7562" instead of vaguely saying "almost pure"?
`is required` - I would say "required" is a bit too strong as we have specified ways to make exceptions. Maybe "is expected" or even "should"?

erc/ERCS/erc-4337.md
```diff
@@ -434,6 +434,16 @@ Otherwise, attackers may be able to use the banned opcodes to detect running on-
 
 When a bundler includes a bundle in a block it must ensure that earlier transactions in the block don't make any UserOperation fail. It should either use access lists to prevent conflicts, or place the bundle as the first transaction in the block.
 
+
+## Rationale
+
+The main challenge with a purely smart contract wallet based account abstraction system is DoS safety: how can a block builder including an operation make sure that it will actually pay fees, without having to first execute the entire operation? Requiring the block builder to execute the entire operation opens a DoS attack vector, as an attacker could easily send many operations that pretend to pay a fee but then revert at the last moment after a long execution. Similarly, to prevent attackers from cheaply clogging the mempool, nodes in the P2P network need to check if an operation will pay a fee before they are willing to forward it.
```
forshtat: `smart contract wallet based account abstraction system` is a mouthful.

"The main challenge of implementing an Account Abstraction system that is based purely on Smart Contract Wallets, lies in designing protections against potential DoS attack vectors."
Something like that?

erc/ERCS/erc-4337.md
```diff
@@ -458,7 +468,7 @@ Paymasters facilitate transaction sponsorship, allowing third-party-designed mec
...
 * Paymasters run the risk of getting griefed, as users could send ops that appear to pay the paymaster but then change their behavior after a block
 
-The paymaster scheme allows a contract to passively pay on users' behalf under arbitrary conditions. It even allows ERC-20 token paymasters to secure a guarantee that they would only need to pay if the user pays them: the paymaster contract can check that there is sufficient approved ERC-20 balance in the `validatePaymasterUserOp` method, and then extract it with `transferFrom` in the `postOp` call; if the op itself transfers out or de-approves too much of the ERC-20s, the inner `postOp` will fail and revert the execution and the outer `postOp` can extract payment (note that because of storage access restrictions the ERC-20 would need to be a wrapper defined within the paymaster itself).
+The paymaster scheme allows a contract to passively pay on users' behalf under arbitrary conditions. for example, a TokenPaymaster uses the precharge refund model. It can precharge the maximum cost with transferFrom in the validatePaymasterUserOp method and then refund the unused portion in the postOp method.
```
forshtat: Why is this change necessary? Also, `TokenPaymaster` is just a specific file name in our repo, I don't think the new text is necessarily better than the old one.

forshtat (changes requested):

erc/ERCS/erc-4337.md
shahafn: I think succinctly writing maybe better:
"Introduction of paymasters allows abstraction of gas, i.e. having a contract, that is not the sender of the transaction, pay for the transaction fees."

erc/ERCS/erc-4337.md
shahafn: Also, this is really not rationale, but a semi deep dive into the technical details of validation rules and gas calculations. Not sure this is the place to do it.

erc/ERCS/erc-4337.md
shahafn: I think that the start of a rationale section should have a short sentence summarizing the actual rationale, not going about it indirectly with the main challenge first.

erc/ERCS/erc-4337.md
yoavw: Does this paragraph even belong in ERC-4337?  Since we moved the restrictions to 7562, @forshtat is right that we should refer to it and not try to make a partial list of restrictions in 4337.

4337 does need to keep the part that isn't in 7562: the separation of validation and execution, having a separate gas limit for it, etc.  In 4337 (and 7560) we only do the minimum to make decentralized AA possible - separating validation to enable restricting it.  Then 7562 makes a proposal of how to restrict it in a way that actually makes it safe.

erc/ERCS/erc-4337.md
yoavw: I believe @drortirosh made this change due to the recent removal of the 2nd postOp.  The old text was based on being able to call postOp a 2nd time after reverting the call.  We replaced it with precharge after making validation a non-view function, and then deprecated the 2nd postOp. The ERC should reflect that.

erc/ERCS/erc-7562.md
```diff
@@ -1,24 +1,45 @@
 ---
-eip: 0
+eip: 7562
 title: Account Abstraction Validation Scope Rules
-description: A set of limitations on validation EVM code that protects Account Abstraction nodes from exponentially complex computations without compensation
+description: A set of limitations on validation EVM code to protect Account Abstraction nodes from computation without compensation
```
yoavw:
```suggestion
description: A set of limitations on validation EVM code to protect Account Abstraction nodes from denial-of-service attacks through unpaid computation.
```
(The real concern is DoS, not wasting some CPU cycles).

erc/ERCS/erc-7562.md
```diff
@@ -1,24 +1,45 @@
...
 type: Standards Track
-category: Core
+category: ERC
 created: 2023-09-01
-requires: 6780
 ---
 
 ## Abstract
 
 This document describes the validation rules we impose on the validation context of Account Abstraction transactions,
```
yoavw:
```suggestion
This document describes the rules we impose on the validation context of Account Abstraction transactions,
```
"validation" appears twice in the sentence.

erc/ERCS/erc-7562.md
```diff
@@ -1,24 +1,45 @@
...
+- abstracting gas payments allows easy onboarding by 3rd party payments, paying with tokens, cross-chain gas payments
+- abstracting execution allows batch transactions
+
+All of the above are missing from the EOA account model.
+
+However, there is one rule a transaction must follow to preserve the decentralized network: once submitted into the network (the mempool), the transaction is guaranteed to pay. This comes to prevent denial of service attacks on the network.
+
+The EOA model implicitly follows the rule: a valid transaction can't become invalid without payment by the account: e.g account balance can't be reduced (except with a higher paying transaction)
+
+For Account-Abstraction system to be sustainable and DoS-protected, we define a set of rules to protect it.
```
yoavw: Here we should first mention that in order to make it sustainable, the account abstraction system must separate validation from execution (or pre-fee-payment and post-fee-payment).  Say that both 4337 and 7560 make this separation, and then we define a set of rules on the pre-payment part to protect the network.

erc/ERCS/erc-7562.md
```diff
@@ -1,24 +1,45 @@
...
+
+For the actual interfaces of those contract-based accounts see the definitions in ERC-4337 and RIP-7560.
+
+This documentation uses the terminology "UserOperation" for a transaction created by a smart contract account, and closely follows [ERC-4337](./erc-4337.md) terminology.
+However, the rules apply to any account-abstraction framework that uses ERC code to perform transaction validation
```
yoavw:
```suggestion
However, the rules apply to any account-abstraction framework that uses EVM code to perform transaction validation and makes a distinction between validation and execution.
```

erc/ERCS/erc-7562.md
```diff
@@ -1,24 +1,45 @@
...
+
+
+## Rationale
+
 All transactions initiated by EOAs have an implicit validation phase where balance, nonce, and signature are
 checked to be valid for the current state of the Ethereum blockchain.
 Once the transaction is checked to be valid by a node, only then another transaction by the same EOA can modify the Ethereum
```
yoavw:
```suggestion
Once the transaction is checked to be valid by a node, only another transaction by the same EOA can modify the Ethereum.
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
-## Specification
+### The high-level goal
+
+The purpose of this specification is to define a consensus between nodes (bundlers or block-builders) when processing incoming UserOperations from an external source.
+This external source for UserOperations is either an end-user node (via RPC) or another node in the p2p network.
+
+The protocol tries to detect "spam" - which are large bursts of UserOperations that cannot be included on-chain (and thus can't pay).
+The network is protected by throttling down requests from such spammer nodes.
+
+All nodes in the network must have the same notion of "spam": otherwise, if some nodes accept some type of UserOperations while others consider them spam, those "forgiving" nodes will be considered "spammers" by the rest of the nodes, and the network effectively gets split.
```
yoavw:
```suggestion
All nodes in the network must have the same definition of "spam": otherwise, if some nodes accept some type of UserOperations and propagate them while others consider them spam, those "forgiving" nodes will be considered "spammers" by the rest of the nodes, and the network effectively gets split.
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+### The processing flow of a UserOperation
+
+- First, a UserOperation is received - either via RPC (submitted on behalf of a single application) or via the p2p protocol, from another node in the mempool.
+- The node performs validation on the UserOperation, and then adds it to its in-memory mempool, and submits it to its peers.
+- Lastly, when building a block, a node collects UserOperations from the mempool, performs a 2nd validation to make sure they are all still valid as a bundle and submits them into the next block.
+
+### The need for 2nd validation before submitting a block
+
+A normal Ethereum transaction in the mempool can be invalidated if another transaction was received with the same nonce. That other transaction had to increase the gas price in order to replace the first one, so it satisfied the rule of "must pay to get included into the mempool"
+With contract-based accounts, since the account depends on some storage, other transactions may cause a given UserOperation to get invalidated, so we must check it before inclusion
```
yoavw:
```suggestion
With contract-based accounts, since the UserOperation validity may depend on mutable state, other transactions may invalidate a previously valid UserOperation, so we must check it before inclusion
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+
+### Rationale of limiting opcodes:
+
+- the validation is performed off-chain, before creating a block. Some opcodes access information that is known only when creating the block.
+- using those opcodes while validating a transaction can easily create a validation rule that will succeed off-chain, but always revert on-chain, and thus cause a DoS attack.
+- a simple example is `require block.number==12345`. It can be valid when validating the UserOperation and adding it to the mempool
+  but will be invalid when attempting to include it on-chain.
```
yoavw:
```suggestion
  but will be invalid when attempting to include it on-chain at a later block.
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+
+### Rationale for limiting storage access
+
+- We need UserOperation validation to be self-contained so that a single storage change can't easily invalidate a large number of UserOperations in the mempool. By limiting UserOperations to access storage associated with the account itself, we know that we can for sure include a single UserOperation for each account in a bundle
```
yoavw: "self-contained" doesn't quite describe it.
```suggestion
- We need UserOperation validations not to overlap so that a single storage change can't easily invalidate a large number of UserOperations in the mempool. By limiting UserOperations to access storage associated with the account itself, we know that we can for sure include a single UserOperation for each account in a bundle
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+- (A bundler MAY include more than one UserOperation of an account in a bundle, MUST first validate them together)
+
+### Rationalte of requiring a stake
```
yoavw:
```suggestion
### Rationale of requiring a stake
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+
+We want to be able to allow globally-used contracts (paymasters, factories) to use less restrictive storage, but still prevent them from
```
yoavw:
```suggestion
We want to be able to allow globally-used contracts (paymasters, factories) to use storage not associated with the account, but still prevent them from
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+spamming the mempool.
+If a contract causes UserOperation failure in their second validation, we can mark such a contract as a spammer.
```
yoavw: We don't mark a contract as a spammer (we do that for bundlers).
```suggestion
If a contract causes too many UserOperations to fail in their second validation after succeeding in their first, we can throttle its use in the mempool.
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+By requiring such a contract to have a stake, we prevent a "Sybil attack", by making it expensive to create a large number of such paymasters to continue the spam attack.
+
+By following the validation rules, we can find out contracts that create spam UserOperations, and block them.
```
yoavw:
```suggestion
By following the validation rules, we can detect contracts that cause spam UserOperations, and throttle them.
```

erc/ERCS/erc-7562.md
```diff
@@ -30,22 +51,67 @@ which means that unrelated `UserOperations` or transactions may invalidate each
...
+The stake comes to prevent the fast re-creation of malicious entities.
+The stake is never slashed (since it is only used for off-chain detection) but is locked for a period of time, which makes such an attack much more expensive.
+
+The stake is never slashed but must be left on a given contract for a while.
```
yoavw: The previous sentence already covered this.  Why is this one needed as well?
```suggestion
```

erc/ERCS/erc-7562.md
```diff
@@ -59,10 +125,36 @@ A `UserOperation` that fails the initial validation by a receiving node without
 considered an attack. The node is expected to apply web2 security measures and throttle requests based on API key,
 source IP address, etc.
 RPC nodes already do that to prevent being spammed with invalid transactions which also have a validation cost.
+P2P nodes already have (and should apply) a scoring mechanism to determine spammer nodes.
 
-### Constants:
+Also, if the invalidation of "n" UserOperations from the mempool has a cost of "n*X", (for X large enough), it is not considered an attack.
```
yoavw:
```suggestion
Also, if the invalidation of `N` UserOperations from the mempool costs `N*X` with a sufficiently large `X`, it is not considered an economically viable attack.
```

erc/ERCS/erc-7562.md
```diff
@@ -59,10 +125,36 @@ A `UserOperation` that fails the initial validation by a receiving node without
...
+- The above value is high, but we take further measures to make such an attack more expensive.
 
 
+## Specification
+
+### Rule types
+
+There are two types of rules:
+
+- **Validation rules** rules that should be applied to each UserOperation before accepting it into the local mempool.
```
yoavw:
```suggestion
- **Network-wide rules** rules that MUST be applied to each UserOperation before accepting it into the local mempool and propagating it.
```

erc/ERCS/erc-7562.md
```diff
@@ -59,10 +125,36 @@ A `UserOperation` that fails the initial validation by a receiving node without
...
+  These rules include the opcode and storage rules.
+    - Failing these validation rules SHOULD drop the UserOperation
+    - Failing these validations during 2nd validation phase (before submitting a bundle) SHOULD trigger
+      reputation drop for the offending entity
```
yoavw:
```suggestion
    - Failing these validations during 2nd validation phase (before submitting a bundle) SHOULD degrade
      the reputation of the offending entity
```

erc/ERCS/erc-7562.md
```diff
@@ -59,10 +125,36 @@ A `UserOperation` that fails the initial validation by a receiving node without
...
+    - Bundler MUST NOT propagate UserOperations that fail the validation rules, otherwise
+      it will be considered a "spammer" by other bundlers in the mempool, and get disconnected.
+
+- **Reputation rules**
```
yoavw:
```suggestion
- **Local rules**
```

erc/ERCS/erc-7562.md
```diff
@@ -59,10 +125,36 @@ A `UserOperation` that fails the initial validation by a receiving node without
...
+  These are "soft" rules, based on the reputation of entities.
+  These rules come to protect the bundler itself from spamming attacks.
+  Bundlers SHOULD not propagate such UserOperations to other bundlers
```
yoavw:
```suggestion
  - Bundlers SHOULD drop such UserOperations without performing validation.
  - Bundlers SHOULD NOT propagate such UserOperations to other bundlers.
  - Bundlers SHOULD NOT consider another bundler a "spammer" if it does.
```

erc/ERCS/erc-7562.md
```diff
@@ -81,11 +173,11 @@ RPC nodes already do that to prevent being spammed with invalid transactions whi
 
 ### **Definitions**:
 1. **Validation Phases**: there are up to 3 phases of validation
-   1. smart account deployment
-   2. smart account validation 
-   3. paymaster validation.
+    1. smart account deployment
+    2. smart account validation
+    3. paymaster validation.
 2. **Entity**: a contract that is explicitly used by the `UserOperation`.
```
yoavw:
```suggestion
2. **Entity**: a contract that is explicitly specified by the `UserOperation`.
```

erc/ERCS/erc-7562.md
```diff
@@ -100,8 +192,8 @@ RPC nodes already do that to prevent being spammed with invalid transactions whi
 ### Reputation Definitions:
 1. **opsSeen**: a per-entity counter of how many times a unique valid `UserOperation` referencing this entity
    was received by this bundler.
-   This includes `UserOperation` received via an incoming RPC calls or through a P2P mempool protocol.
-   * For a `paymaster`, this value is not incremented if factory or account validation fails.    
+   This includes `UserOperation` received via incoming RPC calls or through a P2P mempool protocol.
+    * For a `paymaster`, this value is not incremented if factory or account validation fails.
```
yoavw: It was already incremented when the op was received and passed 1st validation. If a factory or a staked account caused the 2nd validation to fail, should we decrement paymaster's opsSeen in order to "unsee" it (`opsSeen[paymaster] = opsSeen[paymaster] > 0 ? opsSeen[paymaster]-1 : 0`) ?

erc/ERCS/erc-7562.md
```diff
@@ -129,10 +221,10 @@ To help make sense of these params, note that a malicious paymaster can at most
 ### Running the Validation Rules
 
 1. A block builder or a bundler should perform a full validation before accepting a `UserOperation` into its mempool.
-2. During validation phase, the bundler should trace the execution and apply all the rules defined in this document.
+2. During the validation phase, the bundler should trace the execution and apply all the rules defined in this document.
 3. A bundler should also perform a full validation of the entire bundle before submission.
 4. The validation rules prevent an unstaked entity from detecting the bundle validation.
-   However, it is possible for a malicious staked entity to detect the bundle validation and cause a revert.
+   However, a malicious staked entity can detect it runs in a bundle validation and cause a revert.
```
yoavw:
```suggestion
   However, a malicious staked entity can detect that it is running in a bundle validation and cause a revert.
```

erc/ERCS/erc-7562.md
```diff
@@ -129,10 +221,10 @@ To help make sense of these params, note that a malicious paymaster can at most
...
 5. The failed `UserOperation` should be dropped from the bundle.
 6. The staked entity that caused a revert violated the Account Abstraction rules and should be marked as `THROTTLED`.
```
yoavw: We don't have a definition of marking something as `THROTTLED`.  Shouldn't we say how it is done, i.e. specify values to assign to its opsSeen and opsIncluded to cause throttling for a certain amount of time?  (Rule SREP-050 explains how it actually gets BANNED, so no need to go into details, but we also don't have a way to "mark as THROTTLED".

erc/ERCS/erc-7562.md
```diff
@@ -144,7 +236,7 @@ To help make sense of these params, note that a malicious paymaster can at most
 2. Once a `UserOperation` is received from another bundler it should be verified locally by a receiving bundler.
 3. A received `UserOperation` may fail any of the reasonable static checks, such as: \
    invalid format, values below minimum, submitted with a blockhash that isn't recent, etc. \
-   In this case the bundler should drop this particular `UserOperation` but keep the connection.
+   In this case, the bundler should drop this particular `UserOperation` but keep the connection.
 4. The bundler should validate the `UserOperation` against the nonces of last-included bundles. \
```
yoavw: nonces?  Not the UserOp hashes?

erc/ERCS/erc-7562.md
```diff
@@ -144,7 +236,7 @@ To help make sense of these params, note that a malicious paymaster can at most
...
    Silently drop `UserOperations` with `nonce` that was recently included.
```
yoavw: Ditto - did you mean nonce, or hash?

erc/ERCS/erc-7562.md
```diff
@@ -144,7 +236,7 @@ To help make sense of these params, note that a malicious paymaster can at most
...
    This invalidation is likely attributable to a network racing condition and should not cause a reputation change.
```
yoavw:
```suggestion
   This invalidation is likely attributable to a network race condition and should not cause a reputation change.
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
     * **[OP-031]** `CREATE2` is allowed exactly once in the deployment phase and must deploy code for the "sender" address.
 * Access to an address without a deployed code is forbidden:
     * **[OP-041]** For `EXTCODE*` and `*CALL` opcodes
```
yoavw:
```suggestion
    * **[OP-041]** For `EXTCODE*` and `*CALL` opcodes.
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
-    * **[OP-042]** Exception: access to "sender" address is allowed.
+    * **[OP-042]** Exception: access to the "sender" address is allowed.
       This is only possible in `factory` code during the deployment phase.
 * Allowed access to the `EntryPoint` address:
     * **[OP-051]** May call `EXTCODESIZE ISZERO`\
-      This pattern is used to check destination has code before `depositTo` function is called.
+      This pattern is used to check destination has a code before the `depositTo` function is called.
     * **[OP-052]** May call `depositTo(sender)` with any value from either the `sender` or `factory`
     * **[OP-053]** May call the fallback function from the `sender` with any value
     * **[OP-054]** Any other access to the `EntryPoint` is forbidden
```
yoavw:
```suggestion
    * **[OP-052]** May call `depositTo(sender)` with any value from either the `sender` or `factory`.
    * **[OP-053]** May call the fallback function from the `sender` with any value.
    * **[OP-054]** Any other access to the `EntryPoint` is forbidden.
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
 * `*CALL` opcodes:
     * **[OP-061]** `CALL` with `value` is forbidden. The only exception is a call to the `EntryPoint` described above.
     * **[OP-062]** Precompiles:
         * Only allowed the core 9 precompiles.\
```
yoavw: Shouldn't we already include the new secp256r1 precompile defined in https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md at address `0x100`, only on chains that adopted RIP-7212?  We already know it's safe, that it's getting adopted by some chains, and that it will be called during validation.

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
+  entity, or referenced library, may not be changed.\
+  If the code is modified, the UserOperation is considered invalid.
 
 ### Storage rules.
 
-The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase are limited as follows:
+The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase is limited as follows:
 
 * **[STO-010]** Access to the "account" storage is always allowed.
 * Access to associated storage of the account in an external (non-entity contract) is allowed if either:
```
yoavw:
```suggestion
* Access to associated storage of the account in an external (non-entity) contract is allowed if either:
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
 * `*CALL` opcodes:
     * **[OP-061]** `CALL` with `value` is forbidden. The only exception is a call to the `EntryPoint` described above.
     * **[OP-062]** Precompiles:
         * Only allowed the core 9 precompiles.\
-          Specifically the computation precompiles that do not access anything in the blockchain state or environment.
-* **[OP-070] Transient Storage slots defined in [EIP-1153](./eip-1153.md) and accessed using `TLOAD` (`0x5c`) and `TSTORE` (`0x5d`) opcodes
-    are treated exactly like persistent storage (SLOAD/SSTORE)
+          Specifically, the computation precompiles that do not access anything in the blockchain state or environment.
+* **[OP-070]** Transient Storage slots defined in [EIP-1153](./eip-1153) and accessed using `TLOAD` (`0x5c`) and `TSTORE` (`0x5d`) opcodes
+  are treated exactly like persistent storage (SLOAD/SSTORE)
```
yoavw:
```suggestion
  are treated exactly like persistent storage (SLOAD/SSTORE).
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
+  If the code is modified, the UserOperation is considered invalid.
 
 ### Storage rules.
 
-The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase are limited as follows:
+The permanent storage access with `SLOAD` and `SSTORE` instructions within each phase is limited as follows:
 
 * **[STO-010]** Access to the "account" storage is always allowed.
 * Access to associated storage of the account in an external (non-entity contract) is allowed if either:
     * **[STO-021]**  The account already exists
```
yoavw:
```suggestion
    * **[STO-021]**  The account already exists.
```

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
     * **[STO-022]**  There is an `initCode` and the `factory` contract is staked.
 * If the entity (`paymaster`, `factory`) is staked, then it is also allowed:
     * **[STO-031]** Access the entity's own storage.
-    * **[STO-032]** Read/Write Access to storage slots that is associated with the entity, in any non-entity contract.
+    * **[STO-032]** Read/Write Access to storage slots that are associated with the entity, in any non-entity contract.
     * **[STO-033]** Read-only access to any storage in non-entity contract.
 * **[STO-040]** `UserOperation` may not use an entity address (`factory`/`paymaster`/`aggregator`) that is used as an "account" in another `UserOperation` in the mempool. \
```
yoavw: Does this need to be a mempool rule (which implies that it's a "network-wide" mandatory rule) rather than a bundle rule (which implies that it's local, like reputation)?  Isn't it safe as long as they're not included in the same bundle?

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
-This means that a `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
+  This means that `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
 * **[STO-041]** A contract whose Associated Storage slot is accessed during a `UserOperation` may not be an address of a "sender" in another `UserOperation` in the mempool.
```
yoavw: 1. Ditto - shouldn't this be a bundle-level rule rather than mempool-level? It's only unsafe if it happens in the same bundle.
2. I think the rule is the opposite of what we discussed.  It means that a sender can be censored by someone sending a UserOperation that accessess a storage slot associated with it.  IIRC we decided to make it the other way around - always favoring a "sender" over a use of the same address as a non-entity. This way it can't lead to censorship because attackers can't produce a valid UserOperation where the sender is an address they don't already control.

erc/ERCS/erc-7562.md
```diff
@@ -180,42 +272,42 @@ To help make sense of these params, note that a malicious paymaster can at most
...
 
 ### Staked Entities Reputation Rules
```
yoavw: In `SREP-050` (which I can't comment on directly in this PR), we **increment** opsSeen by 10000 (and not touch opsIncluded) to cause a ban. It won't ban a very-popular paymaster that has a high opsIncluded.
- Shouldn't we instead **set** opsIncluded=0 and opsSeen=10000, effectively deleting the reputation it accumulated so far?
- Maybe calculate the amount of time it'll remained banned by these numbers?

erc/ERCS/erc-7562.md
```diff
@@ -238,9 +330,8 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
     * Do not add a `UserOperation` to the mempool if the maximum total gas usage, including the new `UserOperation`, is above the deposit of the `paymaster` at the current gas price.
```
yoavw:
```suggestion
    * Do not add a `UserOperation` to the mempool if the maximum total gas cost, including the new `UserOperation`, is above the deposit of the `paymaster` at the current gas price.
```

Should also be changed in the line above, which I can't touch in this PR review: It's not the total gas but the total gas **cost**.

erc/ERCS/erc-7562.md
```diff
@@ -238,9 +330,8 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
     * TODO: allow "fractional reserve" based on reputation?
```
yoavw: We can't finalize this with a **TODO**.  And I thought we'll allow staked paymasters to over-commit (I wouldn't call it "fractional reserve") by a certain ratio.  This prevents throttling a popular TokenPaymaster if it acts responsibly (either swapping the tokens in postOp when balance gets low, or gets replenished by its owner immediately when needed).  Without this, a highly used paymaster requires a huge deposit that makes it capital-inefficient even if it implemented something like automatic swapping.  Have we decided otherwise?

erc/ERCS/erc-7562.md
```diff
@@ -238,9 +330,8 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
 * **[EREP-020]** A staked factory is "accountable" for account breaking the rules. \
-That is, if the `validateUserOp()` is rejected for any reason, it is treated as if the factory caused this failure, and thus this affects its reputation.
-* **[EREP-030]** A Staked Account is accountable for failures in other entities (`paymaster`, `aggregator`) even if they are staked. \
-(see [https://github.com/eth-infinitism/account-abstraction/issues/274](https://github.com/eth-infinitism/account-abstraction/issues/274))
+  That is, if the `validateUserOp()` is rejected for any reason, it is treated as if the factory caused this failure, and thus this affects its reputation.
```
yoavw:
```suggestion
  That is, if the `validateUserOp()` is rejected for any reason in a `UserOperation` that has an `initCode`, it is treated as if the factory caused this failure, and thus this affects its reputation.
```

erc/ERCS/erc-7562.md
```diff
@@ -238,9 +330,8 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
+* **[EREP-030]** A Staked Account is accountable for failures in other entities (`paymaster`, `aggregator`) even if they are staked.
 * **[EREP-040]** An `aggregator` must be staked, regardless of storage usage.
 * **[EREP-050]** An unstaked `paymaster` may not return a `context`.
     * TODO: can be removed once we remove the "2nd postOp".
```
yoavw: We already removed the 2nd postOp, right?  If so, we should remove these two lines.

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
 
 * Definitions:
     * **`opsSeen`, `opsIncluded`, and reputation calculation** are defined above.
-    * `UnstakedReputation` of an entity is a maximum number of entries using this entity allowed in the mempool.
-    * `opsAllowed` is a reputation based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
-* **[UREP-010]** `UserOperation` with unstaked sender are only allowed up to `SAME_SENDER_MEMPOOL_COUNT` times in the mempool.
+    * `UnstakedReputation` of an entity is the maximum number of entries using this entity allowed in the mempool.
```
yoavw:
```suggestion
    * `UnstakedReputation` of an entity determines the maximum number of entries using this entity allowed in the mempool.
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
+    * `opsAllowed` is a reputation-based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
+* **[UREP-010]** `UserOperation` with unstaked sender is only allowed up to `SAME_SENDER_MEMPOOL_COUNT` times in the mempool.
```
yoavw:
```suggestion
* **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
 * **[UREP-020]** For other entities: \
-    `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
-    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for new entity
+  `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
```
yoavw: I don't understand this calculation.  @shahafn did you mean to have `+` in the 2nd part of the expression?  The parentheses are also mismatched (the last one is not closed). 

As it is now, the expression sums three values:
- SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT - always 10
- (inclusionRate * INCLUSION_RATE_FACTOR) - with inclusionRate being opsIncluded/opsSeen (max 1) and INCLUSION_RATE_FACTOR of 10, this value is at most 10.
- (min(opsIncluded, 10000) - climbs up to 10000 and stays there forever because it is not affected by opsSeen.

So opsAllowed is basically `10 + 10 + opsIncluded`, and doesn't drop when the entity starts misbehaving.  Please check the calculation.

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
+    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for a new entity
 * **[UREP-030]** If an unstaked entity causes an invalidation of a bundle, its `opsSeen` is set to `1000`, effectively blocking it from inclusion for 24 hours.
```
yoavw: Not reflected by the above calculation.  Changing `opsSeen` affects `inclusionRate` (unless if it already has more than 1000 opsIncluded, in which case `inclusionRate` actually becomes bigger than 1).  But then `inclusionRate` itself doesn't really affect `opsAllowed` in the above calculation.

Also, this rule should be consistent `SREP-050` (which specifies opsSeen increment by 10000, and also needs to be fixed).

I think both UREP-030 and SREP-050 should **set** opsSeen to a high number (whether 1000 or 10000 should be calculated to achieve the desired duration) and opsIncluded to 0.

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
 
 ### Alt-mempools rules:
 
 Alternate mempool is an agreed-upon rule that the bundlers may opt in to.
```
yoavw:
```suggestion
Alternate mempool is an agreed-upon rule that the bundlers may opt into, in addition to the canonical mempool
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
+The alt-mempool "topic" is a unique identifier. By convention, this is the IPFS hash of the document describing (in clear test and YAML file) the specifics of this alt mempool
 
 * **[ALT-010]** The bundler listens to the alt-mempool "topic" over the P2P protocol
 * **[ALT-020]** The alt mempool rules MUST be checked only when a canonical rule is violated
     * That is, if validation follows the canonical rules above, it is not considered part of an alt-mempool.
 * **[ALT-030]** Bundlers SHOULD forward `UserOperations` to other bundlers only once, regardless of how many (shared) alt-mempools they have. \
-The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides for which alt-mempools to "tag" it.
+  The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides for which alt-mempools to "tag" it.
```
yoavw:
```suggestion
  The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides which alt-mempools to propagate it to.
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
 
 
 ## Security Considerations
 
+This document describes the security considerations bundlers must take to protect themselves (and the entire mempool network)
+from denial-of-service attacks.
+
 ### Possible Attacks
 
 Below is a list of possible attacks that were considered and a reference to the above rule that prevents it.
```
yoavw:
```suggestion
Below are examples of possible attacks that were considered and a reference to the above rule that prevents them.
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
 
 This list does not attempt to be an exhaustive list of attacks.
-These attacks are examples provided to describe and rationalise the reason for the above rules.
+These attacks are examples provided to describe and rationalize the reason for the above rules.
 
 #### Sample Known (expensive) attack:
 
 * This attack can't be fully mitigated by the above validation rules.
-  But the rules made sure it is expensive to maintain.
+  But the rules made sure it was expensive to maintain.
```
yoavw:
```suggestion
  But the rules make it expensive to sustain.
```

erc/ERCS/erc-7562.md
```diff
@@ -249,44 +340,48 @@ That is, if the `validateUserOp()` is rejected for any reason, it is treated as
...
 * Deploy a paymaster which has the following line in the validation function: \
-`require((perSenderMap[userOp.sender]&1)==0);`
+  `require((perSenderMap[userOp.sender]&1)==0);`
 * Submit a large number of `UserOperations` with different senders into the mempool.
 * Front-run the mempool, and modify the `perSenderMap` for each sender.
 * By the time the `UserOperations` are validated the second time they are all invalid.
-* The cost of first run is 20000 gas per account, but the second run is 5000 gas per account.
+* The cost of the first run is 20000 gas per account, but the second run is 5000 gas per account.
```
yoavw:
```suggestion
* The cost of the first iteration is 20000 gas per account, and the second one is 5000 gas per account.
```

erc/ERCS/erc-7562.md
```diff
@@ -299,41 +394,41 @@ These attacks are examples provided to describe and rationalise the reason for t
 * Total attack cost
     * Polygon (@150 gwei) - $18 + $1.5 per block ($2700/hour)
     * Mainnet: (@40 gwei) - $51840 + $5040 per block ($1.5M/hour)
```
yoavw: Based on what eth price and matic price?

erc/ERCS/erc-7562.md
```diff
@@ -299,41 +394,41 @@ These attacks are examples provided to describe and rationalise the reason for t
...
-    * Note that the above costs ignore the gas-price increase because the underlying network is also overloaded.
+    * Note that the above costs ignore the gas price increase because the underlying network is also overloaded.
```
yoavw:
```suggestion
    * Note that the above costs ignore the gas price increase caused by the increased load on the network.
```

erc/ERCS/erc-7562.md
```diff
@@ -299,41 +394,41 @@ These attacks are examples provided to describe and rationalise the reason for t
...
 * Mitigations
-    * A **staked factory:** 
+    * A **staked factory:**
         * Can attempt such an attack once, since after invalidating a single bundle creation its reputation drops and another staked factory has to be used.
```
yoavw:
```suggestion
        * Can only run one iteration of this attack, since after invalidating a single bundle creation its reputation drops and another staked factory has to be used.
```

erc/ERCS/erc-7562.md
```diff
@@ -299,41 +394,41 @@ These attacks are examples provided to describe and rationalise the reason for t
...
         * So a successful attack needs to deploy and stake a new paymaster on every block. The old paymaster's stake is locked and can be re-used only after 24 hours.
     * For **unstaked factory**:
         * As per **[UREP-020]**, it is initially allowed only 10 `UserOperations` in the mempool.
-        * That value increases fast, but drops (to negative) on invalidations.
+        * That value increases fast but drops (to negative) on invalidations.
         * It needs to successfully deploy 10 `UserOperations` to get a reputation to deploy a batch of 1000 (which attack).
```
yoavw: I don't understand this sentence.  I assume it needs to successfully deploy 10 **accounts** rather than **UserOperations**, but I don't understand the **(which attack)** part.

In general, if I didn't know this attack, I doubt I'd be able to understand the attack and its high costs from just reading this section.  Is it really valuable to have this example in the ERC?  Especially highlighting it as the first item, before even discussing attacks fully mitigated by the rules?

erc/ERCS/erc-7562.md
```diff
@@ -342,3 +437,7 @@ effectively censoring the other sender.
            and the only mitigation at the moment is to bump the gas price of the censored `UserOperation`.
         2. There should be at least one "proper" bundler in the network, that follows rule [STO-041],
```
yoavw: 1. As it is now, STO-041 is written backwards and actually enables this attack.
2. Let's discuss this section. I'm not sure it's valuable the way it's written right now.

yoavw (commented):
Lots of comments/suggestions...

erc/ERCS/erc-4337.md
drortirosh: simplified. it used to try to explain both motivation, implementation and alternate implementations.

erc/ERCS/erc-7562.md
drortirosh: strange. it already reads:

> This document describes the security considerations bundlers must take to protect themselves (and the entire mempool network)
> from denial-of-service attacks.
>

erc/ERCS/erc-7562.md
drortirosh: yes. the same nonce might get included by another node using a higher gas fee, hence different hash but the same nonce.

erc/ERCS/erc-7562.md
drortirosh: right.. this is not a "real" storage rule, but local one - but it is not reputation. we don't have category for that.

erc/ERCS/erc-7562.md
drortirosh: rephrased.

erc/ERCS/erc-7562.md
drortirosh: We can change the rule, and allow staked paymaster any # of userops.
It does require tuning the opsSeen/opsIncluded rules, so that such paymasters are only throttled (not banned) on overload.
but it is complex, and I wanted to keep the basic doc simpler.

erc/ERCS/erc-7562.md
yoavw: Got it.  The `validate` part confused me.  It doesn't actually validate the UserOp, just compares its nonce to other UserOps of the last bundles before validating the current one.
```suggestion
4. The bundler should check the `UserOperation` against the nonces of last-included bundles. \
```

erc/ERCS/erc-7562.md
yoavw: That's why I didn't think the separation should be reputation rules vs. non-reputation rules, but "objective" vs. "subjective", or "network-wide" vs. "local".  That's the separation that bundler devs should care about, since they can't deviate from network-wide rules but have some discretion on local rules.

And regardless of category, does it need to apply for "in the mempool" and not just "in the same bundle"?  We implemented it at the mempool level in our bundler because it was easier, but from security perspective do we care if it happens but not in the same bundle?

erc/ERCS/erc-7562.md
yoavw: I'm looking at the current wording and I still don't read it as what we intended to do:

`* **[STO-041]** A contract whose storage is associated with any entity of a `UserOperation` may not be a "sender" of another UserOperation in the mempool.`

It still sounds to me like "if you have two UserOperations, op1 and op2, where op1.sender is accessed by op2, exclude **op1**".  Whereas the intended behavior is the opposite: "if you have two UserOperations, op1 and op2, where op1.sender is accessed by op2, exclude **op2**"

Isn't that the intended behavior?

And also, does it need to be "in the mempool" rather than "in the same bundle", or is that just an implementation detail in our own bundler?

erc/ERCS/erc-7562.md
yoavw: Right, it's best to keep the doc simpler if possible and not change the way reputation works. Should we still allow some over-commitment ratio for staked entities?  Or just leave it at 1:1?

In any case this is a "local" rule so bundlers may choose to manage their own ratio and ignore this rule if they know what they're doing.  That's why it's important to mark local rules as such.

erc/ERCS/erc-7562.md
yoavw: I see SREP-050 was fixed now. Shouldn't UREP-030 work the same way?

(and the calculation above still seems wrong)

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,64 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
-Alternate mempool is an agreed-upon rule that the bundlers may opt in to.
+Alternate mempool is an agreed-upon rule that the bundlers may opt into, in addition to the canonical mempool
+The alt-mempool "topic" is a unique identifier. By convention, this is the IPFS hash of the document describing (in clear test and YAML file) the specifics of this alt mempool
 
 * **[ALT-010]** The bundler listens to the alt-mempool "topic" over the P2P protocol
 * **[ALT-020]** The alt mempool rules MUST be checked only when a canonical rule is violated
     * That is, if validation follows the canonical rules above, it is not considered part of an alt-mempool.
 * **[ALT-030]** Bundlers SHOULD forward `UserOperations` to other bundlers only once, regardless of how many (shared) alt-mempools they have. \
-The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides for which alt-mempools to "tag" it.
+  The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides which alt-mempools to propagate it to.
```
yoavw: We should add a section about the reputation of alt mempool rules ("AREP-.."?), and throttling them separately to protect them from each other and to protect the canonical mempool.

It's very similar to staked entity reputation. No need for staking because alt mempool rules are explicitly whitelisted by the participating bundlers, but otherwise it's like staked entities. Each rule has opsSeen (incremented when a userop triggers that rule) and opsIncluded. Rules get throttled/banned the same way. If the ratio drops too low, throttle. If a userop that matched the rule causes a bundle to revert, trigger immediate throttling by updating opsSeen and opsIncluded - same as you would with a staked entity.

erc/ERCS/erc-7562.md
shahafn: Yes, I think it should have been 
`opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + inclusionRate * (INCLUSION_RATE_FACTOR + min(opsIncluded, 10000))`

erc/ERCS/erc-7562.md
yoavw: Oddly my reply to this seem to have disappeared, and the comment was resolved.  Unresoving because this rule still says the opposite of what we intended to do.  It limits the sender, not the contract that attacks it by using it in associated storage.  I previously wrote a detailed explanation which is apparently lost somehow.

yoavw: The way I'd interpret the rule as written above:
"If the mempool has op1 and op2, and op1 accesses someContract.balances[op2.sender], exclude op2".
which allows op1 to censor op2.

The intended behavior is the opposite:
"If the mempool has op1 and op2, and op1 accesses someContract.balances[op2.sender], exclude op1".

erc/ERCS/erc-7562.md
```diff
@@ -179,44 +274,49 @@ To help make sense of these params, note that a malicious paymaster can at most
...
     * **[STO-033]** Read-only access to any storage in non-entity contract.
+
+### Local Rules
+
+These are storage rules that help to protect the bundler from denial of service:
 * **[STO-040]** `UserOperation` may not use an entity address (`factory`/`paymaster`/`aggregator`) that is used as an "account" in another `UserOperation` in the mempool. \
-This means that a `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
-* **[STO-041]** A contract whose Associated Storage slot is accessed during a `UserOperation` may not be an address of a "sender" in another `UserOperation` in the mempool.
+  This means that `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
+* **[STO-041]** A contract whose storage is associated with any entity of a `UserOperation` may not be a "sender" of another UserOperation in the mempool.
```
yoavw: Somehow my comment about this keeps disappearing as things move around.  I think this rule is **still** broken.

The way I'd interpret the rule as written above:
"If the mempool has op1 and op2, and op1 accesses someContract.balances[op2.sender], exclude **op2**".
which allows op1 to censor op2.

The intended behavior is the opposite:
"If the mempool has op1 and op2, and op1 accesses someContract.balances[op2.sender], exclude **op1**".

There is no concern that op2 will censor op1 because it requires the attacker to control the address and have it pass validation. But the opposite is not true - op1 can easily censor op2 if it is given priority.

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,64 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
 * **[SREP-040]** An `OK` staked entity is unlimited by the reputation rule.
     * Allowed in unlimited numbers in the mempool.
     * Allowed in unlimited numbers in a bundle.
-* **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` is incremented by `10000`, causing it to be `BANNED`.
+* **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` set to `10000`, and `opsIncluded` to zero, causing it to be `BANNED`.
```
yoavw: We realized that setting opsIncluded to zero is too harsh - requiring opsSeen to go all the way down to zero.  I think @shahafn said he'll work on a better way to trigger banning for an expected duration.  This was in the context of UREP-030 but whatever we do in UREP-030 should apply here as well.

erc/ERCS/erc-7562.md
drortirosh: rephrased as
`UserOperation` may not use associated storage (of either its account or from staked entity) in a contract that is a "sender" of another UserOperation in the mempool.

erc/ERCS/erc-7562.md
yoavw: @drortirosh so what do we decide here?  Unsee it, or do you see a better solution?

erc/ERCS/erc-7562.md
yoavw: @shahafn still pending the definition in terms of opsSeen/opsIncluded based on your simulation.

erc/ERCS/erc-7562.md
yoavw: @shahafn can you share your simulation graph and the recommended numbers for opsSeen and opsIncluded?

erc/ERCS/erc-7562.md
yoavw: @shahafn following our discussions about it, still need the correct expression here.

erc/ERCS/erc-7562.md
yoavw: @shahafn this one is also waiting for the same opsSeen/opsIncluded values based on your simulation. Should be the same as SREP-050 (once it's fixed).

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,72 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
     * That is, if validation follows the canonical rules above, it is not considered part of an alt-mempool.
+*  **[ALT-021]** Such a `UserOperation` (that violates the cannonical rules) is checked against all the "alternate mempools", and is considered part of all those alt-mempools 
 * **[ALT-030]** Bundlers SHOULD forward `UserOperations` to other bundlers only once, regardless of how many (shared) alt-mempools they have. \
-The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides for which alt-mempools to "tag" it.
+  The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides which alt-mempools to propagate it to.
 
+### Alt-mempool reputation:
+
+
+* **[AREP-010]** each alt-mempool manages "opsSeen" and "opsIncluded", much like entities. The opsSeen is incremented after `UserOperation` initial validation, where it is considered part of this mempool.
```
yoavw: Question: should we manage a reputation per alt mempool, or per rule?  I.e. if an alt mempool has 5 rules and one of them turns out bad, do we throttle the entire alt mempool, or only UserOps that trigger that particular rule?

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,72 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
+    The "opsIncluded" is incremented after this UserOperation is included on-chain (either by this bundler, or another)
+* **[AREP-020]** the alt-mempool becomes THROTTLED based on the [Reputation Calculation](#reputation-calculation) 
+* **[AREP-030]** if such a UserOperation that passed initial validation, later causes a bundle to revert, it is immediately marked as BANNED, by setting its opsIncluded to 0 and opsSeen to 10000
```
yoavw: 1. I think we established that (10000,0) doesn't give us the result we wanted.  Waiting for @shahafn 's simulation to figure out the numbers.
2. I wonder if we should have the same numbers as for staked entities, or maybe just THROTTLE. The way BANNING works for alt mempools now, it enables DoS against the alt mempool.  Attack:
    1. Deploy a staked entity that detects bundling and reverts after passing single-simulation.
    2. Send a UserOp that triggers one of the alt mempool rules and uses this staked entity.
    3. Both the malicious staked entity and the alt mempool get BANNED for the same duration.
    4. Repeat this UserOp every time the entity gets unbanned.  The parity between SREP and AREP banning means that a single stake can keep an alt mempool down indefinitely - permanent DoS at the cost of ~$1000.
    
@drortirosh How do we mitigate?  Have a shorter banning period for alt mempool rules?  Remove AREP-030 and rely solely on AREP-020 for alt mempools?  Any other suggestions?

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,72 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
 * **[ALT-020]** The alt mempool rules MUST be checked only when a canonical rule is violated
     * That is, if validation follows the canonical rules above, it is not considered part of an alt-mempool.
+*  **[ALT-021]** Such a `UserOperation` (that violates the cannonical rules) is checked against all the "alternate mempools", and is considered part of all those alt-mempools 
 * **[ALT-030]** Bundlers SHOULD forward `UserOperations` to other bundlers only once, regardless of how many (shared) alt-mempools they have. \
-The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides for which alt-mempools to "tag" it.
+  The receiving bundler validates the `UserOperations`, and based on the above rules (and subscribed alt-mempools) decides which alt-mempools to propagate it to.
 
+### Alt-mempool reputation:
+
+
```
yoavw:
```suggestion
Alt-mempools are served by the same bundlers participating in the canonical mempool, but change the rules and may introduce denial-of-service attack vectors. To prevent them from taking the canonical mempool or other alt mempools down with them, a reputation is managed for each. An alt mempool that causes too many invalidations gets throttled. This limits the scope of the attack and lets the bundler continue doing its work for other mempools.
```

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +330,72 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
+    The "opsIncluded" is incremented after this UserOperation is included on-chain (either by this bundler, or another)
+* **[AREP-020]** the alt-mempool becomes THROTTLED based on the [Reputation Calculation](#reputation-calculation) 
+* **[AREP-030]** if such a UserOperation that passed initial validation, later causes a bundle to revert, it is immediately marked as BANNED, by setting its opsIncluded to 0 and opsSeen to 10000
 
 ## Security Considerations
 
+This document describes the security considerations bundlers must take to protect themselves (and the entire mempool network)
+from denial-of-service attacks.
+
 ### Possible Attacks
```
yoavw: @drortirosh Aren't we removing this entire section?  And is there anything else we want to put in "Security considerations" instead?

erc/ERCS/erc-7562.md
```diff
@@ -179,44 +274,49 @@ To help make sense of these params, note that a malicious paymaster can at most
...
     * **[STO-022]**  There is an `initCode` and the `factory` contract is staked.
 * If the entity (`paymaster`, `factory`) is staked, then it is also allowed:
     * **[STO-031]** Access the entity's own storage.
-    * **[STO-032]** Read/Write Access to storage slots that is associated with the entity, in any non-entity contract.
+    * **[STO-032]** Read/Write Access to storage slots that are associated with the entity, in any non-entity contract.
     * **[STO-033]** Read-only access to any storage in non-entity contract.
+
+### Local Rules
+
+These are storage rules that help to protect the bundler from denial of service:
```
yoavw:
```suggestion
Local storage rules protect the bundler against denial of service at the time of bundling. They do not affect mempool propagation and cannot cause a bundler to be marked as a "spammer".

```

erc/ERCS/erc-7562.md
drortirosh: 1. it is not a "single stake": it is a repeated stake on every block, up to the stake time.
2. we could say that in such a case, the staked entity "takes precedent" and gets banned, not the rule itself (that is, immediate banning of alt-mempool is only if can't blame a staked entity, based on the FailOp code)

erc/ERCS/erc-7562.md
drortirosh: I don't think we should have yet another entity level of "rule" within "mempool". also, we don't know the interactions of these rules with each other.
if they can be applied separately, then they are separate mempools..

erc/ERCS/erc-7562.md
drortirosh: moved into a separate rule (EREP-015)

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +329,66 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
-    * `opsAllowed` is a reputation based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
-* **[UREP-010]** `UserOperation` with unstaked sender are only allowed up to `SAME_SENDER_MEMPOOL_COUNT` times in the mempool.
+    * `UnstakedReputation` of an entity determines the maximum number of entries using this entity allowed in the mempool.
+    * `opsAllowed` is a reputation-based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
+* **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
 * **[UREP-020]** For other entities: \
-    `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
-    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for new entity
+  `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
```
shahafn:
```suggestion
opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + inclusionRate * (INCLUSION_RATE_FACTOR + (min(opsIncluded, 10000))
```

erc/ERCS/erc-7562.md
shahafn: I suggested something else. There are three statuses for an entity: ok, throttled, banned. The opsAllowed calculation formula is only meant for unstaked entities that are OK. Once the entity is throttled or banned, the opsAllowed calcuation is much simpler and doesn't require UREP-020/UREP-030.

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +329,66 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
+    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for a new entity
 * **[UREP-030]** If an unstaked entity causes an invalidation of a bundle, its `opsSeen` is set to `1000`, effectively blocking it from inclusion for 24 hours.
```
shahafn, line 355: I suggest we remove UREP-030, and define that status of the entity as BANNED, therefore handling differently than UREP-020, which was meant for entities that aren't banned/throttled, to handle slow start for unstaked entities.

erc/ERCS/erc-7562.md
shahafn: I think that we should simplify these rules: 
We have a definition of status: OK/THROTTLED/BANNED. 
For banned/throttled, we have a requirement of opsSeen vs opsIncluded, with a certain slack. 
UREP-020 is meant only for OK entities opsAllowed calculation. 
Once the entity is throttled, i.e. `max_seen > opsIncluded + THROTTLING_SLACK`, we no longer calculate opsAllowed by UREP-020, but allowing only `THROTTLED_ENTITY_MEMPOOL_COUNT`  and `THROTTLED_ENTITY_BUNDLE_COUNT` by SREP-030. If it keeps misbehaving, it may get banned, which then SREP-020 handles. 
This should hold for both staked and unstaked entities imo.

erc/ERCS/erc-7562.md
yoavw: There may actually be rules that only apply to some of the UserOps in an alt mempool. E.g. a paymaster has two functions that violate different rules. A typical UserOp only uses one of them, but logically it's a mempool for that particular mempool. Throttling one rule may not affect the other. For example, one rule may apply only when deploying a new account using the mempool, which happens rarely. Throttling that rule doesn't need to throttle other uses of that paymaster even if they require an alt mempool rule.

If we maintain this reputation for rules, it's not **in addition** but instead of a whole-mempool reputation.  But anyway I don't have a strong opinion on this.  If you think the above case should throttle the entire alt mempool, it's fine.

erc/ERCS/erc-7562.md
yoavw: 1. I meant that you only have to stake once.  You don't keep staking more because you don't mind being banned for as long as the mempool is.  That's why it's a single stake - you only lock $1000, not $1000*blocks.
2. This would indeed prevent the attack but opens a DoS vector against the staked entity instead. For example, suppose it's a TokenPaymaster that works with any token (and is generally safe), and an alt mempool allows an unsafe token - which gets abused in the context of the TokenPaymaster.  In that case the paymaster had no way to defend itself and shouldn't be banned. So I would say the opposite - entities that get banned in an alt mempool should **NOT** be banned for other mempools for their "crime".

erc/ERCS/erc-7562.md
yoavw: What is the role of INCLUSION_RATE_FACTOR here?  It is not really used as a factor, just adds up to 10 to the total number.

- The function starts at 10 (so banning never happens). I guess it's fine if it only protects against a drained paymaster.
- After one inclusion it jumps to 20.  From there it increases by 1 for every inclusion until 10010.
- The drop rate is linear, not faster than the increase rate.  If a paymaster's balance is drained after it had 10000 inclusions, it starts dropping from 10010 to 10 at a slow rate.  It'll drop back to 10 after 10000 unpaid simulations.  Is that the intended behavior?

erc/ERCS/erc-7562.md
yoavw: @drortirosh I see you didn't add this suggestion to the latest version.  Do you think it's unneeded? I thought it would make sense to explain the rationale for this additional reputation.

erc/ERCS/erc-7562.md
yoavw: @drortirosh I see the whole list is still there, except for the three lines that say it's not exhaustive (which actually makes it worse).  Shouldn't we remove this entire list?  (and if we do keep it, we must add many others as previously discussed, and address other comments I made about the list - but I still think it's best to just remove it).

erc/ERCS/erc-7562.md
yoavw: I'm confused. I don't think we have status defined anywhere. It is just implied by the opsSeen/opsIncluded ratio. So the only way to force a certain status is to manipulate these two numbers.

I thought UREP-020 is only for unstaked entities, which means that these entities are always "OK" unless UREP changes their numbers.  Do you mean that we keep SREP for unstaked entities as well, and use it in addition to UREP?  In that case SREP is no longer the right name for it, but I didn't think staked reputation applies to unstaked entities.

erc/ERCS/erc-7562.md
yoavw: This comment is about SREP, not UREP.  We need a way to mark something as BANNED for a certain duration (e.g. 24 hours).  We don't have statuses, just opsSeen and opsIncluded.  Therefore we need to calculate values for these two numbers, that will keep the entity banned for that duration until the decay function brings it back to THROTTLED and then OK if it behaves well.

erc/ERCS/erc-7562.md
```diff
@@ -230,63 +329,66 @@ This means that a `Paymaster` and `Factory` contracts cannot practically be an "
...
 
-Below is a list of possible attacks that were considered and a reference to the above rule that prevents it.
+## Security Considerations
 
-This list does not attempt to be an exhaustive list of attacks.
-These attacks are examples provided to describe and rationalise the reason for the above rules.
+This document describes the security considerations bundlers must take to protect themselves (and the entire mempool network)
+from denial-of-service attacks.
 
 #### Sample Known (expensive) attack:
```
yoavw: @drortirosh I still think this ENTIRE section (from here till line 446) should be removed.  If we don't remove it then we need to rewrite it, address all the other issues I raised in it, and add all the "trivial" attacks we block.

erc/ERCS/erc-7562.md
drortirosh: setting opsSeen to 10000 means it gets un-banned after ~70 hours, or 3 days.

drortirosh:
already merged

PR #389, opened 2023-12-21
AA-238 Removing obsolete gnosis safe 4337 module

shahafn:
There's already a module supported by the safe team

https://github.com/safe-global/safe-modules/blob/master/4337/contracts/Safe4337Module.sol


drortirosh:
grep gnosis... it still appears in package.json.

shahafn:
> grep gnosis... it still appears in package.json.

Removed

drortirosh (approved):

PR #390, opened 2023-12-31
gas-optimize deposit

drortirosh:
update storage mapping for major use-case (handleOps)
deposit as uint256 makes handling deposit in handleOps slightly cheaper (300gas per op).
probably makes stake() cost a little higher (up to 15000 more, though didn't check its cost) - but that's one-time, even for the paymaster...


contracts/core/EntryPointSimulations.sol
```diff
@@ -170,13 +170,12 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
...
     function depositTo(address account) public override(IStakeManager, StakeManager) payable {
-        uint x;
-        assembly {
-            //some silly code to waste ~200 gas
-            x := exp(mload(0),100)
-        }
-        if (x == 123) {
-            return;
+        unchecked{
+            uint g = gasleft();
```
forshtat, line 174: What's going on here?

drortirosh: Its a code to waste some gas... we must make sure that the simulation contract always takes more gas, otherwise, it can be used to attack the bundler.
The previous "gas waster" didn't pass the tests, but wasn't easily adaptable, so I changed it to waste gas in a loop.
Strangely, the different functions (with/without value, using fallback or "depositTo") have different delta from the real entrypoint code. so I had to make sure it is above for all.

forshtat (approved):

PR #391, opened 2024-01-01
add separate test for paymaster with postOp

drortirosh:
we use our "gas tests" to check the impact of specific version. it is important to to break tests.
This PR make "simple paymaster" simple again, and adds a separate test of "paymaster with postOp"

tests are ordered so that they can be directly diff'ed with older (e.g. v0.6.0) versions


gascalc/4-paymaster-postop.gas.ts
```diff
@@ -0,0 +1,45 @@
+import { parseEther } from 'ethers/lib/utils'
+import { TestPaymasterAcceptAll__factory } from '../typechain'
+import { ethers } from 'hardhat'
+import { GasChecker } from './GasChecker'
+import { Create2Factory } from '../src/Create2Factory'
+import { hexValue } from '@ethersproject/bytes'
+
+const ethersSigner = ethers.provider.getSigner()
+
+context('Paymaster with PostOp', function () {
```
forshtat, line 10: Can't this test be "parametrized" in `2-paymaster.gas.ts` file? How much of it is the same?

shahafn: Yes it seems very "parametizable"

contracts/test/TestPaymasterAcceptAll.sol
```diff
@@ -22,8 +24,12 @@ contract TestPaymasterAcceptAll is BasePaymaster {
     internal virtual override view
     returns (bytes memory context, uint256 validationData) {
         (userOp, userOpHash, maxCost);
-        // return a context, as it is used for EntryPoint gas checking.
-        return ("1", 0);
+        if ( callPostOp) {
```
shahafn: Wouldn't it be better to have a different paymaster where you inherit this paymaster and add on a postOp function, instead of carrying a flag? I know we did similar things elsewhere, but it seems more readable to just have a different one.

gascalc/4-paymaster-postop.gas.ts
drortirosh: maybe. but I did want to keep the table in a format that I can easily compare to previous versions.
for this reason, the "simple paymaster" is completely separate test from "paymaster with postop"

shahafn (approved):

PR #395, opened 2024-01-06
add postOp param (gasPrice)

drortirosh:
saves TokenPaymaster 731 gas (it currently required to pass maxFeePerGas and maxPriorityFee as context params)
add "plain postOp" (which ignores this param) 24 gas.


yoavw (dismissed):
Makes sense, let's do it.

shahafn (dismissed):

contracts/core/BasePaymaster.sol
```diff
@@ -54,10 +54,11 @@ abstract contract BasePaymaster is IPaymaster, Ownable {
     function postOp(
         PostOpMode mode,
         bytes calldata context,
-        uint256 actualGasCost
+        uint256 actualGasCost,
+        uint gasPrice
```
forshtat: This name is highly confusing. The `postOp` method can actually call `tx.gasprice` and developers will be confused with the fact they are not the same thing. We should name it `actualUserOpFeePerGas` or something like that, so that the name "screams" that this is not `tx.gasprice`.
Also, please, use `uint256` for both params - makes no sense to use two different ways in one line.

contracts/samples/LegacyTokenPaymaster.sol
```diff
@@ -102,11 +102,11 @@ contract LegacyTokenPaymaster is BasePaymaster, ERC20 {
      * the user's TX , back to the state it was before the transaction started (before the validatePaymasterUserOp),
      * and the transaction should succeed there.
      */
-    function _postOp(PostOpMode mode, bytes calldata context, uint256 actualGasCost) internal override {
+    function _postOp(PostOpMode mode, bytes calldata context, uint256 actualGasCost, uint gasPrice) internal override {
```
forshtat: Unrelated - why do we keep this old code around? Can we remove it?

contracts/samples/TokenPaymaster.sol
```diff
@@ -149,16 +149,13 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
...
         unchecked {
             uint256 priceMarkup = tokenPaymasterConfig.priceMarkup;
             (
                 uint256 preCharge,
-                uint256 maxFeePerGas,
-                uint256 maxPriorityFeePerGas,
                 address userOpSender
-            ) = abi.decode(context, (uint256, uint256, uint256, address));
-            uint256 gasPrice = getGasPrice(maxFeePerGas, maxPriorityFeePerGas);
+            ) = abi.decode(context, (uint256, address));
```
forshtat, line 162: This means the entire `getGasPrice` function is no longer needed - please remove.

forshtat (changes requested):

shahafn (approved):

forshtat (approved):

PR #396, opened 2024-01-07
mark assembly as "memory-safe"

drortirosh:
All our assembly code is memory-safe, so we flag it as such. 
it helps the optimizer in general.

Solidity [defines memory safety here](https://docs.soliditylang.org/en/latest/assembly.html#memory-safety).
1. We use scratch memory 0-63.
2. We save `bytes memory` pointer as `bytes32`, and "cast" that pointer back to `bytes memory` to read it. We take advantage of the fact solidity never re-use previously defined memory
3. The memory safety rules allow using "scratch area" starting at the current "free ptr" (without updating the pointer)
    We use solidity "abi.encode" to manipulate this scratch area. So we save the "free pointer" at 0x40, call `abi.encodeCall` and use the resulting buffer to make a `call`. Just after this call, we restore the free pointer. The `bytes memory` items used for this call are not used again.


contracts/core/EntryPoint.sol
```diff
@@ -27,7 +27,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
 
     using UserOperationLib for UserOperation;
 
-    SenderCreator private senderCreator = new SenderCreator();
+    SenderCreator private immutable senderCreator = new SenderCreator();
```
drortirosh: unrelated to memory safety, but nonetheless small optimization for contract creation

contracts/core/EntryPoint.sol
```diff
@@ -100,14 +100,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             innerCall = abi.encodeCall(this.innerHandleOp, (callData, opInfo, context));
         }
         bool success;
-        assembly {
+        assembly ("memory-safe") {
             success := call(gas(), address(), 0, add(innerCall, 0x20), mload(innerCall), 0, 32)
             collected := mload(0)
             mstore(0x40, saveFreePtr)
```
drortirosh: restoring the "free ptr" is considered safe in this case, since we don't use any memory allocated since we saved this pointer: we allocated `executeUserOp` and then `innerCall`, both used by "call" above, and unused again.

contracts/core/EntryPoint.sol
forshtat: Heads up, I remember Solidity really struggled with `immutable ... = new ();` kind of syntax, but I don't remember what the actual problem was.

contracts/core/EntryPoint.sol
forshtat: Are you sure this is the correct interpretation of the rules? If I understand what they wrote about `"memory-safe"` correctly, they are saying that now Solidity compiler can put variables to the memory, so you can't mess with memory in `assembly`. And here we are, resetting a free memory pointer. 

Also, we are writing to the `0..32` range a lot here. I guess this is fine but if `innerCall` in the future needs to return 64 bytes we need to change that, right?

contracts/core/EntryPointSimulations.sol
```diff
@@ -172,10 +172,11 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
     // empiric test showed that without this wrapper, simulation depositTo costs less..
     function depositTo(address account) public override(IStakeManager, StakeManager) payable {
         unchecked{
-            uint g = gasleft();
-            uint x = block.number;
-            while (g - gasleft() < 120) {
-                x = x + block.number;
+        // silly code, to waste some gas to make sure depositTo is always little more
```
forshtat, line 191: Did you intend to rewrite this code again?

contracts/core/EntryPoint.sol
```diff
@@ -770,7 +770,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
     function getMemoryBytesFromOffset(
         uint256 offset
     ) internal pure returns (bytes memory data) {
-        assembly {
+        assembly ("memory-safe") {
```
forshtat, line 762: Like, this function is not actually safe, right? It depends on passing it correct `offset`, but this is not immediately obvious, don't you think?
I can see that currently `getMemoryBytesFromOffset` is only used with `context` which is never overwritten, but I think at least the comments and the name of the function should indicate you can shoot yourself in the foot if you use it.
What do you think?
Screenshot 2024-01-08 at 16 59 34

contracts/core/EntryPoint.sol
drortirosh: removed SenderCreator from this commit...

contracts/core/EntryPoint.sol
drortirosh: 0..32 (and even 0..64) is a "scratch area", and code is allowed freely to write to it.
Yes, this code takes advantage from the fact innerCall return less than 64 bytes. above that you need to allocate buffer (or use multiple returndatacopy calls)

contracts/core/EntryPointSimulations.sol
drortirosh: yes, its simpler..

yoavw (approved):

PR #403, opened 2024-01-12
Fixing reputation rules

shahafn:
Following the discussion on #388 @yoavw @drortirosh @forshtat please review my suggestions here


erc/ERCS/erc-7562.md
```diff
@@ -317,19 +317,24 @@ These are storage rules that help to protect the bundler from denial of service:
...
+
+* **[GREP-010]** A `BANNED` address is not allowed into the mempool.\
   Also, all existing `UserOperations` referencing this address are removed from the mempool.
-* **[SREP-030]** A `THROTTLED` address is limited to:
+* **[GREP-020]** A `THROTTLED` address is limited to:
     * `THROTTLED_ENTITY_MEMPOOL_COUNT` entries in the mempool.
     * `THROTTLED_ENTITY_BUNDLE_COUNT` `UserOperations` in a bundle.
     * Can remain in the mempool only for `THROTTLED_ENTITY_LIVE_BLOCKS`.
-* **[SREP-040]** An `OK` staked entity is unlimited by the reputation rule.
+* **[GREP-040]** If an entity fails the second validation or fails bundle creation, its `opsSeen` set to `10000`, and `opsIncluded` to zero, causing it to be `BANNED`.
```
drortirosh: User constant instead of 10000

erc/ERCS/erc-7562.md
```diff
@@ -348,11 +353,11 @@ These are storage rules that help to protect the bundler from denial of service:
     * **`opsSeen`, `opsIncluded`, and reputation calculation** are defined above.
     * `UnstakedReputation` of an entity determines the maximum number of entries using this entity allowed in the mempool.
     * `opsAllowed` is a reputation-based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
```
drortirosh, line 357: I think need a better name for opsAllowed, eg opsAllowedInMempool (though mempool is confusing here wrt alt-mempools

shahafn: Babe == name I assume?

erc/ERCS/erc-7562.md
shahafn: Maybe this number should be derived from the other constants?

yoavw: What's GREP?
and we are **still** missing a calculation for the correct opsSeen and opsIncluded for a specific duration of banning.  We only know that the current one is wrong.

Suppose we want to ban for 24 hours, we need to calculate opsSeen and opsIncluded such that after 23 hourly decalys it'll still be in BANNED category, but after 24 hourly decays it'll enter THROTTLED category.

yoavw: and why was SREP-040 removed?  Is it no longer the case that we don't limit staked entities while they are OK?

shahafn: It was not removed, just renumbered

erc/ERCS/erc-7562.md
```diff
@@ -317,19 +317,24 @@ These are storage rules that help to protect the bundler from denial of service:
   This means that `Paymaster` and `Factory` contracts cannot practically be an "account" contract as well.
 * **[STO-041]** `UserOperation` may not use associated storage (of either its account or from staked entity) in a contract that is a "sender" of another UserOperation in the mempool.
 
-### Staked Entities Reputation Rules
 
-* **[SREP-010]** The "canonical mempool" defines a staked entity if it has `MIN_STAKE_VALUE` and unstake delay of `MIN_UNSTAKE_DELAY`
-* **[SREP-020]** A `BANNED` address is not allowed into the mempool.\
+### General Reputation rules
+The following reputation rules apply for both staked and unstaked entities.
```
yoavw: Define which entities this applies to:
- All staked entities
- Unstaked paymasters (but not other unstaked entities)

erc/ERCS/erc-7562.md
yoavw:
```suggestion
* **[GREP-040]** If an entity fails bundle creation after passing 2nd validation, its `opsSeen` set to `10000`, and `opsIncluded` to zero, causing it to be `BANNED`.
```

erc/ERCS/erc-7562.md
```diff
@@ -348,11 +353,11 @@ These are storage rules that help to protect the bundler from denial of service:
...
     * `UnstakedReputation` of an entity determines the maximum number of entries using this entity allowed in the mempool.
     * `opsAllowed` is a reputation-based calculation for an unstaked entity, representing how many `UserOperations` it is allowed to have in the mempool.
-* **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
-* **[UREP-020]** For other entities: \
-  `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
-    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for a new entity
-* **[UREP-030]** If an unstaked entity causes an invalidation of a bundle, its `opsSeen` is set to `1000`, effectively blocking it from inclusion for 24 hours.
+  * Rules:
+    * **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
+    * **[UREP-020]** For other unstaked `OK` entities: \
```
yoavw: Maybe elaborate what `OK` means.  This rule only applies when the entity is not THROTTLED or BANNED by other rules.

erc/ERCS/erc-7562.md
```diff
@@ -348,11 +353,11 @@ These are storage rules that help to protect the bundler from denial of service:
...
+      `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + inclusionRate * (INCLUSION_RATE_FACTOR + min(opsIncluded, 10000))`.
```
yoavw: 1. Consider whether INCLUSION_RATE_FACTOR is needed at all.  It just starts the slow-start at 20 instead of 10 and does nothing else.
2. Make both 10000's constants.  We have them in two unrelated places.  One is the max ops allowed for an unstaked entity, the other (which appears in several rules) is used to trigger banning by setting opsSeen=10000. For the latter, specify the number of hours it takes to unban with this value.

erc/ERCS/erc-7562.md
yoavw: This rule was also wrong before this PR as an SREP.  We only insta-BAN an entity that causes bundle invalidation after passing 2nd simulation, because this causes more work and there's no legitimate reason for this to happen. Just failing 2nd validation is legitimate in small numbers (any case where a state change invalidated a UserOp).  If we do this for every invalidation, the whole reputation mechanism is unused.

erc/ERCS/erc-7562.md
shahafn: Fixed

erc/ERCS/erc-7562.md
shahafn: Yes, fixed

erc/ERCS/erc-7562.md
shahafn: Fixed

erc/ERCS/erc-7562.md
shahafn: Fixed and fixed

erc/ERCS/erc-7562.md
```diff
@@ -320,14 +322,15 @@ These are storage rules that help to protect the bundler from denial of service:
 
 ### General Reputation rules
 The following reputation rules apply for both staked and unstaked entities.
+All rules apply to all entities unless specified otherwise.
```
yoavw: Why? We don't want to track opsSeen/opsIncluded for unstaked accounts, and IMHO also not for unstaked factories unless we find a reason to.

erc/ERCS/erc-7562.md
```diff
@@ -342,17 +350,17 @@ These are storage rules that help to protect the bundler from denial of service:
...
-* **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
-* **[UREP-020]** For other entities: \
-  `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + (inclusionRate * INCLUSION_RATE_FACTOR) + (min(opsIncluded, 10000)`.
-    * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for a new entity
-* **[UREP-030]** If an unstaked entity causes an invalidation of a bundle, its `opsSeen` is set to `1000`, effectively blocking it from inclusion for 24 hours.
+  * Rules:
+    * **[UREP-010]** An unstaked sender is only allowed to have `SAME_SENDER_MEMPOOL_COUNT` `UserOperation`s  in the mempool. A staked sender is only limited by the SREP rules.
+    * **[UREP-020]** For an unstaked paymaster only that is not throttled/banned: \
+      `opsAllowed = SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT + inclusionRate * min(opsIncluded, MAX_OPS_ALLOWED_UNSTAKED_ENTITY)`.
+        * This is a default of `SAME_UNSTAKED_ENTITY_MEMPOOL_COUNT` for a new paymaster
```
drortirosh: Better add 
```
    * **[UREP-030]** unstaked factory has no limitation on mempool or bundle inclusion.
```
Since it is not clear from the text above.

(it is also implied that opsSeen/opsIncluded are required only for unstaked paymasters, not other unstaked entities)

yoavw: The problem is line 325, which says `All rules apply to all entities unless specified otherwise.`

I think it should only include staked entities and unstaked paymasters.  No reason to keep these numbers for unstaked factories and accounts, since they don't use a shared resource (like paymaster's deposit).

erc/ERCS/erc-7562.md
shahafn: I think that this is an OK name, given the context and the definition here..

erc/ERCS/erc-7562.md
shahafn: Fixed

erc/ERCS/erc-7562.md
```diff
@@ -317,19 +319,25 @@ Local storage rules protect the bundler against denial of service at the time of
...
+* **[GREP-040]** If an entity fails the bundle creation after passing second validation, its `opsSeen` set to `BAN_OPS_SEEN_PENALTY`, and `opsIncluded` to zero, causing it to be `BANNED`.
+
+### Staked Entities Reputation Rules
+
+* **[SREP-010]** The "canonical mempool" defines a staked entity if it has `MIN_STAKE_VALUE` and unstake delay of `MIN_UNSTAKE_DELAY`
+* **[SREP-020]** An `OK` staked entity is unlimited by the reputation rule.
     * Allowed in unlimited numbers in the mempool.
     * Allowed in unlimited numbers in a bundle.
-* **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` set to `10000`, and `opsIncluded` to zero, causing it to be `BANNED`.
+* **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` is incremented by `BAN_OPS_SEEN_PENALTY`, causing it to be `BANNED`.
```
yoavw:
```suggestion
* **[SREP-050]** If a staked entity fails the second validation or fails bundle creation, its `opsSeen` is set to `BAN_OPS_SEEN_PENALTY` and its `opsIncluded` is set to zero, causing it to be `BANNED`.
```

Why did this change to "increment opsSeen by X" after we already fixed it to "set opsSeen to X and opsIncluded to zero"? The behavior in this new version depends heavily on the current values, and may result in not banning for the expected duration (or at all if opsIncluded is sufficiently high).

yoavw: Damn!  It removed my 2nd suggestion.  Please do not accept until I re-add it

yoavw:
```suggestion
```

Suggesting to remove this rule as it seems to be covered by GREP-040.  The rule also has the wrong condition but I won't correct it if we can just remove it.

drortirosh (dismissed):

yoavw (approved):

PR #406, opened 2024-01-17
AA-276 M-01 export separately account and paymaster validationData, remove intersectTimeRange

drortirosh:
remove the simulateValidation code to intersect the paymaster and account time-ranges (and signature validation).
this calculation should be done off-chain by the bundler. The simulation functions now return the "raw" validationData, which is parsed into
- aggregator (or SIG_VALIDATION_FAILED),
- validUntil
- validAfter


contracts/core/EntryPointSimulations.sol
```diff
@@ -129,8 +119,8 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
         return ExecutionResult(
             opInfo.preOpGas,
             paid,
-            data.validAfter,
-            data.validUntil,
+            validationData,
```
forshtat, line 122: Is it okay that the returned structure does not explicitly say if the call will be accepted or not, and requires some client logic to figure that out?

drortirosh: the method call is for the bundler only. 
(in the past, we did partial work to parse it in solidity, but it was partial anyway and required the bundler to check it - then why bother to do part of the parsing in solidity, and the rest in the bundler's native code? it can be all native.

contracts/core/EntryPointSimulations.sol
```diff
@@ -68,23 +68,17 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
...
+            paymasterValidationData,
             getMemoryBytesFromOffset(outOpInfo.contextOffset)
         );
 
         AggregatorStakeInfo memory aggregatorInfo = NOT_AGGREGATED;
-        if (aggregator != address(0) && aggregator != address(1)) {
+        if (uint160(aggregator) > SIG_VALIDATION_FAILED) {
```
shahafn: Why?

erc/ERCS/erc-4337.md
```diff
@@ -207,13 +207,14 @@ this bundler is supposed to track the `key` and `sequence` pair of the UserOpera
 #### Using signature aggregators
 
 An account signifies it uses signature aggregation returning its address from `validateUserOp`.
-During `simulateValidation`, this aggregator is returned (in the `ValidationResultWithAggregator`)
+During `simulateValidation`, this aggregator is returned as part of the `accountValidationData`.
```
shahafn: Not sure this is correct

erc/ERCS/erc-4337.md
```diff
@@ -379,9 +379,11 @@ The simulated call performs the full validation, by calling:
 2. `account.validateUserOp`.
 3. if specified a paymaster: `paymaster.validatePaymasterUserOp`.
 
-Either `validateUserOp` or `validatePaymasterUserOp` may return a "validAfter" and "validUntil" timestamps, which is the time-range that this UserOperation is valid on-chain.
-The simulateValidation call returns this range.
-A node MAY drop a UserOperation if it expires too soon (e.g. wouldn't make it to the next block)
+The simulateValidation should validate the return value (validationData) returned by the account's `validateUserOp` and paymaster's `validatePaymasterUserOp`.
```
shahafn: You describe implementation details of the EntryPointSimulations. It seems out of place

forshtat (dismissed):

forshtat (dismissed):

shahafn (approved):

forshtat (approved):

PR #410, opened 2024-01-22
fix: allow _senderCreator to be overridden

kopy-kat:
This is useful when overriding the virtual function `initSenderCreator`. One usecase is to pass a different address then `address(this)` in order to calculate the address of the `SenderCreator`, for example when using Foundry's `etch` to move the EntryPoint to a different address.

### DIFF
--- a/contracts/core/EntryPointSimulations.sol
+++ b/contracts/core/EntryPointSimulations.sol
@@ -16,7 +16,7 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
     // solhint-disable-next-line var-name-mixedcase
     AggregatorStakeInfo private NOT_AGGREGATED = AggregatorStakeInfo(address(0), StakeInfo(0, 0));
 
-    SenderCreator private _senderCreator;
+    SenderCreator _senderCreator;
 
     function initSenderCreator() internal virtual {
         //this is the address of the first contract created with CREATE by this address.


drortirosh:
I'm not sure how this change helps you.
1. you can't really return a different address, as this has to match the actual address used by entryPoint. you can't use a constructor parameter, since the deployed code doesn't get installed using a constructor
2. you can always override initSenderCreator, and ignore this field.
3. one possible way we thought of using: return a hard-coded constant (e.g `return SenderCreator(address(0xdead1234));`, and then do a search and replace on the deployedByteCode before using it for state-override.
   we decided not to use this method since it requires a different implementation depending on your evm support in your bundler language.

kopy-kat:
My problem is the following:
1. ep is deployed and SenderCreator deployed to `address(..., address(this), ...)`
2. ep is etched to new address, SenderCreator is not deployed again
3. instead of calculating SenderCreator address to be `address(..., address(new_this), ...)` I want it to be `address(..., address(original_this), ...)` in the etched ep

Imo this change would be the easiest way to achieve this. Happy to send the code for this and also open to different solutions for within foundry. Point 2 above doesn't really help because the EP uses `_senderCreator`.

drortirosh:
> ep is etched to new address,
Etch changes the code at address.
IIUC, you need it for simulation (as a poor man state override)
The idea with simulation is keep the UserOp blind to the fact it is simulation and not execution. This is why we restrict opcodes
This means that even the msg.sender (which is an allowed opcode) at the creation code must be the same in simulation and in execution, hence must use the same sender creator.

kopy-kat:
Yeah, since the EntryPointSimulations inherits from Entrypoint, we use the same etched contract as the EntryPoint to handle UserOps too. I guess another solution would be to etch the code of SenderCreator to `address(…,address(new_ep),…)`

kopy-kat:
For context, this is within an integration testing library for 4337

PR #415, opened 2024-01-25
AA-287 L-07 update pragmas

drortirosh:


contracts/interfaces/INonceManager.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0
-pragma solidity ^0.8.12;
+pragma solidity >=0.8.0;
```
forshtat: Why is this file the only one in the `/interfaces` folder to use `>=0.8.0` instead of `>=0.7.5`?

forshtat (changes requested):

forshtat (approved):

PR #428, opened 2024-01-29
AA-279 M-04 Inconsistent Price Precision

shahafn:
Fixing priceUpdateThreshold


contracts/samples/utils/OracleHelper.sol
```diff
@@ -39,7 +39,7 @@ abstract contract OracleHelper {
         /// @notice 'false' if price is dollars-per-ether, 'true' if price is ether-per-dollar
         bool nativeOracleReverse;
 
-        /// @notice The price update threshold percentage that triggers a price update (1e6 = 100%)
+        /// @notice The price update threshold percentage from PRICE_DENOMINATOR that triggers a price update (1e3 = 100%)
```
forshtat: I think `1e3` is better written as `1000`

contracts/samples/utils/OracleHelper.sol
```diff
@@ -100,11 +100,10 @@ abstract contract OracleHelper {
             oracleHelperConfig.tokenOracleReverse,
             oracleHelperConfig.nativeOracleReverse
         );
-        uint256 priceNewByOld = price * PRICE_DENOMINATOR / _cachedPrice;
-
+        uint256 priceRatioNewByOld = price * PRICE_DENOMINATOR / _cachedPrice;
         bool updateRequired = force ||
-            priceNewByOld > PRICE_DENOMINATOR + priceUpdateThreshold ||
-            priceNewByOld < PRICE_DENOMINATOR - priceUpdateThreshold;
+            priceRatioNewByOld > PRICE_DENOMINATOR * (1 + priceUpdateThreshold / 1e3) ||
```
forshtat: I think you will just get `priceUpdateThreshold / 1e3 == 0` for any `priceUpdateThreshold < 1e3`, am I missing something?

forshtat (changes requested):

contracts/samples/utils/OracleHelper.sol
shahafn: Fixed

contracts/samples/utils/OracleHelper.sol
shahafn: Indeed. There was an issue before as well. I think I fixed it.

contracts/samples/TokenPaymaster.sol
```diff
@@ -26,7 +26,7 @@ import "./utils/OracleHelper.sol";
 contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
 
     struct TokenPaymasterConfig {
-        /// @notice The price markup percentage applied to the token price (1e6 = 100%)
+        /// @notice The price markup percentage applied to the token price (1e26 = 100%). Ranges from 1e26 to 2e26
```
drortirosh, line 29: if 1e26 is 100%, then the range is 100% .. 200% ?

forshtat: I think yes. This is the markup, so using 100% of the price is 0% markup and using 200% of the price means 100% markup, does that make sense?

forshtat (dismissed):

drortirosh (approved):

PR #430, opened 2024-01-29
Pack gas fees

drortirosh:
pack maxFeePerGas and maxPriorityFee into a single field in `PackedUserOperation`


contracts/samples/TokenPaymaster.sol
```diff
@@ -124,7 +124,8 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
             require(dataLength == 0 || dataLength == 32,
                 "TPM: invalid data length"
             );
-            uint256 preChargeNative = requiredPreFund + (tokenPaymasterConfig.refundPostopCost * userOp.maxFeePerGas);
+            (uint256 maxFeePerGas,) = UserOperationLib.unpackAccountGasLimits(userOp.gasFees);
```
forshtat: I think it is a problem using a function named `unpackAccountGasLimits` for unpacking fees per gas - please rename it to something more generic, maybe?

UPD: In fact, unclear meaning of returned values is the reason fort the failed test, as the decoding return values are reversed:
```
(maxPriorityFeePerGas, maxFeePerGas) = UserOperationLib.unpackAccountGasLimits(userOp.gasFees);
````

Comment (drortirosh) [reply contracts/samples/TokenPaymaster.sol]: thanks.
any suggested naming?
probably: `unpackPrioAndFeePerGas() returns (maxPrio, maxFeePerGas)`

Comment (shahafn) [reply contracts/samples/TokenPaymaster.sol]: Are you adding a function or renaming existing one? If you add, maybe unpackGasFees?

File: contracts/core/UserOperationLib.sol [hunk 2]
```diff
@@ -61,23 +60,61 @@ library UserOperationLib {
...
+    }
+
+    function unpackMaxPriorityFeePerGas(PackedUserOperation calldata userOp)
+    internal pure returns (uint256) {
+        return unpackHigh128(userOp.accountGasLimits);
+    }
+
+    function unpackMaxFeePerGas(PackedUserOperation calldata userOp)
+    internal pure returns (uint256) {
+        return unpackLow128(userOp.accountGasLimits);
```
shahafn:
```suggestion
        return unpackLow128(userOp.gasFees);
```

contracts/core/UserOperationLib.sol
```diff
@@ -61,23 +60,61 @@ library UserOperationLib {
...
+    }
+
+    // unpack just the low 128-bits from a packed value
+    function unpackLow128(bytes32 packed) internal pure returns (uint256) {
+        return uint128(uint256(packed));
+    }
+
+    function unpackMaxPriorityFeePerGas(PackedUserOperation calldata userOp)
+    internal pure returns (uint256) {
+        return unpackHigh128(userOp.accountGasLimits);
```
shahafn:
```suggestion
        return unpackHigh128(userOp.gasFees);
```

shahafn (approved):

PR #413, opened 2024-01-25
AA-285 L05 - fix comments.

drortirosh:

PR #440, opened 2024-02-04
L-05 fix more comments

drortirosh:


yoavw (approved):
Confirmed that this, along with [PR #413](https://github.com/eth-infinitism/account-abstraction/pull/413), addressed L-05.

PR #441, opened 2024-02-05
No revert on no prefund

drortirosh:


jayden-sudo (approved):

contracts/core/EntryPoint.sol
```diff
@@ -145,6 +148,23 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
         }
     }
 
+    function emitPrefundTooLow(uint256 actualGas, UserOpInfo memory opInfo) internal {
+        emit UserOperationPrefundTooLow(
+            opInfo.userOpHash,
+            opInfo.mUserOp.sender,
+            opInfo.mUserOp.nonce
+        );
```
xinweb4: Account or paymaster does not need to pay the gas?

contracts/core/EntryPoint.sol
```diff
@@ -698,7 +715,15 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
+                    return opInfo.prefund;
+                } else {
+                    assembly ("memory-safe") {
+                        mstore(0, INNER_REVERT_LOW_PREFUND)
+                        revert(0, 32)
+                    }
+                }
             }
             uint256 refund = opInfo.prefund - actualGasCost;
             _incrementDeposit(refundAddress, refund);
```
xinweb4: this code about paying gas and emitting event can be moved to '_executeUserOp'

contracts/core/EntryPoint.sol
drortirosh: they do: they pay the entire "prefund".

contracts/core/EntryPoint.sol
drortirosh: the entire "_postExecution" is about event and refund

contracts/core/EntryPoint.sol
```diff
@@ -121,9 +122,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             }
             // handleOps was called with gas limit too low. abort entire bundle.
             if (innerRevertCode == INNER_OUT_OF_GAS) {
-                //report paymaster, since if it is not deliberately caused by the bundler,
-                // it must be a revert caused by paymaster.
+                //can only be caused by bundler (leaving not enough gas for inner call)
                 revert FailedOp(opIndex, "AA95 out of gas");
+            } else if (innerRevertCode == INNER_REVERT_LOW_PREFUND) {
+                uint256 actualGas1 = preGas - gasleft() + opInfo.preOpGas;
```
shahafn: Please rename actualGas1

contracts/core/EntryPoint.sol
```diff
@@ -121,9 +122,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
+                uint256 prefund = opInfo.prefund;
```
forshtat: I think this optimisation is not needed, this branch is pretty exotic, don't you think?

contracts/core/EntryPoint.sol
```diff
@@ -121,9 +122,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
+                emitPrefundTooLow(opInfo);
+                emitUserOperationEvent(opInfo, false, prefund, actualGas1);
```
forshtat: Also naming, but I think having `actualGasCost = prefund` may be confusing as it does not have any relation to `actualGas1`

contracts/core/EntryPoint.sol
```diff
@@ -144,6 +149,26 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
+            opInfo.mUserOp.paymaster,
+            opInfo.mUserOp.nonce,
+            success,
+            actualGasCost,
+            actualGas
+        );
+    }
+
+    function emitPrefundTooLow(UserOpInfo memory opInfo) internal virtual {
+        emit UserOperationPrefundTooLow(
```
forshtat, line 166: Why not emit the details right here?

forshtat (dismissed):

contracts/core/EntryPoint.sol
```diff
@@ -121,9 +122,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
...
-                //report paymaster, since if it is not deliberately caused by the bundler,
-                // it must be a revert caused by paymaster.
+                //can only be caused by bundler (leaving not enough gas for inner call)
                 revert FailedOp(opIndex, "AA95 out of gas");
+            } else if (innerRevertCode == INNER_REVERT_LOW_PREFUND) {
+                uint256 actualGas1 = preGas - gasleft() + opInfo.preOpGas;
+                uint256 prefund = opInfo.prefund;
+                emitPrefundTooLow(opInfo);
+                emitUserOperationEvent(opInfo, false, prefund, actualGas1);
+                return prefund;
```
yoavw: 1. I think you mentioned that you moved the `return` to avoid a solidity inefficiency.  Did you push that change?
2. This `return` prevents the call to `_postExecution`, and since `inner` already reverted with `INNER_REVERT_LOW_PREFUND`, it means that Paymaster's `postOp` won't be called in this case.  Until now we guaranteed that it is always called. What are the implications of removing this guarantee?

contracts/core/EntryPoint.sol
drortirosh: it is the effective gas cost. its true that if you calculate the gas price (gascost/gasused) it'll be little lower than expected, but it is the amount we paid for it.

contracts/core/EntryPoint.sol
drortirosh: if inner reverts on INNER_REVERT_LOW_PREFUND, it means that postOp didn't revert (or didn't exist at all), and the flow reaches emitPrefundTooLow at line 132

If postOp does revert, it reaches "emit PostOpRevertReason at line 135, and if gas is not enough, it is handled in _postExecution at line 724

contracts/core/EntryPoint.sol
drortirosh: it would be very nice to add the actualGasCost and prefund - but note that we revert inside innerCall, catch the revert reason and then submit the event. it is possible to transfer the actualGasCost through this revert, but it adds complexity.

contracts/core/EntryPoint.sol
yoavw: The inner postOp (L695) gets reverted by the subsequent INNER_REVERT_LOW_PREFUND revert in L728, so it hasn't happened even though it was executed successfully.

The outer postOp doesn't happen because the `return` in L132 prevents the `_postExecution` in L143 from being reached.

What flow can lead to a successful postOp (in a context that doesn't get reverted later), when a UserOp triggers INNER_REVERT_LOW_PREFUND?

drortirosh: if INNER_REVERT_LOW_PREFUND is triggered, then both account and paymaster postop are reverted. we can't revert only one of them, which means the paymaster fully paid the prefund.

yoavw: I know. It just means that paymaster devs need to be aware not to count on postOp getting called for post-processing.  For things like TokenPaymaster it doesn't matter since it already precharged during validation.  But for paymasters that expect a certain outcome, e.g. voting, it means they need to change the state in validation (e.g. ban the account by default) and then change it back in postOp.  It's a better pattern in any case, but now it becomes even more important.

forshtat (approved):

5twelve (approved):

PR #445, opened 2024-02-08
Fix missing @openzeppelin/contracts dependecy in contracts package, update its imports in the entrypoint contract

mmv08:
Currently, if you use the `@account-abstraction/contracts` package and try to compile the contracts, it will fail if you don't have OpenZeppelin contracts in your project's dependency tree. The reason is a missing dependency in the `@account-abstraction/contracts`. 

This PR fixes that by adding the dependency. It also updates the named import of the ERC165 contract from OpenZeppelin since gnosis contracts were removed in https://github.com/eth-infinitism/account-abstraction/pull/389


drortirosh:
To merge this issue, you need to run `yarn gas-calc`, to update the gas usage (almost any source code change alters the gas usage, even if by 10 gas..), and check in the updated `reports/gas-checker.txt`
(you can either run local `geth`, or let the script run it for you - but you then you need `docker` installed)

mmv08:
> To merge this issue, you need to run `yarn gas-calc`, to update the gas usage (almost any source code change alters the gas usage, even if by 10 gas..), and check in the updated `reports/gas-checker.txt` (you can either run local `geth`, or let the script run it for you - but you then you need `docker` installed)

done sir 🙏

drortirosh (approved):

PR #447, opened 2024-02-08
remove AA40 - global verification gas check

drortirosh:
remove AA40, 
instead add separate gas check for paymaster and for factory/account, so error can be attributed to failed entity.


shahafn (dismissed):

forshtat (dismissed):

contracts/core/EntryPoint.sol
```diff
@@ -506,8 +513,9 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
                 revert FailedOp(opIndex, "AA31 paymaster deposit too low");
             }
             paymasterInfo.deposit = deposit - requiredPreFund;
+            uint256 pmVerificationGasLimit = mUserOp.paymasterVerificationGasLimit;
```
yoavw, line 511: I'm curious why this variable makes things more efficient than accessing `mUserOp.paymasterVerificationGasLimit` directly below.

contracts/core/EntryPoint.sol
```diff
@@ -518,6 +526,9 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             } catch {
                 revert FailedOpWithRevert(opIndex, "AA33 reverted", Exec.getReturnData(REVERT_REASON_MAX_LEN));
             }
+            if (preGas - gasleft() > pmVerificationGasLimit) {
+                revert FailedOp(opIndex, "AA36 over verificationGasLimit");
```
yoavw: The revert message is identical to the account's case (except for AA36 vs. AA26).  Shouldn't it say "over paymasterVerificationGasLimit"?

contracts/core/EntryPoint.sol
```diff
@@ -628,11 +641,6 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             );
         }
         unchecked {
-            uint256 gasUsed = preGas - gasleft();
-
-            if (mUserOp.verificationGasLimit + mUserOp.paymasterVerificationGasLimit < gasUsed) {
-                revert FailedOp(opIndex, "AA40 over verificationGasLimit");
```
yoavw, line 634: We removed this since we now check each of the limits separately.  Is it guaranteed that this can no longer happen even with EntryPoint overhead?  I assume it is, just checking.

drortirosh: you're right. we didn't cover some validations. Now account validation gas limit would cover also nonce check and copy

contracts/core/EntryPoint.sol
drortirosh: I use it twice in this method. so it saves (a little..) to keep the memory var on the stack.

yoavw (dismissed):

contracts/core/EntryPoint.sol
```diff
@@ -611,13 +618,20 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
             opIndex,
             userOp,
             outOpInfo,
-            requiredPreFund
+            requiredPreFund,
+            verificationGasLimit
         );
 
         if (!_validateAndUpdateNonce(mUserOp.sender, mUserOp.nonce)) {
             revert FailedOp(opIndex, "AA25 invalid account nonce");
```
xinweb4, line 626: Just a minor issue, why not incorporate the Nonce verification into _validateAccountPrepayment? It seems that doing so would make the responsibilities of each method clearer.

drortirosh: I think the method names declare their responsibilities: 
- _validateAccountPrepayment - performs calls validateUserOp, and validates its response (the payment). 
- _validateAndUpdateNonce - performs nonce validation
(yes, we could wrap them both in "account validation), which does nothing but call them both)

shahafn (approved):

PR #453, opened 2024-02-22
Use the `returndatasize` opcode instead of caching the return size in a variable

mmv08:
Per @nlordell's suggestion, the `returndatasize` opcode is a single byte opcode and costs just 2 gas. Its use is more efficient than creating a stack variable or pushing a one-byte constant. 

The gas report proves it's more efficient.


drortirosh:
Thanks for the suggestion.
But please take a look at the gas calculations (the failed "gas-checks" test, or you can run it locally with `yarn gas-calc`)
your fix costs an extra 36 gas for a single UserOp bundle, and an extra 336 gas for a 10-user bundle.
(I can't say why, because looking at your suggestion, it does seem to save gas, not to waste)

mmv08:
> Thanks for the suggestion. But please take a look at the gas calculations (the failed "gas-checks" test, or you can run it locally with `yarn gas-calc`) your fix costs an extra 36 gas for a single UserOp bundle, and an extra 336 gas for a 10-user bundle. (I can't say why, because looking at your suggestion, it does seem to save gas, not to waste)

I did run it locally and pushed the updated calculations as well. For some reason, if I run the CI job locally, it passes. I'm not sure why the CI run reports higher gas.

mmv08:
Ok, we found it - the command hasn't recompiled the code, and it was benchmarking the older version.

PR #454, opened 2024-02-24
fix: debug_bundler_dumpReputation status type

arddluma:
Changed debug_bundler_dumpReputation `status` to `number` instead of `string` as per [spec-tests](https://github.com/eth-infinitism/bundler-spec-tests/blob/master/tests/single/reputation/test_reputation.py#L20-L23)


drortirosh:
please see the [spec](https://github.com/eth-infinitism/bundler-spec/blob/update-actions-versions/src/schemas/bundle.yaml#L38), where we explicitly defined it as enum, not number

PR #459, opened 2024-03-10
fix: token paymaster decimal bug

a00012025:
# Summary

When I tried the latest `TokenPaymaster` in sample, I noticed that it only considers the case of token's decimal being 18, which will cause amount calculation error. This PR fixes this issue by querying the token's decimals when calculating the token amount to transfer to/from paymaster

### DIFF
--- a/contracts/samples/TokenPaymaster.sol
+++ b/contracts/samples/TokenPaymaster.sol
@@ -139,7 +139,12 @@ contract TokenPaymaster is BasePaymaster, UniswapHelper, OracleHelper {
                     cachedPriceWithMarkup = clientSuppliedPrice;
                 }
             }
-            uint256 tokenAmount = weiToToken(preChargeNative, cachedPriceWithMarkup);
+            uint8 tokenDecimals = token.decimals();
+            uint256 tokenAmount = weiToToken(
+                preChargeNative, 
+                tokenDecimals,
+                cachedPriceWithMarkup
+            );
             SafeERC20.safeTransferFrom(token, userOp.sender, address(this), tokenAmount);
             context = abi.encode(tokenAmount, userOp.sender);
             validationResult = _packValidationData(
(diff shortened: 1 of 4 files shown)


drortirosh:
are you sure?
we do have tests that work with decimals=8.
If you do think there is a problem, please also try add a test to it as well.

alleneubank:
Funny, I encountered the same thing when adding the TokenPaymaster to https://github.com/0xsend/sendapp. Really racked my head on this one. I believe for the most part all the calculations are correct except for the ones in UniswapHelper. I think it's missing the decimal scaling when converting. I had a similar approach in our version of UniswapHelper. You can see that here, 
https://github.com/eth-infinitism/account-abstraction/compare/develop...0xBigBoss:account-abstraction:develop.

I didn't spend too much time looking at fixing the tests, but you can see now that the actualTokenCharge amounts are inline with what they should be. 

```shell
actualTokenChargeEvents= 1106
actualTokenCharge= 1106
actualTokenChargeEvents= 1106
expectedTokenCharge= 1106842507747897
    1) should be able to sponsor the UserOp while charging correct amount of ERC-20 tokens
```

If I understand correctly, the decimal for the token in `TokenPaymaster.test.ts` is 6, which means this transaction would have used $0.001106 as opposed to $1,106,842,507.747897 😅 

Screenshot 2024-04-01 at 21 05 44

I re-wrote most of the tests in solidity for those interested. https://github.com/0xsend/sendapp/blob/7fa242d2069dbfa71c475f73020b17f96670a970/packages/contracts/test/TokenPaymaster6.t.sol#L27

a00012025:
@drortirosh Hi, I've updated the test to reflect my changes and make them pass. I also updated the assertion of charged token price to avoid floating point error. Please help check it, thanks.

a00012025:
@drortirosh Hi, can you help check this? Thanks

Dropkick13 (approved):
[]()

0xFirekeeper:
+1, had to make same change for usdc with chainlink oracle

drortirosh:
moving all samples to another repo.

PR #461, opened 2024-03-12
refactor: `PackedUserOperation` & `MemoryUserOp`, `UserOperationLib` views

pegahcarter:
- Refactored `MemoryUserOp` to have a packed struct, reducing the size of the struct from 10 slots into 7.  The changes are converting gas values to uint128 instead of uint256, which is already assumed given we are unpacking a bytes32 into two equal-sized uints.
- Reordered `PackedUserOperation` to match the refactored `MemoryUserOp`.
- Fixed data types returned in `UserOperationLib` to match their type within `PackedUserOperation`.

### DIFF
--- a/contracts/core/EntryPoint.sol
+++ b/contracts/core/EntryPoint.sol
@@ -80,7 +80,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuard,
     )
     internal
     returns
-    (uint256 collected) {
+    (uint128 collected) {
         uint256 preGas = gasleft();
         bytes memory context = getMemoryBytesFromOffset(opInfo.contextOffset);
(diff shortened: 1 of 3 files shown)


drortirosh:
1. this code doesn't compile, so we can't really consider it.
2. before submitting an optimization, make sure to run `yarn gas-calc`, to see the gas usage of the new code, and if indeed it saves gas for different batch sizes.
3. Specifically for this issue: the public interface is `PackedUserOperation`, and the internal structure (for optimized memory usage) is `MemoryUserOp`, so it makes little sense to refactor the public interface to match the inner structure
4. From our experience, solidity performs better when in-memory variables (like MemoryUserOp) are aligned on 256-bit. packing saves gas when used as storage or as calldata

PR #464, opened 2024-03-21
Update `actualGasCost` description.

doublespending:
Origin description of `actualGasCost` makes developers easy to think it as `actualGas`. However, it is actually the `actualGasCost = actualGas * gasPrice`.
```
* @param actualGasCost - Actual gas used so far (without this postOp call).
```

Referring to the description of `maxCost`, I update the `actualGasCost` description.
```
* @param maxCost         - The maximum cost of this transaction (based on maximum gas and gas price from userOp).
```
```
* @param actualGasCost - The actual cost of this transaction so far (without this postOp call).
```


Dropkick13 (approved):
[]()

shahafn:
While technically true, this file is imported in EntryPoint and would change its address. Once we work on a new version it will be introduced.

PR #476, opened 2024-04-25
remove postOpGasLimit from estimateUserOperationGas

drortirosh:
The value of postOpGasLimit can't be reliably estimated: the paymaster depends on this value, and can't trust the account to estimate a proper value, so it must require a fixed, known value.


Dropkick13 (commented):
[]()

Dropkick13 (approved):
[]()

PR #478, opened 2024-04-29
Fix parameters list for eth_sendUserOperation

hazim-j:
Noticed `factory` and `paymaster` fields are listed as parameters for `eth_sendUserOperation`. Assuming this is a mistake since they would already be encoded in `UserOperation`.


drortirosh:
no, this is not a mistake.
Notice that with ver0.7, a `UserOperation` struct contains all factory and paymaster fields as optional.
The `PackedUserOperation` which is the struct which we call the contracts are encoded.
The idea that sdk/wallet can work with the expanded fields, and doesn't have to know about the packed version, except for the purpose of signing
(much like normal transaction, where a wallet always work with a "json struct" , and only care about RLP-encoded version in order to create a signature)

Dropkick13 (commented):
[]()

Dropkick13 (approved):
[]()

PR #480, opened 2024-05-09
Packed int values should be defined as uint128 to prevent overflow

hazim-j:
Since all gas related values (with the exception of `preVerificationGas`) are capped to 16 bytes in a `PackedUserOperation`, should these values be defined as type `uint128` in a `UserOperation`?


drortirosh:
we now maintain the ERCs on the public repo. please open the PR there..

Dropkick13 (commented):
[]()

PR #482, opened 2024-05-09
Remove v0.6 deployment files, ERC-4337.md and ERC-7562.md files

forshtat:
ERCs are now maintained on the public repo, at
https://github.com/ethereum/ERCs/blob/master/ERCS/erc-4337.md


drortirosh (approved):

bh2smith:
Hey --- I noticed that the deployment addresses were removed here, but that file referenced here also doesn't contain deployment addresses! Without these, I am not sure how we are expected to find them (except requesting them from bundlers via `eth_supportedEntryPoints`)

PR #488, opened 2024-06-19
Add wallet factory integration to EntryPoint contract

ylv-io:
… and new IWalletFactory interface.


drortirosh:
EntryPoint is immutable contract, and should never rely in external factory. Otherwise, it might behave differently on different chains.

PR #496, opened 2024-08-24
Update ECDSA signature length check

YanhuiJessica:
`recover(bytes32,bytes)` no longer accepts compact signatures (i.e. 64-byte signatures) since 4.7.3.
Because `package.json` uses Openzeppelin ^5.0.0, consider only checking if the signature is 65 bytes long.


drortirosh:
actually, the entire signature length check should be removed from here... the only reason it was added, was to have an error reason starting with "VerifyingPaymaster" rather than "ECDSA".
But with the latest version, the inner exception is typed error anyway, and we use FailedOpWithRevert, to carry both AA error and actual inner revert reason

PR #503, opened 2024-10-18
Allows Bundler to be the paymaster without Paymaster contract

nivida:
Currently, how the ``balance`` is checked against the ``requiredPreFund`` value does not allow the Bundler itself to fund the execution without having a paymaster contract involved. Also, it will increase the readability and correctness.


drortirosh:
There is a clear separation between a bundler (off-chain entity) and a paymaster (a contract)

nivida:
Yes, that's true. But shouldn't we allow the balance to be equal to the required funds to execute it and not only bigger?

drortirosh:
technically, you're right. but the difference is one WEI. the `requiredPreFund` is already an upper boundary - the maximum of all gas limits, multiplied by the maxFeePerGas (not actual gas).
Actual fee is measured in 1-100 gwei, that is, more than 10**9, so this 1 wei buffer is many order of magnitudes insignificant.

Issue #500, opened 2024-09-26
Add ERC1271 support for SimpleAccount


chaitanyapotti:
Now that ERC1271, ERC6492 are supported broadly in viem and many dapps are adopting this standard, it makes sense to add a default implementation in SimpleAccount.

livingrockrises:
I've added this PR
https://github.com/eth-infinitism/account-abstraction/pull/511
I will try and write the hardhat test cases over the weekend

drortirosh:
closed, as we're moving elaborated samples to a separate repo

PR #511, opened 2024-11-28
feat: simple account 1271

livingrockrises:
https://github.com/eth-infinitism/account-abstraction/issues/500


drortirosh:
1. We're in a process of moving samples into a separate project
2. The SimpleAccount is meant as MVP to showcase the usage of ERC-4337 protocol. There are many other features of account that are missing to make it a full-fledged account, erc-1271 is only one of them

PR #512, opened 2024-12-10
PBHSignatureAggregator

karankurbur:
- change interface to remove view modifier from validateSignatures and validateUserOpSignature. This is required since we may have state change inside PBHValidator 
- create PBHSignatureAggregator that inherit the BLSSignatureAggregator class
- add new function to return encoded BLS signature + proofs array
- modify the validateSignatures and validateUserOpSignature to remove the proof before verifying the BLS signature through the BLSSignatureAggregator
- verify the proof through PBHVerifier

TODO:
- Check E2E if modifying the interface reverts from EntryPoint
- Unit tests to validate encoding/decoding logic
- Refactor BLSSignatureAggregator to use created internal functions

### DIFF
--- a/contracts/interfaces/IAggregator.sol
+++ b/contracts/interfaces/IAggregator.sol
@@ -16,7 +16,7 @@ interface IAggregator {
     function validateSignatures(
         PackedUserOperation[] calldata userOps,
         bytes calldata signature
-    ) external view;
+    ) external;
 
     /**
      * Validate signature of a single userOp.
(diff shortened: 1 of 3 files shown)


drortirosh:
Thank you for your submission, 
1. We are in a process of moving samples out of the core project.
2. Still, we try to keep the samples project minimal, and I'm not sure we'll add more technical samples. a ZK-based account can be in a separate project.
3. Indeed, we do think of removing the "view" from `validateSignatures`, as it is too restrictive. This is the only aggregator method that is called by the EntryPoint on-chain.
4. I don't see the reason of making `aggregateSignatures` (or `validateUserOpSignature`) non-view, as these methods are never called on-chain, only off-chain by the bundler. making them Mutable might even give the false illusion that they can modify the state..
5. A full sample should include also a sample account that uses the new scheme - the account must return the aggregator in its validate function.

drortirosh:
closing this issue. if you think it is still valuable, you may re-submit it into the "samples" project.

PR #514, opened 2024-12-17
AA-466: Require SenderCreator to be called from EntryPoint & AA-470: Make SenderCreator public

forshtat:


contracts/core/SenderCreator.sol
```diff
@@ -15,6 +21,9 @@ contract SenderCreator {
     function createSender(
         bytes calldata initCode
     ) external returns (address sender) {
+        if (msg.sender != entryPoint) {
+            revert("AAxx should call from EntryPoint");
```
drortirosh: this is not an expected error case to happen with entrypoint, only by "side-calling".
So it is an AA9x.
Can reuse "AA92 only from entrypoint"

drortirosh:
the change is SenderCreator is not enough and not clear by itself: need to add 
`require(msg.sender==entrypoint.senderCreator())`
To `SimpleAccountFactory` (that's exactly why a sample working account implementation is part of this repo: To show how it is used with the framework)

contracts/core/SenderCreator.sol
forshtat: Didn't want to reuse error codes, defined it s `AA97`.

drortirosh (approved):

PR #522, opened 2024-12-26
AA-499 fix EntryPointSimulations config

drortirosh:
EntryPointSimulations should be compiled with the same compiler settings as EntryPoint


forshtat (approved):
@drortirosh this change looks very ad-hoc for the `EntryPoint` contract, don't you agree?
If it's unavoidable we can merge it.

PR #526, opened 2025-01-13
AA-508 eip712 user op hash

drortirosh:


contracts/core/UserOperationLib.sol
```diff
@@ -131,7 +137,7 @@ library UserOperationLib {
      * Hash the user operation data.
      * @param userOp - The user operation data.
      */
-    function hash(
+    function hash1(
```
shahafn: Why?

contracts/core/EntryPointSimulations.sol
```diff
@@ -191,4 +196,24 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
...
+    bytes32 private constant TYPE_HASH =
+    keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+
+    function __buildDomainSeparator() private view returns (bytes32) {
+        bytes32 _hashedName = keccak256(bytes(DOMAIN_NAME));
+        bytes32 _hashedVersion = keccak256(bytes(DOMAIN_VERSION));
+        return keccak256(abi.encode(TYPE_HASH, _hashedName, _hashedVersion, block.chainid, address(this)));
+    }
+
+    //can't rely on "immutable" (constructor-initialized) variables" in simulation
```
shahafn, line 209: Can't it be supplied with `stateOverrides` that we supply anyway?

drortirosh: This is a storage cell that only exists in simulation. seems simpler to initialize it in init-code, just like senderCreator.

contracts/samples/SimpleAccount.sol
```diff
@@ -104,8 +104,9 @@ contract SimpleAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, In
     /// implement template method of BaseAccount
     function _validateSignature(PackedUserOperation calldata userOp, bytes32 userOpHash)
     internal override virtual returns (uint256 validationData) {
-        bytes32 hash = MessageHashUtils.toEthSignedMessageHash(userOpHash);
-        if (owner != ECDSA.recover(hash, userOp.signature))
+
+        //userOpHash can be generated using eth_signTypedData_v4
```
forshtat, line 108: This comment is unnecessary

contracts/core/EntryPointSimulations.sol
```diff
@@ -16,10 +16,15 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
 
     SenderCreator private _senderCreator;
 
+    //non-immutable, as EntryPointSimulations is used with state-override, without a constructor
```
forshtat: Nobody needs this comment really

forshtat (dismissed):

contracts/samples/SimpleAccount.sol
drortirosh: I think it is important, as there is no other place in our codebase that explains that this sig is directly a signed-typed-data signature (note that the old signature that was removed by this commit explicitly had "toEthSignedMessageHash", but for the new signature, you need to look back to where we calculate the userOpHash to see it is an erc-712 signature

shahafn (approved):

Speedcode88ss (approved):

PR #529, opened 2025-01-23
AA-521 EntryPoint support for eip-7702

drortirosh:
Native support for EIP-7702 in EntryPoint.
Depends on EIP-7702 PR https://github.com/ethereum/EIPs/pull/9248 (extcodehash to return the delegate)


contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,58 @@
...
+import "../core/UserOperationLib.sol";
+// SPDX-License-Identifier: MIT
+
+// EIP-7702 code prefix. Also, we use this prefix as a marker in the initCode. To specify this account is EIP-7702.
+uint256 constant EIP7702_PREFIX = 0xef0100;
+
+using UserOperationLib for PackedUserOperation;
+
+    //get alternate InitCode (just for hashing) when using EIP-7702
+    function _getEip7702InitCodeOverride(PackedUserOperation calldata userOp) view returns (bytes32) {
```
forshtat: This function returns a hash, this is not a correct name.

contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,58 @@
...
+        }
+        address delegate = _getEip7702Delegate(userOp.getSender());
+        if (initCode.length < 20)
+            return keccak256(abi.encodePacked(delegate));
+        else
+            return keccak256(abi.encodePacked(delegate, initCode[20 :]));
+    }
+
+
+    function _isEip7702InitCode(bytes calldata initCode) pure returns (bool) {
```
forshtat: If you write this function without the assembly tricks, how much more gas does it spend? This code is too elaborate for the task of "compare first three bytes".

contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,58 @@
...
+ * requires EXTCODECOPY pr: https://github.com/ethereum/EIPs/pull/9248 (not yet merged or implemented)
+ **/
+    function _getEip7702Delegate(address sender) view returns (address) {
+        uint256 senderCode;
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly ("memory-safe") {
+            extcodecopy(sender, 0, 0, 32)
+            senderCode := mload(0)
+        }
```
forshtat: You don't need any assembly, you can just do `sender.code`. Also if the code is not exactly 23 bytes long it should revert.

contracts/core/EntryPoint.sol
```diff
@@ -441,6 +443,11 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
     ) internal {
         if (initCode.length != 0) {
             address sender = opInfo.mUserOp.sender;
+            if ( _isEip7702InitCode(initCode) ) {
+                //already validated it is an EIP-7702 delegate (and hence, already has code)
+                senderCreator().initEip7702Sender(sender, initCode[20:]);
```
forshtat: I was under impression that you wanted to call `initEip7702Sender` only if `initCode` contains some data in addition to `0xff0100`, but this code seems to call it every time.

contracts/core/EntryPointSimulations.sol
```diff
@@ -192,7 +192,7 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
 
     //slightly stricter gas limit than the real EntryPoint
     function _getVerificationGasLimit(uint256 verificationGasLimit) internal pure virtual override returns (uint256) {
-        return verificationGasLimit - 300;
+        return verificationGasLimit - 350;
```
forshtat: Magic number. As you need to change it, please extract a constant.

contracts/core/UserOperationLib.sol
```diff
@@ -58,10 +58,22 @@ library UserOperationLib {
...
+     * @param overrideInitCode - If set, encode this instead of the initCode field in the userOp.
+     */
+    function encode(
+        PackedUserOperation calldata userOp,
+        bytes32 overrideInitCode
     ) internal pure returns (bytes memory ret) {
         address sender = getSender(userOp);
         uint256 nonce = userOp.nonce;
-        bytes32 hashInitCode = calldataKeccak(userOp.initCode);
+        bytes32 hashInitCode = overrideInitCode==0 ? calldataKeccak(userOp.initCode) : overrideInitCode;
```
forshtat: Just make a function `_getEip7702InitCodeHash` and call it from here, maybe?

forshtat (changes requested):

contracts/core/EntryPointSimulations.sol
drortirosh: unrelated. we're going to remove "Simulation" from the critical path, and thus all this code will be removed.

contracts/core/Eip7702Support.sol
drortirosh: 1. using assembly to avoid creating a dynamic bytes array (even with sender.code, we'd need assembly to grab just the address out of it, as there is no "slice" for memory..)
2. No need to check it is 23 bytes: since eip-3541, no deployed contract can start with "0xEF", unless it is EOF. The "0xEF0100" prefix of eip-7702 is invalid as EOF, such checking the prefix implies the 23 bytes length...

forshtat: > Even with sender.code, we'd need assembly to grab just the address out of it

I think Solidity now supports just converting `bytes` to `bytes32` to `uint256` so as far as I can tell there should be no difference. I may be wrong though.
```
        bytes memory b = sender.code;
        uint256 senderCode = uint256(bytes32(b));
```

> No need to check it is 23 bytes since EIP-3541

I am concerned some up-and-coming L2s may be a little less rigorous in their implementation of EIP-3541 and that would allow some kind of an attack down the line. It would be dirt-cheap to just double-check.

contracts/core/Eip7702Support.sol
drortirosh: initCode starts with a prefix, and can be padded with zeros.
in assembly, we take advantage of ABI encoding, which pads shorter values with zeros to next 32 boundary.
In solidity, this padding would have to be done manually. 
not only it would cost more, it would also be a longer code.

contracts/core/Eip7702Support.sol
drortirosh: yes, it is possible. but it reads read entire code, and then extract just the first 32 bytes. 
rough check shows it to be 300 gas.
EIP-3541 doesn't really matter here: even if  account's code starts with "EF", it will never pass validation anyway (or any other execution call)

forshtat: > it reads read entire code, and then extract just the first 32 bytes

But the code is always 23 bytes long for EIP-7702, what is the problem here?
Please remove assembly if it is technically possible here. No need to overcomplicate things.

contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,65 @@
...
+            initCodeStart := calldataload(initCode.offset)
+        }
+        // make sure first 20 bytes of initCode are "0xff0100" (padded with zeros)
+        // initCode can be shorter (e.g. only 3), but then it is already zero-padded.
+        return (initCodeStart >> (256 - 160)) == ((EIP7702_PREFIX << (160 - 24)));
+    }
+
+/**
+ * get the EIP-7702 delegate from contract code.
+ * requires EXTCODECOPY pr: https://github.com/ethereum/EIPs/pull/9248 (not yet merged or implemented)
```
forshtat: This PR is merged so the comment may be removed.

contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,65 @@
...
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly ("memory-safe") {
+            extcodecopy(sender, 0, 0, 32)
+            senderCode := mload(0)
+        }
+        // senderCode is the first 32 bytes of the sender's code
+        // If it is an EIP-7702 delegate, then top 24 bits are the EIP7702_PREFIX
+        // next 160 bytes are the delegate address
+        if (senderCode >> (256 - 24) != EIP7702_PREFIX) {
```
forshtat: You could convert it by doing something like this:

```
bytes3 EIP7702_PREFIX = 0xef0100;
bytes32 senderCode = ...;
if (bytes3(senderCode) != EIP7702_PREFIX) {}
```

It is basically the same thing but `>> (256 - 24) !=` does not look good...

You could even define it as `bytes3 senderCode` if you wanted to, I think. Not sure about that.

contracts/core/SenderCreator.sol
```diff
@@ -40,10 +41,24 @@ contract SenderCreator is ISenderCreator {
...
-            sender := mload(0)
+            if success {
+                sender := mload(0)
+            }
         }
+    }
+
+    // use initCode to initialize an EIP-7702 account
+    // caller (EntryPoint) already verified it is an EIP-7702 account.
+    function initEip7702Sender(
```
forshtat, line 55: Why can't we use `createSender` again? It is doing basically the same thing, calling `address(initCode[:20]` with `initCode[20:]`, what are the main differences? 
If they are not critical, let's merge these two methods?

contracts/core/EntryPoint.sol
```diff
@@ -378,8 +379,9 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
     function getUserOpHash(
         PackedUserOperation calldata userOp
     ) public view returns (bytes32) {
+        bytes32 overrideInitCodeHash = _getEip7702InitCodeHashOverride(userOp);
         return
-            MessageHashUtils.toTypedDataHash(getDomainSeparatorV4(), userOp.hash());
+            MessageHashUtils.toTypedDataHash(getDomainSeparatorV4(), userOp.hash(overrideInitCodeHash));
```
forshtat: `userOp.hash()` may only ever be called with the result of `_getEip7702InitCodeHashOverride()`, which can only be called with the `userOp` as a parameter. The value of `overrideInitCodeHash` is only used once.

Do we need this value to bubble up into the EntryPoint itself? Why not call the `_getEip7702InitCodeHashOverride` directly inside the `userOp.hash`?

contracts/core/EntryPoint.sol
```diff
@@ -441,6 +443,13 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
     ) internal {
         if (initCode.length != 0) {
             address sender = opInfo.mUserOp.sender;
+            if ( _isEip7702InitCode(initCode) ) {
+                if (initCode.length>20 ) {
+                    //already validated it is an EIP-7702 delegate (and hence, already has code)
+                    senderCreator().initEip7702Sender(sender, initCode[20:]);
```
forshtat, line 452: This means that EIP-7702 accounts can get two validation phase calls instead of one whenever they want, right?

forshtat (changes requested):

contracts/core/EntryPoint.sol
drortirosh: yes. eip-7702 account is never "new" (from the EntryPoint's view) - it is always "already created".
So we always allow an eip-7702 account to run "initialization" code - as many times as it wishes.
The alternative was to require it to put this initialization data into the "callCode", use it from "validate" and then ignore it during execution.

contracts/core/EntryPoint.sol
drortirosh: Separation of concerns: 
UserOperationLib is a library, and its methods (including `hash()`) are pure.
`_getEip7702InitCodeHashOverride()` is a view function, reading the current delegate.

contracts/core/SenderCreator.sol
drortirosh: 1. createSender is used to physically **create**, not initialize.
2. it would require changing createSender signature, since it requires an extra parameter (not extracting it from initCode)
3. we already suggested to use `call({from:entryPoint, to:senderCreator, data: encode(createSender(...) }` as a view function to get account address. changing the API would break it.
4. so seemed like adding a new one was a better choice.

contracts/core/Eip7702Support.sol
drortirosh: I know it is 23 bytes and you know it too, but solidity doesn't... `sender.code` can possibly be any length, so it copies it as dynamic value before extracting the first word/

contracts/core/Eip7702Support.sol
```diff
@@ -0,0 +1,61 @@
+pragma solidity ^0.8;
+// SPDX-License-Identifier: MIT
+// solhint-disable no-inline-assembly
+
+import "../interfaces/PackedUserOperation.sol";
+import "../core/UserOperationLib.sol";
+
+// EIP-7702 code prefix. Also, we use this prefix as a marker in the initCode. To specify this account is EIP-7702.
+bytes3 constant EIP7702_PREFIX = 0xef0100;
```
forshtat: For the record, I suggest here splitting the two values as there isn't actually an overlap in the meaning:

```
// EIP-7702 code prefix
uint256 constant EIP7702_PREFIX = 0xef0100;
// Prefix as a marker in the "initCode" to specify this account is EIP-7702
uint256 constant EIP7702_PREFIX_INITCODE = 0xff7702;
```

forshtat (dismissed):

forshtat (dismissed):

forshtat (approved):

PR #532, opened 2025-01-30
AA-467 Fixing actual gas used penalty

shahafn:


contracts/core/EntryPoint.sol
```diff
@@ -732,22 +740,12 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-                uint256 executionGasUsed = actualGas - opInfo.preOpGas;
-                // this check is required for the gas used within EntryPoint and not covered by explicit gas limits
-                if (executionGasLimit > executionGasUsed) {
-                    uint256 unusedGas = executionGasLimit - executionGasUsed;
-                    uint256 unusedGasPenalty = (unusedGas * PENALTY_PERCENT) / 100;
-                    actualGas += unusedGasPenalty;
-                }
-            }
-
+            // Calculating a penalty for unused postOp gas
```
forshtat: This comment is not in the right place

contracts/core/EntryPoint.sol
```diff
@@ -818,4 +816,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
         (bool success, bytes memory ret) = target.delegatecall(data);
         revert DelegateAndRevert(success, ret);
     }
+
+    function _getUnusedGasPenalty(uint256 gasUsed, uint256 gasLimit) internal pure returns (uint256) {
+        if (gasLimit > gasUsed) {
```
forshtat: How can it be the other way around?

Also, flip this `if` to exit early with 0 please

contracts/core/EntryPointSimulations.sol
```diff
@@ -192,7 +192,7 @@ contract EntryPointSimulations is EntryPoint, IEntryPointSimulations {
 
     //slightly stricter gas limit than the real EntryPoint
     function _getVerificationGasLimit(uint256 verificationGasLimit) internal pure virtual override returns (uint256) {
-        return verificationGasLimit - 300;
+        return verificationGasLimit - 500;
```
forshtat, line 195: Magic number. Let's create a constant, seems like we need to change it again and again.

test/entrypoint.test.ts
```diff
@@ -74,10 +77,11 @@ describe('EntryPoint', function () {
 
   let accountOwner: Wallet
   const ethersSigner = ethers.provider.getSigner()
-  let account: SimpleAccount
+  let simpleAccount: SimpleAccount
```
forshtat, line 80: Why was it renamed? It is not related to the issue, it seems.

forshtat (changes requested):

contracts/core/EntryPoint.sol
shahafn: Yes moved some lines

contracts/core/EntryPoint.sol
shahafn: Fixed

contracts/core/EntryPoint.sol
```diff
@@ -818,4 +816,14 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
         (bool success, bytes memory ret) = target.delegatecall(data);
         revert DelegateAndRevert(success, ret);
     }
+
+    function _getUnusedGasPenalty(uint256 gasUsed, uint256 gasLimit) internal pure returns (uint256) {
+        if (gasLimit <= gasUsed) {
```
drortirosh: wrap with "unchecked"

contracts/test/TestPaymasterWithPostOp.sol
```diff
@@ -20,7 +21,8 @@ contract TestPaymasterWithPostOp is TestPaymasterAcceptAll {
         return ("1", SIG_VALIDATION_SUCCESS);
     }
 
-    function _postOp(PostOpMode, bytes calldata, uint256, uint256)
+    function _postOp(PostOpMode, bytes calldata, uint256 actualGasCost, uint256)
     internal override {
+        emit PostOpActualGasCost(actualGasCost);
```
drortirosh, line 26: This new emit breaks the gas-calc, as it adds another >1000 gas to the postOp.
The gas-calc for paymaster+postOp raised by 1200, while the cost of the EntryPoint code change (for postOp penalty cost) was about 150

We should have a separate GasCalcPaymasterWithPostOp.

shahafn: Actually I removed its usage so it can be removed

contracts/core/EntryPointSimulations.sol
drortirosh: I don't think it is needed at all. and with handleOps for validation, it will soon go away anyway.

contracts/test/TestPaymasterWithPostOp.sol
drortirosh: that would be nice, though I currently see it is referenced from a test.

test/entrypoint.test.ts
shahafn: Because in a test file with 1500 lines and many tests, all accounts are called "account" which is not convenient.

contracts/core/EntryPoint.sol
shahafn: Fixed

contracts/test/TestPaymasterWithPostOp.sol
shahafn: Added a gas calc paymaster

drortirosh (approved):

forshtat (approved):

PR #534, opened 2025-02-10
AA-507 Min Penalty Threshold

shahafn:


contracts/core/EntryPoint.sol
```diff
@@ -50,7 +50,10 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
     bytes32 private constant INNER_REVERT_LOW_PREFUND = hex"deadaa51";
 
     uint256 private constant REVERT_REASON_MAX_LEN = 2048;
+    // Penalty charged for either unused execution gas or postOp gas
     uint256 private constant PENALTY_PERCENT = 10;
+    // Threshold below which no penalty would be charged
+    uint256 private constant PENALTY_GAS_THRESHOLD = 4e4;
```
drortirosh, line 56: 1. don't use exponential notation in solidity code. better 40_000
2. what is the rationale for this value?

contracts/core/EntryPoint.sol
```diff
@@ -819,7 +822,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
 
     function _getUnusedGasPenalty(uint256 gasUsed, uint256 gasLimit) internal pure returns (uint256) {
         unchecked {
-            if (gasLimit <= gasUsed) {
+            if (gasLimit <= gasUsed || gasLimit - gasUsed <= PENALTY_GAS_THRESHOLD) {
```
drortirosh, line 825: Damn solidity, doesn't optimize `gasLimit-gasUsed` which is done twice...
better add below `if (unusedGas<=PENALTY_GAS_THRESHOLD) return 0;`
(which is also slightly more readable)

drortirosh (approved):

PR #536, opened 2025-02-11
AA-525: Create Simple7702Account

drortirosh:
Sample ERC-4337 account that uses EIP-7702.
Supports also direct (batched) calls from the EOA account itself.


contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Simple7702Account.sol
+ * A minimal account to be used with EIP-7702 (for batching) and ERC-4337 (for gas sponsoring)
+ */
+contract Simple7702Account is BaseAccount, IERC165, IERC1271, ERC1155Holder, ERC721Holder {
+
+    // temporary address of entryPoint v0.8
+    function entryPoint() public pure override returns (IEntryPoint) {
+        return IEntryPoint(0x6F4F5099a64044D69EB7419d66760fD4106fcE3C);
+    }
```
forshtat: Why `function` and not an an immutable parameter?

contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+        bytes32 userOpHash
+    ) internal virtual override returns (uint256 validationData) {
+
+        if (address(this) != ECDSA.recover(userOpHash, userOp.signature)) {
+            return SIG_VALIDATION_FAILED;
+        }
+        return 0;
+    }
+
+    function _requireFromSelfOrEntryPoint() internal view virtual {
```
forshtat, line 49: Why `function` and not a `modifier`?

contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    function execute(Call[] calldata calls) external {
+        _requireFromSelfOrEntryPoint();
+
+        for (uint256 i = 0; i < calls.length; i++) {
+            Call calldata call = calls[i];
+            (bool ok, bytes memory ret) = call.target.call{value: call.value}(call.data);
+            if (!ok) {
+                // solhint-disable-next-line no-inline-assembly
+                assembly { revert(add(ret, 32), mload(ret)) }
```
forshtat: On one hand, it is nice and simple how we are not wrapping the reverted data into our own exception. On the other hand, we are losing some information here - specifically, which `Call` is the one that reverted. I think this information may be useful enough for the user and developers, so we should reconsider and revert with a custom error here. 
What do you think?

contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+        return
+            id == type(IERC165).interfaceId ||
+            id == type(IAccount).interfaceId ||
+            id == type(IERC1271).interfaceId ||
+            id == type(IERC1155Receiver).interfaceId ||
+            id == type(IERC721Receiver).interfaceId;
+    }
+
+    //deliberately return the same signature as returned by the EOA itself: This way,
+    // ERC-1271 can be used regardless if the account currently has this code or not.
```
forshtat: This comment is not very clear and not properly formatted.

contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Make this account callable through ERC-4337 EntryPoint.
+     * The UserOperation should be signed by this account's private key.
+     */
+    function _validateSignature(
+        PackedUserOperation calldata userOp,
+        bytes32 userOpHash
+    ) internal virtual override returns (uint256 validationData) {
+
+        if (address(this) != ECDSA.recover(userOpHash, userOp.signature)) {
```
forshtat: There is a method `isValidSignature` that does almost the same thing, so the code should kind of be de-duplicated a little bit here.

contracts/samples/Simple7702Account.sol
```diff
@@ -0,0 +1,84 @@
...
+            id == type(IERC721Receiver).interfaceId;
+    }
+
+    //deliberately return the same signature as returned by the EOA itself: This way,
+    // ERC-1271 can be used regardless if the account currently has this code or not.
+    function isValidSignature(bytes32 hash, bytes memory signature) public view returns (bytes4 magicValue) {
+        return ECDSA.recover(hash, signature) == address(this) ? this.isValidSignature.selector : bytes4(0);
+    }
+
+    receive() external payable {
```
forshtat, line 84: If I recall correctly, `receive()` will revert if there is any `msg.data` included in an incoming transaction, and you must add `fallback() payable` as well to catch all incoming value.

forshtat (changes requested):

contracts/samples/Simple7702Account.sol
drortirosh: it `override`s our base contract virtual function

contracts/samples/Simple7702Account.sol
drortirosh: `fallback()` means "don't revert on any incoming calls with signatures you don't know". This is not a good practice.

contracts/samples/Simple7702Account.sol
drortirosh: Yes, the logic is the same, the inputs (and outputs) differ.
In some wallets, their implementations might differ, so I'm not sure it worthwhile to create a shared function.
e.g. for SCA in general, isValidSignature should encode the owner as well as the account itself (eip7702 account is unique, in that they are the same.)

contracts/samples/Simple7702Account.sol
drortirosh: Because I dislike modifiers... 
They are syntactic sugar, and defeat the optimizer (e.g. multiple usages will always duplicate the code)
Also, they can't be inherited and overridden in a subclass.

contracts/samples/Simple7702Account.sol
drortirosh: This is an explanation of why the signature should be the same, but not sure it is needed at all.

contracts/samples/Simple7702Account.sol
drortirosh: We also use a "require" function in BaseAccount ( `_requireFromEntryPoint()`) - but here we explicitly need self-or-entrypoint.

contracts/samples/Simple7702Account.sol
forshtat: Say it when a transaction sending you 1000 ETH reverts because it assumed your address is an EOA and put some calldata into it...

contracts/samples/Simple7702Account.sol
forshtat: > In some wallets, their implementations might differ

But not in this one, so why do you care about what other wallets might do?

contracts/samples/Simple7702Account.sol
drortirosh: You mean accept every method call into the account without revert? 
This is not the best practice for an SCA (though it better matches the behaviour of an EOA)

forshtat: I think a compromise would be to revert if the value is zero, but accept all incoming value. There will be a transition period where dapps might not know 7702 accounts have code and treat them as EOAs, stuff like that.

contracts/samples/Simple7702Account.sol
drortirosh: even if we tell which one reverted, there might still some inner revert info which caused it (e.g. if an external contract revere. for truly "decoding" who reverted, you need tracing.
creating here a wrapper revert doesn't add much information.

forshtat (approved):

PR #537, opened 2025-02-16
execute in BaseAccount

drortirosh:


contracts/samples/SimpleAccount.sol
```diff
@@ -97,7 +60,7 @@ contract SimpleAccount is BaseAccount, TokenCallbackHandler, UUPSUpgradeable, In
     }
 
     // Require the function call went through EntryPoint or owner
-    function _requireFromEntryPointOrOwner() internal view {
+    function _requireForExecute() internal view override virtual {
```
forshtat, line 68: This doesn't look like a good rename to me. Also, `virtual`? When do you expect the subclasses will need to override this function?

contracts/core/BaseAccount.sol
```diff
@@ -31,6 +41,64 @@ abstract contract BaseAccount is IAccount {
      */
     function entryPoint() public view virtual returns (IEntryPoint);
 
+    /**
+     * execute a single call from the account.
+     */
+    function execute(address target, uint256 value, bytes calldata data) virtual external {
```
forshtat, line 47: I have to note that this `execute` function is a strict subset of what `executeBatch` can do. It is trivially easy for any wallet to always make a call to `executeBatch` with a single call.
There is also no large legacy infrastructure around `execute(address,uint256,bytes)` functions in smart contract wallets.
I know we always had it in the Sample Account, but moving both of them to the **core/**`BaseAccount`  is a little bit too much.
I vote we drop the single `execute` and only leave `executeBatch`.

contracts/core/BaseAccount.sol
```diff
@@ -31,6 +41,64 @@ abstract contract BaseAccount is IAccount {
...
+    }
+
+    /**
+     * withdraw value from the account's deposit
+     * @param withdrawAddress target to send to
+     * @param amount to withdraw
+     */
+    function withdrawDepositTo(address payable withdrawAddress, uint256 amount) public onlyOwner {
+        entryPoint().withdrawTo(withdrawAddress, amount);
+    }
```
forshtat: I am not convinced these three must be a part of the BaseAccount. This are utility functions that we can trust the wallets to implement themselves.

contracts/utils/Exec.sol
```diff
@@ -40,11 +40,14 @@ library Exec {
     }
 
     // get returned data from last call or calldelegate
+    // maxLen - maximum length of data to return, or zero, for the full length
     function getReturnData(uint256 maxLen) internal pure returns (bytes memory returnData) {
         assembly ("memory-safe") {
             let len := returndatasize()
-            if gt(len, maxLen) {
-                len := maxLen
+            if iszero(iszero(maxLen)) {
```
forshtat: 1. I believe you mean `not(iszero(maxLen))` here. If so, it is concerning that this function is not covered by a test.
2. You don't need 0 as a special value, `getReturnData` is an `internal pure` function and it costs nothing to pass `type(uint256).max` as a `maxLen` parameter, which is also more explicit.

forshtat (changes requested):

contracts/utils/Exec.sol
forshtat: 3. And anyways, why did you decide to leave this return data unlimited and not use `REVERT_REASON_MAX_LEN` ?

contracts/samples/SimpleAccount.sol
drortirosh: it has different implementations for account with owner and 7702

contracts/core/BaseAccount.sol
drortirosh: execute is slightly more efficient, and I can see wallets do use it, when a batch is not required.

contracts/core/BaseAccount.sol
drortirosh: we now have 2 wallets (SimpleAccount, Simple7702Account) and both ned these methods.
A deposit is an extension of account's balance, and without it, it can't easily use it.
(It can call the entrypoint - but that requires the application to know the entrypoint interface)

contracts/utils/Exec.sol
drortirosh: "not" is bitwise, so not(1) is 0xfffffe ... iszero(iszero)) means "nonzero"
blocking maxlen is used during validation, where we want to prevent griefing. during execution, we try to put as few limitations as possible.
e.g. CCIP is a mechanism that returns (via revert) information to client to perform off-chain validation, before making a callback with the result. there is no hard limit on the data size, and we shouldn't arbitrarily put any.

yoavw (commented):
One thing that concerns me is whether BaseAccount is used by other projects where the new functions will conflict with theirs or have unexpected side effects.  Some of them are not virtual so if someone previously inherited it and added their own versions of these functions, won't it cause issues?  If others use this base contract and the only reason we moved the functions from SimpleAccount to BaseAccount is because they're also needed by Simple7702Account, wouldn't it make sense to add a new BaseSimpleAccount (that inherits BaseAccount) rather than add all the functions to Base?

contracts/core/BaseAccount.sol
drortirosh: actually, I agree: Simple7702Account doesn't need a deposit: it can be used as a direct batch call if it doesn't use a paymaster. removing from base.

contracts/samples/SimpleAccount.sol
drortirosh: suggest a better name: it is "entryPointOrSelf" for 7702, and "entryPointOrOwner" for any other account (e.g. SimpleAccount), so neither name is good as "base name"

shahafn: Why is EntryPointOrOwner not good? Self can be seen as owner technically. Plus, requireAccessControl or requireAccessPermitted are better ifbyou want to understand what they are for

shahafn (approved):

forshtat (approved):

Skinwalker778:
It will be fine

PR #545, opened 2025-02-23
A set of smaller changes to solidity contracts

forshtat:


contracts/core/EntryPoint.sol
```diff
@@ -233,7 +233,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
             if (address(aggregator) != address(0)) {
                 // solhint-disable-next-line no-empty-blocks
                 try aggregator.validateSignatures(ops, opa.signature) {} catch {
-                    revert SignatureValidationFailed(address(aggregator));
+                    revert SignatureValidationFailed(totalOps + i, address(aggregator));
```
drortirosh: opIndex is wrong here: aggregator failure is not attached to a specific UserOperation, but for a bulk of them.
aggregator should be unique, and thus reverting with an aggregator should be enough.
If you do want to give an index, then it should be the "aggIndex" which is just `i`

(Same index mistake with "AA96" above, which is also not a FailedOp - but there it is less critical, since it is just an "input sanitizer", on the ABI call from the bundler)

forshtat: @drortirosh could you please push a fix to both errors you've highlighted here?

contracts/core/EntryPoint.sol
```diff
@@ -227,7 +240,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
             //address(1) is special marker of "signature error"
             require(
                 address(aggregator) != address(1),
-                FailedOp(totalOps + i, "AA96 invalid aggregator")
+                SignatureValidationFailed(address(aggregator))
```
forshtat, line 243: I don't think this is correct

drortirosh: I think it is ridiculous to add specific revert reason which basically means "you violated the parameters usage". In most places, we used simple revert (with "Error(string)"). Using FailedOp is a mistake, so either the SignatureValidationFailed(aggregator) - or revert back to Error(string)

shahafn (approved):

PR #546, opened 2025-02-23
AA-542: add _tryDecrementDeposit

drortirosh:
refactor: use _tryDecrementDeposit to deduct value from account/pm


contracts/core/StakeManager.sol
```diff
@@ -43,17 +43,31 @@ abstract contract StakeManager is IStakeManager {
...
-        return newAmount;
+        unchecked {
+            DepositInfo storage info = deposits[account];
+            uint256 newAmount = info.deposit + amount;
+            info.deposit = newAmount;
+            return newAmount;
+        }
+    }
+
+    function _decrementDeposit(address account, uint256 amount) internal returns(bool) {
```
shahafn: Why returning bool when increment returns the new amount?

forshtat: @drortirosh I strongly suggest returning new value, reverting on underflow.

drortirosh: reverting in this method requires StakeManager to know the FailedOp error, and also the opIndex - both are irrelevant for it.

drortirosh: we return what is need.
too unfortunate that solidity can't return "err", like go. it would make "decrement" look better. returning amount doesn't help, since we only need to know if decrement caused an error (withdraw more than balance)
increment can't cause an error (as long as inputs are not out of bounds)

forshtat (dismissed):

contracts/core/StakeManager.sol
shahafn: At least rename to `_tryDecrementDeposit()`? The function doesn't always succeed

forshtat (approved):

contracts/core/StakeManager.sol
```diff
@@ -43,17 +43,38 @@ abstract contract StakeManager is IStakeManager {
...
+        unchecked {
+            DepositInfo storage info = deposits[account];
+            uint256 newAmount = info.deposit + amount;
+            info.deposit = newAmount;
+            return newAmount;
+        }
+    }
+
+    /**
+     * Try to dncrement the account's deposit.
```
gpersoon, line 63: typo: dncrement -> decrement

contracts/core/StakeManager.sol
```diff
@@ -43,17 +43,38 @@ abstract contract StakeManager is IStakeManager {
...
+     * @param account - The account to increment.
+     * @param amount  - The amount to increment by.
```
gpersoon, line 65: increment should be decrement

PR #547, opened 2025-02-25
AA-540: Extract the 'iterateValidationPhase' function

forshtat:


shahafn (approved):
I still think that at least for consistency, the same should be done with the execution loop. In both cases (validation and execution) we don't spare many duplicate lines anyway.

PR #548, opened 2025-02-25
AA-541: Make BasePaymaster use "Ownable2Step" instead of "Ownable"

forshtat:


drortirosh (dismissed):
1. why is it needed?
2. missing a test.
3. since new owner must be msg.sender, it doesn't work well for AA ...

shahafn:
@drortirosh:
1. ownable2step seems to be more secured, as it requires accepting the ownership transfer by the new owner. How is that an issue? how frequently do you expect people who use BasePaymaster, change ownership? and how inconvenient is it for them to accept ownership, given that the owner is an address they should control?
2. Testing OZ implementation?
3. Why not?

shahafn (approved):

drortirosh:
it is impossible for a "gasless" owner to use it. the "owner" of SimplerAccount is not an EOA: it is used just as signer, and we should not require it to use non-4337 transaction to change ownership

PR #552, opened 2025-02-27
AA-543: Refactor post execution

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -113,47 +116,68 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-            assembly ("memory-safe") {
-                let len := returndatasize()
-                if eq(32,len) {
-                    returndatacopy(0, 0, 32)
-                    innerRevertCode := mload(0)
-                }
-            }
+        bool callSuccess;
+        if (success) {
+            callSuccess = returnData != 0;
```
forshtat: I would suggest `executionSuccess` or something like that, more explicit.

contracts/core/EntryPoint.sol
```diff
@@ -113,47 +116,68 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
                 revert FailedOp(opIndex, "AA95 out of gas");
             } else if (innerRevertCode == INNER_REVERT_LOW_PREFUND) {
-                // innerCall reverted on prefund too low. treat entire prefund as "gas cost"
-                uint256 actualGas = preGas - gasleft() + opInfo.preOpGas;
-                uint256 actualGasCost = opInfo.prefund;
-                emitPrefundTooLow(opInfo);
-                emitUserOperationEvent(opInfo, false, actualGasCost, actualGas);
-                collected = actualGasCost;
+                // innerCall reverted on prefund too low.
+                // below we re-calculate the cast cost and refund.
```
forshtat: Where? Put function name, lines or some other meaningful reference to what you mean here.

Also, this entire `if` is empty, could we maybe "invert" it to be `else if (!innerRevertCode == INNER_REVERT_LOW_PREFUND) { ... } ` ??

contracts/core/EntryPoint.sol
```diff
@@ -113,47 +116,68 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+            callSuccess,
+            actualGas
+        );
+    }
+
+    /**
+     * call after innerCall executed to call the account's callData and paymaster's postOp.
+     * - calculate the overall gas used and unuased penalty.
+     * - refund payer if needed
+     * - emit UserOperationEvent
```
forshtat: Rewrite this comment. Use ChatGPT.

contracts/core/EntryPoint.sol
```diff
@@ -113,47 +116,68 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+     */
+    function _postInnerCall(
+        UserOpInfo memory opInfo,
+        bool callSuccess,
+        uint256 actualGas)
+    internal virtual returns (uint256 collected) {
+        unchecked {
+            uint256 prefund = opInfo.prefund;
+            uint256 actualGasCost = actualGas * getUserOpGasPrice(opInfo.mUserOp);
+            uint256 refund;
```
forshtat: Are you seriously creating two variables - `refund` and `prefund` here? Absolutely not. These two variables MUST have two or three words names that don't sound anything alike.

contracts/core/EntryPoint.sol
```diff
@@ -113,47 +116,68 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+            if (prefund >= actualGasCost) {
+                refund = prefund - actualGasCost;
+            } else {
+                actualGasCost = prefund;
+                //depending where the over-gas-used was found, we either reverted innerCall or not.
```
forshtat: Instead of writing this comment add a boolean parameter to the `emitPrefundTooLow` - we don't always need it but it will help make the code flow more clear.

contracts/core/EntryPoint.sol
```diff
@@ -356,9 +381,11 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
                 mUserOp.paymasterPostOpGasLimit +
                 INNER_GAS_OVERHEAD
             ) {
+                uint256 gasUsed = preGas - gasleft();
                 assembly ("memory-safe") {
                     mstore(0, INNER_OUT_OF_GAS)
-                    revert(0, 32)
+                    mstore(32, gasUsed)
+                    revert(0, 64)
```
forshtat, line 396: If we are making a  big refactoring now, would you consider extracting this entire `unchecked { ... }` block to some `_verifyEnoughGasForInner` function? This is a very distracting code.

contracts/core/EntryPoint.sol
```diff
@@ -753,75 +781,60 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-                        {} catch {
-                            bytes memory reason = Exec.getReturnData(REVERT_REASON_MAX_LEN);
-                            revert PostOpReverted(reason);
-                        }
-                    }
-                    // Calculating a penalty for unused postOp gas
-                    uint256 postOpGasUsed = postOpPreGas - gasleft();
-                    postOpUnusedGasPenalty = _getUnusedGasPenalty(postOpGasUsed, mUserOp.paymasterPostOpGasLimit);
+            actualGas = executionGas + _getUnusedGasPenalty(executionGas, mUserOp.callGasLimit) + opInfo.preOpGas;
+            if (context.length > 0) {
```
forshtat, line 807: Also for the entire `if (context.length > 0) { ... }` block - it would probably cost nothing to make it an internal function.

contracts/core/EntryPoint.sol
```diff
@@ -753,75 +781,60 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+                uint256 postOpPreGas = gasleft();
+                uint256 actualGasCostForPostOp = actualGas * gasPrice;
+                uint256 paymasterPostOpGasLimit = mUserOp.paymasterPostOpGasLimit;
```
forshtat: Why create a variable if `mUserOp` is in memory anyways? MLOAD costs 3 gas, no?

contracts/core/EntryPoint.sol
```diff
@@ -753,75 +781,60 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+                // todo: no real need to check if paymaster has code. we already called validatePaymasdterUserOp
```
forshtat: Now is not a time to add new TODOs. Either go full assembly (not recommened) or ignore the unnecessary "has code" checks you get from solidity (recommended).

contracts/core/EntryPoint.sol
```diff
@@ -753,75 +781,60 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+                try IPaymaster(paymaster).postOp{
+                        gas: paymasterPostOpGasLimit
+                    }(mode, context, actualGasCostForPostOp, gasPrice)
```
forshtat: We can totally remove the `postOpReverted` "mode" from the codebase by the way.

contracts/core/EntryPoint.sol
drortirosh: suggest better names.

forshtat: I don't know, `originalPrefund` and `remainingGasRefund`?
@shahafn could you help me out here? 😄

contracts/core/EntryPoint.sol
drortirosh: removed.
mUserOp is "double-reference", and in the past, it was better to copy a struct member to a local (stack) one.
but apparently, the compiler now behaves better if we don't.
(btw: its not good optimization of the compiler: it should be agnostic to the way we read again the same memory var)

forshtat:
A remider to self to remove `postOpReverted` from ERC-4337 if this PR is merged.

drortirosh:
Deferred

PR #553, opened 2025-02-27
AA-545: Add 'EIP_7702_MARKER_CODE'

forshtat:


drortirosh:
its a loop: account-abstraction/utils (part of bundler project) already depends on account-abstraction/contracts.
all javascript in this project is for testing, and should be standalone.

forshtat:
> its a loop: account-abstraction/utils (part of bundler project) already depends on account-abstraction/contracts. all javascript in this project is for testing, and should be standalone.

Can we flatten this loop out? I would like this code to exist in only one place.

drortirosh:
4337 contracts is a stand-alone, contracts-only project. all other (bundler, utils) depend on it.
it means that its tests should also not depend on utils - unless we add a "do-nothing" utils, which are independent of contracts...
I prefer little duplicate than cross-repo dependencies.

PR #556, opened 2025-03-02
Refactor _callInnerHandleOp

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -418,7 +423,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
         unchecked {
             uint256 executionGas = preGas - gasleft();
             uint256 actualGas = _postExecution(mode, opInfo, context, executionGas);
-            return (mode == IPaymaster.PostOpMode.opSucceeded, actualGas);
+            return (mode != IPaymaster.PostOpMode.opSucceeded, actualGas);
```
shahafn, line 426: Why reversing the condition? That means that the assembly code that you extracted to an internal function not just moved, but also changed.

drortirosh: Because I wanted 0=success, 1=exec-revert, other=inner-call+reverted

drortirosh:
Deferred from this audit

PR #559, opened 2025-03-02
Remove excessive 'getSender' helper

forshtat:
There are dozens of places where `UserOp` fields are read using normal solidity. There are hundreds of places where using assembly would optimize gas usage.
This optimizations saves a trivial amount of gas compared to the cost of the UserOperation.


shahafn:
Adding in #565

PR #562, opened 2025-03-03
Discussion: bubble up the reverts manually

forshtat:


contracts/core/EntryPoint.sol
```diff
@@ -831,16 +845,19 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
         uint256 actualGasCostForPostOp = actualGas * gasPrice;
+        postOpSuccess = true;
+        // TODO: Use EXEC to catch reverted calls
         try IPaymaster(mUserOp.paymaster).postOp{
                 gas: mUserOp.paymasterPostOpGasLimit
             }(mode, context, actualGasCostForPostOp, gasPrice)
         {} catch {
-            bytes memory reason = Exec.getReturnData(REVERT_REASON_MAX_LEN);
-            revert PostOpReverted(reason);
+            postOpSuccess = false;
```
drortirosh: Why is it better to bubble up 2 parameters, just to throw the same revert?

drortirosh:
Why do you think it is better? 
I think handling exception as close to the Uber call is better, to get it x out of the way"

drortirosh:
in addition, it also doesn't work..

PR #566, opened 2025-03-05
_callValidatePaymaster

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -581,42 +581,73 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+        uint256 freePtr = _getFreePtr();
+        bytes memory validatePaymasterCall = abi.encodeCall(
+            IPaymaster.validatePaymasterUserOp,
+            (op, opInfo.userOpHash, opInfo.prefund)
+        );
+        address paymaster = opInfo.mUserOp.paymaster;
+        bool success;
+        assembly ("memory-safe") {
+            //call and return 3 first words: offset, validation, context-length
+            success := call(gas(), paymaster, 0, add(validatePaymasterCall, 0x20), mload(validatePaymasterCall), freePtr, 96)
```
forshtat: I think that if you read the entire return data here instead of limiting yourself to 96 bytes, you will only lose 64 bytes of "garbage data" per UserOp . However, it will remove the need to manually check and read the data with `returndatacopy` later .
What do you think?

drortirosh: I read the first 3 words: offset, validationData, context-length
in a separte returndatacopy I read the entire context.
yes, I could read entire blob in a single returndatacopy , and then return a pointer into the middle of this struct..
that is:
always do:
```
 returndatacopy(freePtr, returndatasize())
context := add(freePtr,64)
```
(and then all sanity is the same)

is it much simpler?

contracts/core/EntryPoint.sol
```diff
@@ -581,42 +585,80 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+            (op, opInfo.userOpHash, opInfo.prefund)
+        );
+        address paymaster = opInfo.mUserOp.paymaster;
+        bool success;
+        uint256 contextLength;
+        uint256 contextOffset;
+        uint256 maxContextLength;
+        assembly ("memory-safe") {
+            //call and return 3 first words: offset, validation, context-length
+            success := call(gas(), paymaster, 0, add(validatePaymasterCall, 0x20), mload(validatePaymasterCall), freePtr, 96)
```
shahafn: I think that removing `verificationGasLimit` and `paymasterVerificationGasLimit` from the call here is not good: we introduce another unpredictable factor to simulation vs execution of this validation, depending on the transaction gas limit provided by the bundler to `handleOps()`, and it's almost always going to differ between simulating a userOp and simulating the entire bundle. 
Even if we don't see the issue now, is there a benefit of introducing this?

drortirosh: I added it here, since that was the behaviour beforehand.
but still: I think we should remove the verification-limit from teh calls, since the actual value is checked with a `preGas-gasleft()`. that is, the gas-limit on the calls can never be reached, since the gas-limit should cover also calling into the function (copying data in reading result, and for verification-gas, it also covers creation and nonce increment.

moreover: currently, you can't tell OOG from revert.... but if we separate, you have a clear separation between AA36 (gas limit) and AA33 (revert) ...

it requires few tests to fix, because they chekc the wrong error code.

contracts/core/EntryPoint.sol
```diff
@@ -595,42 +599,79 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
+        uint256 maxContextLength;
+        assembly ("memory-safe") {
+            //call and return 3 first words: offset, validation, context-length
+            success := call(paymasterVerificationGasLimit, paymaster, 0, add(validatePaymasterCall, 0x20), mload(validatePaymasterCall), freePtr, 96)
+            validationData := mload(add(freePtr, 32))
+            contextLength := mload(add(freePtr, 64))
+            contextOffset := mload(freePtr)
+            maxContextLength := sub(returndatasize(), 96)
+        }
+        if (!success || contextOffset != 64 || contextLength > maxContextLength) {
```
forshtat, line 663: Regarding `contextOffset != 64`, this is not a strict requirement of the ABI encoding as far as I can tell. 

If someone were to implement the paymaster interface this way:
```
function validatePaymasterUserOp(
...
) external returns (
    bytes memory context,
    uint256 validationData,
    uint256 myBirthday
);
```

This function would've been valid with Solidity but not with this assmebly, I assume.
Is this so? Should we care about it?

forshtat:
@drortirosh is there a way to "check" that we made the `finalize_allocation` correctly? The tests don't check much, I can call it with `finalize_allocation(freePtr, add(contextDataLen, 10000))` and tests don't fail, do they?

drortirosh:
> @drortirosh is there a way to "check" that we made the `finalize_allocation` correctly? The tests don't check much, I can call it with `finalize_allocation(freePtr, add(contextDataLen, 10000))` and tests don't fail, do they?

It's hard to check on the final code..
If you comment it out, the postop.test would pass on the small context tests and fail on the larger ones (once context is larger than the size of the encodeCall  to validatePaymasterUserOp)

contracts/core/EntryPoint.sol
shahafn: Why check the context offset? We implicitly force the abi encoding of it

shahafn: At least document the magic numbers, and explain that you're assuming a specific abi encoding of the return value

shahafn (approved):

PR #567, opened 2025-03-09
Remove excessive assembly sanity-checking

forshtat:


contracts/core/EntryPoint.sol
```diff
@@ -621,43 +621,26 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
         );
         address paymaster = opInfo.mUserOp.paymaster;
         bool success;
+        uint256 contextOffset;
         uint256 contextLength;
+        uint256 returnDataSize;
         assembly ("memory-safe") {
-            //call and return 3 first words: offset, validation, context-length
-            success := call(gas(), paymaster, 0, add(validatePaymasterCall, 0x20), mload(validatePaymasterCall), freePtr, 96)
+            success := call(gas(), paymaster, 0, add(validatePaymasterCall, 0x20), mload(validatePaymasterCall), freePtr, 2112)
```
drortirosh, line 628: where this number came from? also, we don't need to copy always the worst case, as it should be configurable (and not in the solidity code, but in bundler

contracts/core/EntryPoint.sol
```diff
@@ -621,43 +621,26 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-            revert FailedOpWithRevert(opIndex, "AA33 reverted", Exec.getReturnData(REVERT_REASON_MAX_LEN));
         }
-        assembly ("memory-safe") {
-            // we use freePtr, fetched before calling encodeCall, as return data pointer.
-            // this way we reuse that memory without unnecessary memory expansion
-            context := freePtr
-            let contextDataLen := add(contextLength, 32)
-            //read entire context (including length)
-            returndatacopy(context, 64, contextDataLen)
-            finalize_allocation(freePtr, contextDataLen)
```
drortirosh, line 648: your code never finalized allocation. running it in a batch will override the context by next userop.

contracts/core/EntryPoint.sol
```diff
@@ -621,43 +621,26 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-
-            function finalize_allocation(memPtr, size) {
-                let newFreePtr := add(memPtr, round_up_to_mul_of_32(size))
-                mstore(64, newFreePtr)
-            }
+        bool failedOp =
+                !success ||
```
drortirosh, line 636: you're right... testing conditions in solidity looks nicer.

PR #568, opened 2025-03-10
Changing solidity version

shahafn:


contracts/accounts/Simple7702Account.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8;
+pragma solidity ^0.8.28;
```
drortirosh, line 2: I don't think it is a good idea to "shotgun" all sources wiht exact minor version we use. better stick with "major" , that is"0.8"
(the exception is EntryPoint itself, where the actual sub-version is important)

contracts/interfaces/IAccount.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.7.5;
```
drortirosh, line 2: interfaces are deliberately ">=" and not "^". That is, we allow any version (That supports ABIv2 encoding). we don't want to change them anymore, even when compiler changes to 0.9 or 1.0

contracts/accounts/Simple7702Account.sol
drortirosh: (that is, I do think we need most of this refactor - but remove the minor subversion, instead of updating it to the current latest)

forshtat: @drortirosh 
That is the complete opposite of what we need. Solidity is actually making some pretty important, breaking changes in "minor" versions, and we do rely on them a lot. Our code will not compile with older versions.
Also, this is not a code that has to be imported and inherited in other projects, but only used when deployed on-chain.
We MUST specify the exact solidity version we are using, 100%.

contracts/interfaces/IAccount.sol
forshtat: I'm okay with having a wide range for interfaces. Not necessary as there are many alternative repos for these interface declarations now.

shahafn: So why not remove this requirement from interfaces altogether?

forshtat (approved):

PR #570, opened 2025-03-11
Minor changes 5

forshtat:


contracts/interfaces/IEntryPoint.sol
```diff
@@ -144,6 +144,13 @@ interface IEntryPoint is IStakeManager, INonceManager {
         bytes signature;
     }
 
+    /**
+     * Get the EntryPoint contract's version.
+     *
+     * @return the ERC-712 value used as DOMAIN_VERSION for UserOperation hash and signature calculation.
```
drortirosh, line 150: Remove the DOMAIN_VERSION, and make it non-public, as it is meaningless.
It is used to calculate the getDomainSeparator, which is public.
also, the current value "0.8" is not a valid ERC-712 domain (should be the "major version").
also, it is the version of the ABI, not the current implementation version.

PR #571, opened 2025-03-11
remove gas: paramter from verification calls

drortirosh:
remove gas passed to validation calls
from all contract creation, validateUserOp, validatePaymasterUserOp
the reasoning:
- actual gas limit is enforced by `require( preGas-gasleft() < verificationGasLimit )`, so there is no need for duplicate calls
- because the above mentioned tests, gas limit to a method can never be reached.
- it allows distinguishing revert on OOG from revert AA36/AA26 for "oog", and AA13/AA23/AA33 for reverts on the respective entity
- the only thing that that the above gas limits added was to let the bundler run the simulation with a fixed (huge) gas limit. now the bundler has to pass `verificationGasLimit+paymasterVerificationGasLimit+preVerificationGasLimit`, otherwise a UserOp might cause wasting a lot of gas during simulation.


drortirosh:
defered

PR #574, opened 2025-03-13
EntryPoint: move all public/external methods first.

drortirosh:


contracts/core/EntryPoint.sol
```diff
@@ -55,15 +179,7 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ReentrancyGuardT
...
-    /// @inheritdoc IERC165
-    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
-        // note: solidity "type(IEntryPoint).interfaceId" is without inherited methods but we want to check everything
-        return interfaceId == (type(IEntryPoint).interfaceId ^ type(IStakeManager).interfaceId ^ type(INonceManager).interfaceId) ||
-        interfaceId == type(IEntryPoint).interfaceId ||
-        interfaceId == type(IStakeManager).interfaceId ||
-        interfaceId == type(INonceManager).interfaceId ||
-            super.supportsInterface(interfaceId);
-    }
+    SenderCreator private immutable _senderCreator = new SenderCreator();
```
forshtat: Please let's keep ALL constants and storage variables before ANY  function code?

forshtat (changes requested):

forshtat:
@drortirosh we will lose all the "git blame" history with this purely cosmetic change.

drortirosh:
Yes, we lose some git blame history, not all.

contracts/core/EntryPoint.sol
drortirosh: the reason of putting public/external first is to keep all external ABI first.
nothing about the private constants is external. I don't really see a reason to put them first, as they mean nothing outside their context.

forshtat (dismissed):

forshtat (approved):

PR #607, opened 2025-10-03
fix: bug call eip:4337

Dargon789:


shahafn:
Please reopen a new PR if you want to suggest a change, this one reintroduces files we already removed from the repo long ago, and no explanation is given here

PR #609, opened 2025-10-30
Release ERC-4337 v0.9

forshtat:
This version is ABI-compatible with v0.8 and v0.7. No breaking changes were made to the EntryPoint ABI.

Bundlers must update to join the v0.9 UserOp mempool.

Features:
* Enable parallel Paymaster signing via 'paymasterSignature' field
* Allow 'validAfter' & 'validUntil' to use block numbers (via high bit)
* Ignore 'initCode' if account already exists (enables 2D nonce for first userOps)
* Add 'getCurrentUserOpHash()' to EntryPoint
* Improve error messages and add 'EIP7702AccountInitialized' event
* 'BasePaymaster' constructor now accepts 'owner' (instead of hard-coded msg.sender)


drortirosh (approved):

livingrockrises:
https://docs.erc4337.io/paymasters/paymaster-signature.html

does this mean paymasterAndData must not include paymaster signature anymore? (for verifying paymaster earlier it was part of paymasterAndData)

gonzaotc:
From reviewing the code, I think it is also notable for the release notes that the `nonReentrant` modifier rework now enforces that only pure EOAs (no code) can submit bundles. Contracts and EIP-7702 delegated EOAs are now restricted from bundle submission, which I understand simplifies the security model. I think it would also make sense to annotate this restriction in the ERC/docs

gonzaotc:
+1 to @livingrockrises' question. In OZ, we have a [PaymasterSigner](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/account/paymaster/PaymasterSigner.sol) implementation that validates signatures encoded in the `paymasterData` field. Would you consider it a good idea to migrate this signature to the new `paymasterSignature` field even when parallelization isn't required, for the sake of standardization and consistency now that this field is available?

PR #616, opened 2025-12-19
Fix accounts for gas estimation

Arvolear:
Hey, while testing out the provided account implementations, I've noticed that it's impossible to estimate gas with them. This is because during the estimation, bundler sets the user operation signature as empty (to `0x`) and the `ECDSA.recover()` function call reverts with the error "InvalidSignatureLength".

I've added a very quick fix to address this issue.


drortirosh:
bundler will only set the signature if you don't.
as a user that uses estimateUserOp, you should put in a signature that passes all encoding requirement, but only fails on actual signature check 
e.g: you can sign once a bogus userop, and keep use that "stub" signature for simulations. 
it will pass all encoding, length and other validations, and "only" recover to the wrong result and thus fail on SIG_VALIDATION_FAILED.

Arvolear:
Ah, I see, that's now very obvious. It would be good to see it documented anywhere. Should I add a few lines to the account interface about that?

PR #623, opened 2026-08-05
fix: nonReentrant locks out EIP-7702-delegated bundler EOAs

SashaMIT:
## Summary

- v0.9's `nonReentrant` requires `tx.origin == msg.sender && msg.sender.code.length == 0`.
- The `tx.origin == msg.sender` clause alone stops reentrancy: any reentrant call into `handleOps`/`handleAggregatedOps` necessarily arrives through a contract, which fails that check.
- The `code.length == 0` clause adds nothing against reentrancy but **permanently locks out every bundler EOA that has ever set an EIP-7702 delegation** — their code length is nonzero after delegation even though they initiate the transaction exactly like any other EOA (no code executes in an EOA-initiated call).
- Remove the redundant clause and add a regression test: a permanently 7702-delegated bundler successfully submits `handleOps`. Previously this reverted with `Reentrancy()`; no test covered the path (no test referenced the guard at all).

## Impact if unsolved

As 7702 adoption grows, bundlers that delegate their EOAs (batching, gas sponsorship, session keys) are silently unable to submit to the new EntryPoint — a liveness/compat regression baked into the protocol's core modifier, discovered only at integration time.

## Test plan

- New test: `should accept handleOps from an EIP-7702-delegated bundler EOA` (mirrors the repo's existing raw-7702 transaction pattern).
- `test/entrypoint.test.ts`: 75/75 pass with the modifier change.
- `test/entrypoint-7702.test.ts`: 13/13 code tests pass (1 env failure — local geth harness needs Docker, unavailable here; CI covers it).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/contracts/core/EntryPoint.sol
+++ b/contracts/core/EntryPoint.sol
@@ -66,9 +66,16 @@ contract EntryPoint is IEntryPoint, StakeManager, NonceManager, ERC165, EIP712 {
     }
 
     modifier nonReentrant() {
+        // tx.origin == msg.sender already guarantees the caller is the transaction-signing
+        // account itself, so no contract code runs in this call chain. Requiring
+        // msg.sender.code.length == 0 as well adds nothing against reentrancy — a
+        // reentrant call always arrives through a contract, which fails the tx.origin
+        // check — but it permanently locks out EIP-7702-delegated bundler EOAs, whose
+        // code length is nonzero after delegation even though they initiate the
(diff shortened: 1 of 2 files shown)


drortirosh:
Bundlers already have "pool of addresses", so there is no reason for them to re-use the same address using EIP-7702
However, there are other vulnerabilities we wanted to prevent when source address is not an EOA.

PR #625, opened 2026-08-14
Fix local geth dev script and EIP-7702 test gas limit

witchinghourartcollective:
## Summary
- `scripts/geth.sh`: quote the `--http.corsdomain`/`--http.vhosts` wildcards so bash doesn't glob-expand `*` to filenames in the repo root before they reach `docker run` (this was breaking the local geth dev container — it would receive repo filenames like `README.md` as bogus args and fail to start).
- `test/entrypoint-7702.test.ts`: set an explicit `verificationGasLimit: 300000` on the "should fail without sender delegate" case — the default `150000` isn't enough for the delegate-init flow.

## Test plan
- [x] `yarn test` — 146/146 passing, including the geth-integration EIP-7702 cases

### DIFF
--- a/scripts/geth.sh
+++ b/scripts/geth.sh
@@ -3,5 +3,6 @@ name=geth-$$
 trap "echo killing docker; docker kill $name 2> /dev/null" EXIT
 port=$1
 shift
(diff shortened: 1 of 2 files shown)


witchinghourartcollective (commented):
Approve

